PR #2, opened 2021-01-29
Add support for go.mod file generation

A-Hilaly:
Issue: N/A
Changes:
- Add support for go.mod file generation
- Expand `ack-generate` configuration to allow specifying required modules versions

`generator.yaml` configuration example:
```yaml
go_module:
  compiler_version: v1.13.13
  required_modules_versions:
    aws_sdk_go: v1.35.34
    ack_core: v0.0.2 #for now sets 'github.com/aws/aws-controllers-k8s' version
```

### DIFF
--- a/pkg/generate/ack/config.go
+++ b/pkg/generate/ack/config.go
@@ -23,4 +23,15 @@ var DefaultConfig = config.Config{
 	},
 	IncludeACKMetadata:             true,
 	SetManyOutputNotFoundErrReturn: "return nil, ackerr.NotFound",
+	GoModule: config.GoModule{
+		CompilerVersion: "1.15",
+		RequiredModulesVersions: config.RequiredModulesVersions{
+			AWSSDKGo:             "v1.35.5",
+			ACKCore:              "v0.0.2",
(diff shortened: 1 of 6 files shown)


jaypipes (commented):
@A-Hilaly I actually don't think we should generate the go.mod file in the code generator. I think placing a default `go.mod` file in the github.com/aws-controllers-k8s/template-controller` is the best solution, and then let the service teams adapt their module dependencies as needed (in particular, updating the aws-sdk-go pinned version.

Having go.mod generated by the code generator means that these module dependencies would need to be put in the `generator.yaml` config file and that means we'd need to add some sort of disclaimer in the generated `go.mod` template that essentially says "DO NOT EDIT", which itself seems odd for a `go.mod` file.

A-Hilaly:
> Having go.mod generated by the code generator means that these module dependencies would need to be put in the generator.yaml config file and that means we'd need to add some sort of disclaimer in the generated go.mod template that essentially says "DO NOT EDIT", which itself seems odd for a go.mod file.

Very good point! I'm closing this PR and adding `go.mod` to the `template-controller` repository

PR #9, opened 2021-02-04
crossplane: code generator for lateInitialize function

muvaf:
Description of changes: Crossplane uses late initialization functions to fill up the empty fields in spec with the current state in the AWS API, defaults most of the time. This PR introduces code generator to generate that late initialization function.

Fixes https://github.com/crossplane/provider-aws/issues/491

An example output looks like the following:
```go
func lateInitialize(cr *svcapitypes.Route, resp *svcsdk.GetRouteOutput) (bool, error) {
	li := resource.NewLateInitializer()
	cr.Spec.ForProvider.APIKeyRequired = li.LateInitializeBoolPtr(cr.Spec.ForProvider.APIKeyRequired, resp.ApiKeyRequired)
	if len(resp.AuthorizationScopes) != 0 && len(cr.Spec.ForProvider.AuthorizationScopes) == 0 {
		cr.Spec.ForProvider.AuthorizationScopes = make([]*string, len(resp.AuthorizationScopes))
		for i0 := range resp.AuthorizationScopes {
			cr.Spec.ForProvider.AuthorizationScopes[i0] = li.LateInitializeStringPtr(cr.Spec.ForProvider.AuthorizationScopes[i0], resp.AuthorizationScopes[i0])
		}
	}
	cr.Spec.ForProvider.AuthorizationType = li.LateInitializeStringPtr(cr.Spec.ForProvider.AuthorizationType, resp.AuthorizationType)
	cr.Spec.ForProvider.AuthorizerID = li.LateInitializeStringPtr(cr.Spec.ForProvider.AuthorizerID, resp.AuthorizerId)
	cr.Spec.ForProvider.ModelSelectionExpression = li.LateInitializeStringPtr(cr.Spec.ForProvider.ModelSelectionExpression, resp.ModelSelectionExpression)
	cr.Spec.ForProvider.OperationName = li.LateInitializeStringPtr(cr.Spec.ForProvider.OperationName, resp.OperationName)
	if resp.RequestModels != nil {
		if cr.Spec.ForProvider.RequestModels == nil {
			cr.Spec.ForProvider.RequestModels = map[string]*string{}
		}
		for key0 := range resp.RequestModels {
			cr.Spec.ForProvider.RequestModels[key0] = li.LateInitializeStringPtr(cr.Spec.ForProvider.RequestModels[key0], resp.RequestModels[key0])
		}
	}
	if resp.RequestParameters != nil {
		if cr.Spec.ForProvider.RequestParameters == nil {
			cr.Spec.ForProvider.RequestParameters = map[string]*svcapitypes.ParameterConstraints{}
		}
		for key0 := range resp.RequestParameters {
			if resp.RequestParameters[key0] != nil {
				if cr.Spec.ForProvider.RequestParameters[key0] == nil {
					cr.Spec.ForProvider.RequestParameters[key0] = &svcapitypes.ParameterConstraints{}
				}
				cr.Spec.ForProvider.RequestParameters[key0].Required = li.LateInitializeBoolPtr(cr.Spec.ForProvider.RequestParameters[key0].Required, resp.RequestParameters[key0].Required)
			}
		}
	}
	cr.Spec.ForProvider.RouteKey = li.LateInitializeStringPtr(cr.Spec.ForProvider.RouteKey, resp.RouteKey)
	cr.Spec.ForProvider.RouteResponseSelectionExpression = li.LateInitializeStringPtr(cr.Spec.ForProvider.RouteResponseSelectionExpression, resp.RouteResponseSelectionExpression)
	cr.Spec.ForProvider.Target = li.LateInitializeStringPtr(cr.Spec.ForProvider.Target, resp.Target)
	return li.IsChanged(), nil
}
```

It's been tested with `ReadOne`, `ReadMany` and `GetAttributes` outputs.

@jaypipes I wanted to get your take on this approach ~~before writing unit tests~~ (unit tests are written now). I started by copying `SetResource` functions but couldn't really make it work for this scenario so I wrote it from scratch. Though the end result is similar anyway.

### DIFF
--- a/pkg/generate/code/late_initialization.go
+++ b/pkg/generate/code/late_initialization.go
@@ -1,0 +1,246 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package code
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+
+	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
+
+	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/generate/config"
+	"github.com/aws-controllers-k8s/code-generator/pkg/model"
+	"github.com/aws-controllers-k8s/code-generator/pkg/names"
+)
+
+func LateInitializeReadOne(
+	cfg *ackgenconfig.Config,
+	r *model.CRD,
+// String representing the name of the variable that we will grab the
+// Output shape from. This will likely be "resp" since in the templates
+// that call this method, the "source variable" is the response struct
+// returned by the aws-sdk-go's SDK API call corresponding to the Operation
+	sourceVarName string,
+// String representing the name of the variable that we will be **setting**
+// with values we get from the Output shape. This will likely be
+// "ko.Status" since that is the name of the "target variable" that the
+// templates that call this method use.
+	targetPath string,
+) string {
+	out := ""
+	path := sourceVarName
+	shape := r.Ops.ReadOne.OutputRef.Shape
+	if len(shape.MemberRefs) == 1 {
+		for key, val := range shape.MemberRefs {
+			if val.Shape.Type == "structure" {
+				path = fmt.Sprintf("%s.%s", path, key)
+				shape = val.Shape
+				break
+			}
+			panic("there has to be a structure field in unwrapped readone output shape")
+		}
+	}
+	for _, memberName := range shape.MemberNames() {
+		f, found := r.SpecFields[memberName]
+		if !found {
+			continue
+		}
+		responsePath := fmt.Sprintf("%s.%s", path, f.Names.Original)
+		crPath := fmt.Sprintf("%s%s.%s", targetPath, cfg.PrefixConfig.SpecField, f.Names.Camel)
+		out += lateInit(responsePath, crPath, r, f.ShapeRef, 0)
+	}
+	return strings.TrimSuffix(out, "\n")
+}
+
+func LateInitializeReadMany(
+	cfg *ackgenconfig.Config,
+	r *model.CRD,
+// String representing the name of the variable that we will grab the
+// Output shape from. This will likely be "resp" since in the templates
+// that call this method, the "source variable" is the response struct
+// returned by the aws-sdk-go's SDK API call corresponding to the Operation
+	sourceVarName string,
+// String representing the name of the variable that we will be **setting**
+// with values we get from the Output shape. This will likely be
+// "ko.Status" since that is the name of the "target variable" that the
+// templates that call this method use.
+	targetPath string,
+) string {
+	listFieldName := ""
+	var respShapeRef *awssdkmodel.ShapeRef
+	for name, ref := range r.Ops.ReadMany.OutputRef.Shape.MemberRefs {
+		if ref.Shape.Type == "list"{
+			listFieldName = name
+			respShapeRef = &ref.Shape.MemberRef
+			break
+		}
+	}
+	if respShapeRef == nil {
+		panic("could not find a list shaped member in readmany output shape")
+	}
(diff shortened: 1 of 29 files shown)


pkg/generate/code/late_initialization.go
```diff
@@ -0,0 +1,246 @@
...
+	path := sourceVarName
+	shape := r.Ops.ReadOne.OutputRef.Shape
+	if len(shape.MemberRefs) == 1 {
+		for key, val := range shape.MemberRefs {
+			if val.Shape.Type == "structure" {
+				path = fmt.Sprintf("%s.%s", path, key)
+				shape = val.Shape
+				break
+			}
+			panic("there has to be a structure field in unwrapped readone output shape")
```
jaypipes, line 52: I'm sure we'll find an AWS API that will prove this wrong :)

pkg/generate/code/late_initialization.go
```diff
@@ -0,0 +1,246 @@
...
+	"strings"
+
+	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
+
+	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/generate/config"
+	"github.com/aws-controllers-k8s/code-generator/pkg/model"
+	"github.com/aws-controllers-k8s/code-generator/pkg/names"
+)
+
+func LateInitializeReadOne(
```
jaypipes, line 28: For all the exported functions in this file, please add docstrings.

I would also like to have a docstring block at the top of this file that covers the basic concepts of what the `LateIntialize` code is for. I know at the moment it's crossplane-specific, which is totes fine, but having some documentation about *how Crossplane uses this generated code* would be super helpful in determining whether ACK service controllers may eventually align their generated code to use the Crossplane LateInitialize functionality :)

pkg/generate/code/late_initialization.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+	goType := s.GoTypeWithPkgName()
+	goType = model.ReplacePkgName(goType, r.SDKAPIPackageName(), "svcapitypes", keepPointer)
+	goTypeNoPkg := strings.Split(goType, ".")[1]
+	goPkg := strings.Split(goType, ".")[0]
+	if r.TypeRenames()[goTypeNoPkg] != "" {
+		goTypeNoPkg = r.TypeRenames()[goTypeNoPkg]
+	}
+	return fmt.Sprintf("%s.%s", goPkg, goTypeNoPkg)
+}
```
jaypipes, line 246: The above seems to be reimplementing some parts of this function:

https://github.com/aws-controllers-k8s/code-generator/blob/41d11f752387871a48a9a88df99b7a751af8a75e/pkg/model/types.go#L28-L74

and some parts of these functions:

https://github.com/aws/aws-sdk-go/blob/27d5e03d86c284f01aabb029636efb92ec3d167e/private/model/api/shape.go#L306-L428

Is that your intention here? If so, you may be able to simplify the code above by either exporting the `pkg/model:cleanGoType` function and/or directly using some of the `aws-sdk-go/private/model/api:Shape` methods.

pkg/generate/code/late_initialization.go
```diff
@@ -0,0 +1,246 @@
...
+	case "structure":
+		return lateInitStruct(responsePath, crPath, r, str, level)
+	case "map":
+		return lateInitMap(responsePath, crPath, r, str, level)
+	default:
+		panic(fmt.Sprintf("unknown shape type %s", str.Shape.Type))
+	}
+}
+
+func GetCorrespondingCRDType(s *awssdkmodel.Shape, r *model.CRD, keepPointer bool) string {
```
jaypipes, line 225: Definitely this needs a docstring :) Also, let's rename this function to be more explicit. *What* specifically is the returned type "corresponding" to? Perhaps rename to `GetCRDStructType()` to match the aws-sdk-go naming convention?

jaypipes (changes requested):
Hi @muvaf! Sorry for the delayed review on this! I left some requests inline, mostly to do with documentation. Code generally looks great, though, thank you!

pkg/generate/code/late_initialization.go
muvaf: `pkg/model:cleanGoType` seems to calculate the new type name for the struct in the CRD package but here we need something similar to [`varEmptyConstructorK8sType`](https://github.com/aws-controllers-k8s/code-generator/blob/eef7ca5/pkg/generate/code/set_sdk.go#L1094) because the type string is used to write the initialization statement. In fact, it seems like we can reuse the `GetCorrespondingCRDType` function there as well.

> using some of the aws-sdk-go/private/model/api:Shape methods.

I believe except for `strings` operations, it's all methods of the `Shape` object already but seems like we don't need to take every shape type into account as `GoTypeWithPkgName`/`GoType` functions already do that for us. Pushed an update.

PR #11, opened 2021-02-05
crossplane: referencer and selector fields generation

muvaf:
This PR is migrated from https://github.com/aws/aws-controllers-k8s/pull/643

Description of changes: Crossplane uses two additional fields for the ones that can be referenced by another CRD - `Reference` and `Selector`. This PR allows users to enter the type of the referenced object so that we automatically generate these fields. The type information is necessary for the resolver but since we're not yet generating it, only the existence of a value in `referenced_type` is used. When we generate the resolver code, we'll need type information as well.

One caveat is that any field on the CR (including the ones in `types.go`) could be a candidate as a reference field but from what I can tell, we are only able to address the top level fields in `Config` object. For now, that will be a limitation we'll live with.

Fixes https://github.com/crossplane/provider-aws/issues/484

### DIFF
--- a/pkg/generate/config/field.go
+++ b/pkg/generate/config/field.go
@@ -136,6 +136,10 @@ type FieldConfig struct {
 	// that owns the resource. This is a special field that we direct to
 	// storage in the common `Status.ACKResourceMetadata.OwnerAccountID` field.
 	IsOwnerAccountID bool `json:"is_owner_account_id"`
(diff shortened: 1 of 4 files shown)


muvaf:
Coming back to this, I think it provides little value because of the fact that it's not able to add the ref/selector fields to types in `types.go`. So, I'll close this PR and open a new one in the future when this becomes possible.

PR #13, opened 2021-02-07
Add support for non-default AWS profile

abyss:
Description of changes:
Currently, the kind-test uses the `daws` script to run AWS cli commands. However, it does not respect $AWS_PROFILE, and instead always uses default.

This pull request changes it, such that if $AWS_PROFILE is set, it will be passed as an environment variable to the Docker run.

There is a tiny bit of wonkiness so that it does not pass anything when $AWS_PROFILE is not set - because in testing, I could not get the --profile option (which may never be used in this setting, either way) to override the environment variable even if it's set to empty.

### DIFF
--- a/scripts/lib/aws.sh
+++ b/scripts/lib/aws.sh
@@ -21,9 +21,10 @@ DEFAULT_AWS_CLI_VERSION="2.0.52"
 # To use a specific version of the AWS CLI, set the ACK_AWS_CLI_IMAGE_VERSION
 # environment variable, otherwise the value of DEFAULT_AWS_CLI_VERSION is used.
 daws() {
+    aws_cli_profile_env=$([ ! -z "$AWS_PROFILE" ] && echo "--env AWS_PROFILE=$AWS_PROFILE")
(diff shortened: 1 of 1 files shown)


jaypipes:
@A-Hilaly can you take a look at this please? I know we have support for multiple profiles in the `daws` command...

A-Hilaly:
@jaypipes we use profiles in two different files: [s3.sh#L33-L40](https://github.com/aws-controllers-k8s/community/blob/main/scripts/lib/aws/s3.sh#L33-L40) - [sns.sh#L33-L39](https://github.com/aws-controllers-k8s/community/blob/main/scripts/lib/aws/sns.sh#L33-L39), and both of them use the `--profile` flag. Currently `daws` uses the default credentials/profile (if `--profile` is not provided). 

The changes introduced in this PR, add the ability to set the default profile used in all `daws` calls. `s3`/`sns` will continue to work as expected according to https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html#cli-configure-quickstart-precedence

jaypipes (approved):
:+1: thanks for keeping me honest @A-Hilaly :) And thank you @abyss for the patch!

PR #14, opened 2021-02-10
Adopted resource template changes

RedbackThomson:
Issue #, if available: 

Description of changes:
- Store `NameField` as part of each CRD - for use in templates
- Add the API scheme to each controller
- Provide each controller with permissions to access the `AdoptedResource` CRD
- Provide method to override resource object meta

### DIFF
--- a/pkg/generate/code/set_sdk.go
+++ b/pkg/generate/code/set_sdk.go
@@ -223,7 +223,7 @@ func SetSDK(
 			out += fmt.Sprintf(
 				"%s} else {\n", indent,
 			)
-			nameField := r.NameField()
+			nameField := *r.SpecIdentifierField()
 			out += fmt.Sprintf(
 				"%s\t%s.Set%s(rm.ARNFromName(*%s.Spec.%s))\n",
 				indent, targetVarName, memberName, sourceVarName, nameField,
@@ -444,7 +444,7 @@ func SetSDKGetAttributes(
 			out += fmt.Sprintf(
 				"%s} else {\n", indent,
 			)
-			nameField := r.NameField()
+			nameField := *r.SpecIdentifierField()
 			out += fmt.Sprintf(
 				"%s\t%s.Set%s(rm.ARNFromName(*%s.Spec.%s))\n",
 				indent, targetVarName, memberName, sourceVarName, nameField,
@@ -617,7 +617,7 @@ func SetSDKSetAttributes(
 			out += fmt.Sprintf(
 				"%s} else {\n", indent,
 			)
-			nameField := r.NameField()
+			nameField := *r.SpecIdentifierField()
 			out += fmt.Sprintf(
 				"%s\t%s.Set%s(rm.ARNFromName(*%s.Spec.%s))\n",
 				indent, targetVarName, memberName, sourceVarName, nameField,
--- a/pkg/generate/config/resource.go
+++ b/pkg/generate/config/resource.go
@@ -69,6 +69,10 @@ type ResourceConfig struct {
 	// All ShortNames must be distinct from any other ShortNames installed into the cluster,
 	// otherwise the CRD will fail to install.
 	ShortNames []string `json:"shortNames,omitempty"`
+	// IsAdoptable determines whether the CRD should be accepted by the adoption reconciler.
+	// If set to false, the user will be given an error if they attempt to adopt a resource
+	// with this type.
+	IsAdoptable *bool `json:"is_adoptable,omitempty"`
 }
 
 // CompareConfig informs instruct the code generator on how to compare two different
@@ -348,4 +352,20 @@ func (c *Config) ResourceShortNames(resourceName string) []string {
 		return nil
 	}
 	return rConfig.ShortNames
+}
+
+// ResourceIsAdoptable returns whether the given CRD is adoptable
+func (c *Config) ResourceIsAdoptable(resourceName string) bool {
+	if c == nil {
+		return true
+	}
+	rConfig, ok := c.Resources[resourceName]
+	if !ok {
+		return true
+	}
+	// Default to True
+	if rConfig.IsAdoptable == nil {
+		return true
+	}
+	return *rConfig.IsAdoptable
 }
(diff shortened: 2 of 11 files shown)


pkg/generate/generator.go
```diff
@@ -223,6 +223,9 @@ func (g *Generator) GetCRDs() ([]*ackmodel.CRD, error) {
 			}
 		}
 
+		// TODO(nithomso): Is this the correct casing?
+		crd.NameField = crd.GetNameField()
```
RedbackThomson, line 227: Yeah this is unclean. This needs to be called after all the fields have been added, though. Thoughts on better method?

jaypipes (commented):
@RedbackThomson this needs a rebase, yeah?

templates/pkg/resource/registry.go.tpl
```diff
@@ -7,6 +7,11 @@ import (
 	acktypes "github.com/aws-controllers-k8s/runtime/pkg/types"
 )
 
+// +kubebuilder:rbac:groups=services.k8s.aws,resources=adoptedresources,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=services.k8s.aws,resources=adoptedresources/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups="",resources=namespaces,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch
```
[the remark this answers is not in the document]
RedbackThomson: ^^ Removed from `manager.go.tpl` and put them here.

pkg/generate/config/resource.go
```diff
@@ -349,3 +353,19 @@ func (c *Config) ResourceShortNames(resourceName string) []string {
...
+// ResourceIsAdoptable returns whether the given CRD is adoptable
+func (c *Config) ResourceIsAdoptable(resourceName string) bool {
+	if c == nil {
+		return true
+	}
+	rConfig, ok := c.Resources[resourceName]
+	if !ok {
+		return true
+	}
+	// Default to True
```
jaypipes, line 366: I'm wondering actually if we should default to `false` instead. My reasoning is two-fold. First, the zero value of a `bool` is false, so an "empty" `is_adoptable` field in the generator YAML would more naturally be considered `false`, not `true`. Secondly, an "opt-in" approach I think is the most prudent choice here, especially for highly-nested or related resources. Consider, for example, the case of API Gateway v2 where you have a top-level `Api` resource that has many related sub-resources under it, for example `Stage`, `Authorizer`, `Deployment`, etc. We don't (can't?) allow those sub-resources to be adopted because those sub-objects depend on the existence of a parent resource (an `Api`), and since our AdoptedResource struct only allows for singular primary key identifiers, adoption of dependent related objects isn't possible.

pkg/generate/generator.go
```diff
@@ -223,6 +223,9 @@ func (g *Generator) GetCRDs() ([]*ackmodel.CRD, error) {
 			}
 		}
 
+		crd.SpecIdentifierField = crd.GetSpecIdentifierField()
+		crd.IsAdoptable = g.cfg.ResourceIsAdoptable(crdName)
```
jaypipes, line 227: For the above, please move the assignment of CRD fields into the `ackmodel.NewCRD` function.

pkg/model/crd.go
```diff
@@ -94,6 +94,12 @@ type CRD struct {
 	// ShortNames represent the CRD list of aliases. Short names allow shorter
 	// strings to match a CR on the CLI.
 	ShortNames []string
+	// SpecIdentifierField is the original SDK member name of the identifier
+	// spec field
+	SpecIdentifierField *string
+	// IsAdoptable determines whether the resource should be accepted by the
+	// adoption reconciler
+	IsAdoptable bool
```
jaypipes, line 102: I'd prefer to have a `CRD.IsAdoptable()` method and not cache this information as a bool field... the other CRD methods that return config-driven attributes work this way and I don't see a need to optimize for performance in the code generator in this way. Same for `GetSpecIdentifierField()`... just rename it to `SpecIdentifierField` and get rid of the CRD.SpecIdentifierField field...

templates/pkg/resource/descriptor.go.tpl
```diff
@@ -172,3 +173,21 @@ func (d *resourceDescriptor) MarkUnmanaged(
...
+		// Should not happen. If it does, there is a bug in the code
+		panic("nil RuntimeMetaObject in AWSResource")
+	}
+	curr := obj.GetAnnotations()
+	if curr == nil {
+		curr = make(map[string]string)
+	}
+	curr[ackv1alpha1.AnnotationAdopted] = "true"
+	obj.SetAnnotations(curr)
+}
```
jaypipes, line 193: I wonder if we'll need a `MarkUnadopted` ever? Probably not, and the user can manually remove the annotation if they ever really want to....

jaypipes (changes requested):
couple requested changes inline, and this will need a rebase after the sorted printer column fields stuff merges, but this is almost over the finish line IMHO.

pkg/generate/config/resource.go
RedbackThomson: I would believe (and that's based on naivety) that the majority of AWS resources are able to be adopted without any issue. Setting the default to `false` would require each service team to manually enable each of their resources - the majority of which would work without any generator modifications. For the sake of compatibility, enabling adoption for as many resources as possible would be preferable to customers hoping to adopt in the future.

jaypipes:
@RedbackThomson mind `git rebase`'ing this to squash into a single commit please?

pkg/generate/code/set_sdk.go
```diff
@@ -223,7 +223,7 @@ func SetSDK(
 			out += fmt.Sprintf(
 				"%s} else {\n", indent,
 			)
-			nameField := r.NameField()
+			nameField := *r.SpecIdentifierField()
```
jaypipes, line 226: So I just noticed this... are you relying on the fact that the above will panic if the `SpecIdentifierField()` method returns `nil` for a resource?

jaypipes: No, just wanted to double-check that this was your intention :) I'm totally fine with it

jaypipes (approved):
:+1:

A-Hilaly (approved):
LGTM :+1:

PR #17, opened 2021-02-16
crossplane: code generation for IsUpToDate function

muvaf:
Description of changes: In Crossplane, we run a comparison of current and desired state before issuing an `Update` call. This PR automatically calculates the result for matching fields between the output of `Read` operation and the desired state in spec to see whether any update is due. It checks only the fields that appear in `Update` call, if exists.

Fixes https://github.com/crossplane/provider-aws/issues/490

### DIFF
--- a/pkg/generate/code/isuptodate.go
+++ b/pkg/generate/code/isuptodate.go
@@ -1,0 +1,201 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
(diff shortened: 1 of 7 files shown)


muvaf:
Tested this with a `ReadOne` example that has `Update` (DynamoDB Table) and a `ReadMany` example (RDS DBCluster) that has `Update`. It works as expected.

jaypipes:
@muvaf With our [refactoring proposals](https://github.com/aws-controllers-k8s/community/pull/1100), I'm hoping that ACK's code generator will become more generic and more useful for Crossplane. I'm keeping this PR open (even though it's old and will definitely need a rebase) as a reminder to the ACK core team why extensibility and abstraction in our code generator is super important.

PR #18, opened 2021-02-16
crossplane: add postDelete hook

muvaf:
Description of changes: `Secret` resource from `SecretsManager` has special logic around deletion and we need to work on the response of the deletion call, hence `postDelete` hook.


jaypipes (approved):
:+1:

PR #20, opened 2021-02-22
link ack-runtime with code-generator via go.mod

jaypipes:
Even though Go code in the code-generator doesn't actually import the
ack-runtime library (github.com/aws-controllers-k8s/runtime), the
templates contained in the code-generator *do* produce Go code that
imports the ack-runtime. So, in order to fake out a Go module dependency
between the code-generator and the ack-runtime, I've added a simple test
file in pkg/generate/ack/runtime_test.go that effectively couples the
ack-runtime library at v0.0.3.

When we make changes in the ack-runtime that break backwards
compatibility, we can now update the SemVer in the ack-runtime repo and
the commits to the code-generator that change the produced Go code that
follows the new interfaces in ack-runtime can update the depended-on
ack-runtime version in the code-generator repository's go.mod file.


go.mod
```diff
@@ -14,5 +15,5 @@ require (
 	github.com/spf13/cobra v1.1.1
 	github.com/stretchr/testify v1.7.0
 	golang.org/x/mod v0.4.1
-	k8s.io/apimachinery v0.20.2
+	k8s.io/apimachinery v0.18.6
```
jaypipes, line 18: Note to reviewers: this was necessary in order to avoid this issue: https://github.com/kubernetes/client-go/issues/741

A-Hilaly (approved):
LGTM, let's 🚢 it!

surajkota (approved):
tested locally, lgtm!

Thanks

PR #21, opened 2021-02-23
SecretKeyRef - CodeGen changes

nmvk:
- Introduced a new field config for secrets.
- This will turn a given field into SecretKeyRef.
- Initial approach to solve ElastiCache
    AuthToken usecase.

Other PR -
Runtime changes -
ElastiCache generated code - 

### DIFF
--- a/pkg/generate/code/set_sdk.go
+++ b/pkg/generate/code/set_sdk.go
@@ -165,7 +165,7 @@ func SetSDK(
 	}
 
 	opConfig, override := cfg.OverrideValues(op.Name)
-
+	fieldConfigs := cfg.ResourceFields(r.Names.Original)
 	for memberIndex, memberName := range inputShape.MemberNames() {
 		if r.UnpacksAttributesMap() && memberName == "Attributes" {
 			continue
@@ -188,6 +188,22 @@ func SetSDK(
 				out += fmt.Sprintf("%s%s.Set%s(%s)\n", indent, targetVarName, memberName, value)
 				continue
 			}
+		}
+
+		fc, ok := fieldConfigs[memberName]
+		if ok && fc.IsSecret {
+			out += fmt.Sprintf("%sif %s.Spec.%s != nil {\n", indent, sourceVarName, memberName)
+			out += fmt.Sprintf("%s%stmpSecret, err := rm.rr.SecretValueFromReference(ctx, %s.Spec.%s)\n", indent,
+				indent, sourceVarName, memberName)
+			out += fmt.Sprintf("%s%sif err != nil {\n", indent, indent)
+			out += fmt.Sprintf("%s%s%sreturn nil, err\n", indent, indent, indent)
+			out += fmt.Sprintf("%s%s}\n", indent, indent)
+			out += fmt.Sprintf("%s%sif tmpSecret != \"\" {\n", indent, indent)
+			out += fmt.Sprintf("%s%s%s%s.Set%s(%s)\n", indent, indent, indent,
+				targetVarName, memberName, "tmpSecret")
+			out += fmt.Sprintf("%s%s}\n", indent, indent)
+			out += fmt.Sprintf("%s}\n", indent)
+			continue
 		}
 
 		if r.IsPrimaryARNField(memberName) {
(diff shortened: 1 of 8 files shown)


pkg/generate/config/field.go
```diff
@@ -139,4 +139,7 @@ type FieldConfig struct {
 	// From instructs the code generator that the value of the field should
 	// be retrieved from the specified operation and member path
 	From *SourceFieldConfig `json:"from,omitempty"`
+	// IsSecret instructs the code generator that this field should be a
+	// SecretKeyReference.
+	IsSecret bool `json:"is_secret"`
```
jaypipes, line 144: Please place this with the other boolean "IsXXX" fields above :)

pkg/model/field.go
```diff
@@ -58,7 +57,14 @@ func newField(
 	if shapeRef != nil {
 		shape = shapeRef.Shape
 	}
-	if shape != nil {
+
+	if cfg != nil && cfg.IsSecret {
+		gt = "*acktypes.SecretKeyReference"
+		gte = "acktypes.SecretKeyReference"
+		gtwp = "*acktypes.SecretKeyReference"
```
jaypipes, line 64: This should be `ackv1alpha` not `acktypes` (see my comment on the runtime PR about putting SecretKeyReference into `apis/core/v1alpha1`)

pkg/model/field.go
```diff
@@ -58,7 +57,14 @@ func newField(
...
+
+		crd.AddTypeImport("github.com/aws-controllers-k8s/runtime/pkg/types", "acktypes")
```
jaypipes, line 66: No need for the above once you change to `ackv1alpha1`. See here:

https://github.com/aws-controllers-k8s/code-generator/blob/fe3ed9d62dbca1230c3850d8149c8da7fe15eddb/templates/apis/crd.go.tpl#L12

jaypipes (changes requested):

pkg/generate/code/set_sdk.go
```diff
@@ -190,6 +190,22 @@ func SetSDK(
 			}
 		}
 
+		fc, ok := fieldConfigs[memberName]
+		if ok && fc.IsSecret {
+			out += fmt.Sprintf("%sif %s.Spec.%s != nil {\n", indent, sourceVarName, memberName)
+			out += fmt.Sprintf("%s%sresp, err := rm.rr.SecretValueFromReference(ctx, %s.Spec.%s)\n", indent,
```
jaypipes, line 196: Please name the temporary variable containing the secret value something other than `resp`. The name `resp` is already used as the variable name for most HTTP API client call response objects. How about something like `tmpSecret` or something like that?

pkg/generate/code/set_sdk.go
```diff
@@ -190,6 +190,22 @@ func SetSDK(
...
+				indent, sourceVarName, memberName)
+			out += fmt.Sprintf("%s%sif err != nil {\n", indent, indent)
+			out += fmt.Sprintf("%s%s%sreturn nil, err\n", indent, indent, indent)
+			out += fmt.Sprintf("%s%s}\n", indent, indent)
+			out += fmt.Sprintf("%s%sif resp != nil {\n", indent, indent)
+			out += fmt.Sprintf("%s%s%s%s.Set%s(%s)\n", indent, indent, indent,
```
jaypipes, line 202: see my comment on the runtime PR, but this will need to change to pass the returned secret value as-is instead of dereferencing a string pointer.

pkg/generate/code/set_sdk.go
```diff
@@ -190,6 +190,22 @@ func SetSDK(
 			}
 		}
 
+		fc, ok := fieldConfigs[memberName]
+		if ok && fc.IsSecret {
```
jaypipes, line 194: something to address in a future PR, but it would be nice to have a code comment here similar to below on line 210 that shows what the generated Go code looks like from this section.

jaypipes (approved):
:+1: nice work on this @nmvk!

kumargauravsharma (approved):

PR #22, opened 2021-02-23
Add OLM Bundle Generator Workflow

komish:
This PR adds an additional generator which lays down assets that inform the generation of an Operator Lifecycle Manager bundle.

Fixes: https://github.com/aws-controllers-k8s/community/issues/667

## Summary

- Add the `olm` subcommand
  - `cmd/ack-generate/command/olm.go`
- Add templates and copy targets
  - `   templates/Makefile.tpl`
  - `templates/config/controller/olm-kustomization.yaml.tpl`
  - `templates/config/controller/user-env.yaml.tpl`
  - `templates/config/manifests/bases/clusterserviceversion.yaml.tpl`
  - `templates/config/manifests/kustomization.yaml`
  - `templates/config/samples/kustomization.yaml`
  - `templates/config/samples/samples.yaml.tpl`
  - `templates/config/scorecard/bases/config.yaml`
  - `templates/config/scorecard/kustomization.yaml`
  - `templates/config/scorecard/patches/basic.config.yaml`
  - `templates/config/scorecard/patches/olm.config.yaml`

- Define models for olm metadata
  - `pkg/model/olm.go`

- Add logic to handle template rendering for olm bundle assets
  - `pkg/generate/olm/olm.go`

* Adjust `controller` generator to include an additional template for the kustomization.yaml. This allows the `olm` generator to also target the same base.
  * `templates/config/controller/kustomization.yaml.tpl`
  * `pkg/generate/ack/controller.go`
  * `templates/config/controller/kustomization_def.yaml.tpl`

## Implementation

Once all previous required generators have been executed (e.g. what happens when build-controller.sh is run), a user can then generate OLM bundle assets by running the new generator

```
ack-generate olm <service> <version>
```

OLM bundles are representations of the controller at a specific version, so the `<version>` positional argument is considered required.

A service controller configuration can be passed in as an input, which adjusts the resulting assets to have values appropriate for the service controller. Here's a minimal example for S3.

```yaml
---
annotations:
  capabilityLevel: basic install
  shortDescription: AWS S3 Controller is a Service Controller for Managing AWS S3 Buckets
    in Kubernetes
displayName: AWS Simple Storage Service (S3) for Kubernetes
projectLongDescription: Manage AWS S3 Buckets from within your Kubernetes cluster.
  This controller is a component of the [AWS Controller for Kubernetes](https://github.com/aws/aws-controllers-k8s)
  project. This project is currently in **developer preview**.
samples:
- kind: Bucket
  spec: |
    name: "example-bucket-created-by-ack"
- kind: MultipartUpload
  spec: |
    bucket: "example-bucket-created-by-ack"
      key: "some-key"
maintainers:
- name: "The ACK Team"
  email: "ack@example.com"
```

Users can pass in the path to this configuration using the `--olm-config` option. Alternatively, the path is assumed to be in the format `<service>-olmconfig.yaml` (e.g. `s3-olmconfig.yaml`) at the root of the project. More information on some of the keys and values that are configurable can be seen [here](https://github.com/operator-framework/community-operators/blob/master/docs/packaging-required-fields.md).

Because `code-generator` is a common project for various service controllers, the `olm` generator includes a few global metadata items in the resulting assets. Specifically, these are `Keywords` and `Links` which helps with discoverability and documentation when consuming the controller using an OLM cluster with the console installed. Common keywords and links are injected by default, but can be disabled with the `--no-common-keywords` and `--no-common-links` flags, respectively.

_EDIT: The makefile is no longer included, in favor of using build scripts to optionally perform the same tasks. Below snippets are preserved for contextual purposes._
~Once `ack-generate olm s3 <version>` is executed, the necessary files to generate a bundle should be in place at the output path. A generated Makefile has targets necessary for the next steps. The following are executed from within the output path for the service:~

```shell
make bundle # create the OLM bundle
make BUNDLE_IMG=/your/bundle/image:tag bundle-build # created the container image containing the bundle
docker push <your newly created bundle> # push to a container registry of your choosing
```

~This can then be tested using `operator-sdk` in a cluster using the `operator-sdk run bundle` command.~

## Follow Up

* Each service needs to have the OLM input metadata stored somewhere. Where should it be stored?
* The generation logic will drop a Makefile with a subset of targets that kubebuilder/operator-sdk projects scaffold. It's created at the root of a controller's directory, and I see existing Makefile targets there in the repositories posted here on GitHub (although they don't seem to come from the generator). How should we handle the conflict?
* With the separation of service controllers from the code generation and other common logic, I saw an issue generating the s3-controller in testing because the generated controller arguments were not compatible with the published versions of the controller image with the s3 tag (specifically, the `--resource-tags` flag). Is there some kind of correlation between the code-generator logic and the desired output such that we can generate the proper OLM bundle assets for the existing published controllers? I believe this has a workaround, but I figured it was worth asking in case I'm missing something obvious. 

Let me know if I can make adjustments to things!


cmd/ack-generate/command/olm.go
```diff
@@ -0,0 +1,167 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/ghodss/yaml"
+
+	generate "github.com/aws-controllers-k8s/code-generator/pkg/generate"
+	ackgenerate "github.com/aws-controllers-k8s/code-generator/pkg/generate/ack"
+	olmgenerate "github.com/aws-controllers-k8s/code-generator/pkg/generate/olm"
+	ackmodel "github.com/aws-controllers-k8s/code-generator/pkg/model"
+	"github.com/spf13/cobra"
```
jaypipes, line 28: femto-nit: I generally keep the third-party imports together separated by a newline and then the "local" imports, so sweep this cobra import up under the ghodss/yaml import line :)

templates/config/manifests/bases/clusterserviceversion.yaml.tpl
```diff
@@ -0,0 +1,73 @@
...
+    alm-examples: '[]'
+    capabilities: {{.Annotations.CapabilityLevel}}
+    operatorframework.io/suggested-namespace: "ack-system"
+    repository: {{.Annotations.Repository}}
+    containerImage: {{.Annotations.ContainerImage}}:{{.ServiceIDClean}}-v{{.Version}}
+    description: {{.Annotations.ShortDescription}}
+    createdAt: {{.CreatedAt}}
+    support: {{.Annotations.Support}}
+    certified: {{.Annotations.IsCertified}}
+  name: ack-{{.ServiceIDClean }}-controller.v0.0.0
```
jaypipes, line 15: is the `v0.0.0` suffix hard-coded here or should it be templatized like above on line 10?

templates/Makefile.tpl
```diff
@@ -0,0 +1,53 @@
...
+.PHONY: bundle
+bundle: kustomize
+	cd config/controller && $(KUSTOMIZE) edit set image controller=$(IMG)
+	$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS) --package ack-{{ .ServiceIDClean }}-controller
+	operator-sdk bundle validate ./bundle
+
+# Build the bundle image.
+.PHONY: bundle-build
+bundle-build:
+	docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
```
jaypipes, line 53: Is the above Makefile necessary for OLM to work? Looking at the above, I'd prefer to modify the `scripts/build-controller-release.sh` script in the main community source repository to optionally call `operator-sdk generate bundle` instead of generating a templatized Makefile like this...

pkg/model/olm.go
```diff
@@ -0,0 +1,162 @@
+package model
```
jaypipes, line 1: This file seems odd. Is there not an `import`-able OLM configuration struct? If not, I'd prefer we place all of this into the `pkg/generate/olm` package instead of the `pkg/model` package which is really for modeling of the AWS APIs. :)

templates/config/manifests/bases/clusterserviceversion.yaml.tpl
```diff
@@ -0,0 +1,73 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  annotations:
+    category: "Cloud Provider"
+    alm-examples: '[]'
```
jaypipes, line 6: this isn't a typo is it? "alm" not "olm"?

jaypipes (commented):
Hi @komish! Great start on this, really appreciate it! Some questions and comments inline :)

cmd/ack-generate/command/olm.go
```diff
@@ -0,0 +1,167 @@
...
+var optDisableCommonKeywords bool
+
+// olmCmd is the command that generates a service ClusterServiceVersion base
+// for generating an operator lifecycle manager bundle.
+var olmCmd = &cobra.Command{
+	Use: "olm <service> <version>",
+	Short: `Generate an Operator Lifecycle Manager's ClusterServiceVersion
+resources for AWS service API. Expects a configuration file to be passed in
+via the --olm-config option, or otherwise reads the config at the root of the
+project with the filename <servicealias>-olmconfig.yaml`,
```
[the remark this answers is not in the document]
komish: Yes, definitely! I haven't tested running `ack-generate` from a directory outside of the `code-generator` repository, so the "default" behavior is to look for the file at the base of `code-generator` if nothing is passed in via the `--olm-config` flag. In practice, the olm-config file is in a weird place because it's not generated and always needs to be read from some external source. It's similar to the API definitions that are pulled in from aws-sdk-go in my mind. On the other hand, it makes logical sense to store alongside the service controller.

As an aside - In theory, I could have looked for it in the default location for output files (which I believe is `$svcAlias/`, but the file is _never_ generated, so it didn't make sense to me to look for it at that location. 

TL;DR - Yes! The plan was to include it in service controller repositories. The minimal configuration, in theory, should mean it should need infrequent updates.

templates/config/manifests/bases/clusterserviceversion.yaml.tpl
komish: I'm not sure that it _needs_ to be `v0.0.0` in order to be parsed by operator-sdk, but operator-sdk does replace this with inputs that we pass it when creating the bundle. To that end, the Makefile's target `bundle` has our target version in place - and operator-sdk does the substitution.

Operator-SDK does **not** substitute the annotation value in line 10 as of this moment, so it's templated.

templates/config/manifests/bases/clusterserviceversion.yaml.tpl
komish: Correct - this is not a typo: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/building-your-csv.md#crd-templates

templates/Makefile.tpl
komish: Sure thing, it's not required - just a nicety. I'll have to get more familiar with the script but given that it's a separate repo, I can work on a separate PR there. I'll drop the Makefile and submit another commit shortly.

pkg/model/olm.go
komish: There definitely is an importable - https://github.com/operator-framework/api/blob/master/pkg/operators/v1alpha1/clusterserviceversion_types.go#L268

Since we're not generating a full CSV (and are instead generating a base) we're only using a subset of the keys. With that said, I can reconfigure this to use the importable struct instead if that's preferred? Should work fine. Let me know!

cmd/ack-generate/command/olm.go
komish: Fixed!

templates/Makefile.tpl
komish: The Makefile template has been removed and is no longer part of the rendering logic

pkg/model/olm.go
komish: @jaypipes here's what it could look like with the imported `ClusterServiceVersionSpec` as a PoC. A few fields had to be reconfigured in the templates to accommodate, as well as subtle changes in the user-provider service config. Overall, the workflow stays the same.  

https://github.com/komish/code-generator/commit/0b8df4c3cfde35ca552d1abfa1a5f2e2dbb48b33

Samples and Annotations aren't strictly part of the representation, so I still found myself having to define those. Otherwise this drops the model in favor of the upstream ClusterServiceVersionSpec and a few of those previously defined types - all now stored in the `pkg/generate/olm` package.

From the service controller maintainer perspective, it might make sense to relabel "Annotations" here to something else  given that it's a limited set of acceptable keys. Disambiguation might be helpful so it isn't perceived as passing through arbitrary keys/values. I'm on the fence. Could be swayed either way.

If this looks more like what you expect, I'll work on collapsing that branch into this PR.

templates/Makefile.tpl
komish: I've started playing with https://github.com/aws-controllers-k8s/community/blob/main/scripts/build-controller-release.sh and am wondering if we should recreate all targets there or put them in a separate script. I'm also thinking the calls out to build the image using docker can probably be left to "Documentation".

I've put together the `ack-generate olm` into that script and that makes sense to me, but doing the bundle generation (i.e. targets that lay down bundle/ and bundle.Dockerfile` requires OperatorSDK. If it's optional, we don't want to require OperatorSDK be installed for folks who are not interested. Don't want to cause extra work for those folks.

As a side note, should I go ahead and start an issue in community so we can discuss this separately?

cmd/ack-generate/command/olm.go
```diff
@@ -0,0 +1,167 @@
...
+			"please specify the service alias and version " +
+				"for the AWS service API to generate",
+		)
+	}
+	svcAlias := strings.ToLower(args[0])
+	if optOutputPath == "" {
+		optOutputPath = filepath.Join(optServicesDir, svcAlias)
+	}
+
+	version := args[1]
```
komish, line 81: After working with the `build-controller-release.sh` script, I'm wondering if I should accept "v0.0.1" instead of "0.0.0" here. The script accepts the leading `v`, this expects just the numbers. Small nitpick on my end. I generally prefer just the numbers but also want to keep things consistent.

templates/Makefile.tpl
jaypipes: > I've started playing with https://github.com/aws-controllers-k8s/community/blob/main/scripts/build-controller-release.sh and am wondering if we should recreate all targets there or put them in a separate script. I'm also thinking the calls out to build the image using docker can probably be left to "Documentation".
> 
> I've put together the `ack-generate olm` into that script and that makes sense to me, but doing the bundle generation (i.e. targets that lay down bundle/ and bundle.Dockerfile` requires OperatorSDK. If it's optional, we don't want to require OperatorSDK be installed for folks who are not interested. Don't want to cause extra work for those folks.
> 
> As a side note, should I go ahead and start an issue in community so we can discuss this separately?

@komish yes, please do open a new issue for this topic. sorry about delay in reviewing this. Will try to get a full review done today!

cmd/ack-generate/command/olm.go
jaypipes: Up to you, @komish. We could also (long-term, not needed in this PR) use https://github.com/Masterminds/semver to handle this kind of common stuff.

pkg/model/olm.go
jaypipes: @komish yeah, let's definitely go with komish@0b8df4c approach. I like that much better than adding to `pkg/model`

jaypipes: > From the service controller maintainer perspective, it might make sense to relabel "Annotations" here to something else given that it's a limited set of acceptable keys. Disambiguation might be helpful so it isn't perceived as passing through arbitrary keys/values. I'm on the fence. Could be swayed either way.

Meh, I think `OLMMetadata.Annotations` is fine and appropriately confers that this is annotation data specific to OLM needs. But, if you want to have a more explicit name, go for it :)

jaypipes (changes requested):
@komish cool, let's go with you komish@0b8df4c code for this. Merge that branch into here and I'm good to merge away. :)

templates/Makefile.tpl
komish: I'll get an issue created and CC you in the coming week. Good to resolve here when you are.

komish:
@jaypipes Merged the alternate branch. Thanks for your help with this! 

Quick note that I did a `go mod tidy` before merging and added the changes. It bumped APIMachinery a bit and I wanted to make sure to draw your attention to it.

jaypipes (approved):
:+1: great work, @komish! happy to see this get into the code-generator.

jaypipes:
> Quick note that I did a `go mod tidy` before merging and added the changes. It bumped APIMachinery a bit and I wanted to make sure to draw your attention to it.

Yep, saw that. All good.

PR #25, opened 2021-02-26
crossplane: fix controller template for resources that might have both ReadOne and ReadMany

muvaf:
Description of changes: There are cases like `Lambda Function` where there are both ReadMany and ReadOne operations available. The first choice is always ReadOne, so we need to take that into account here as well.


jaypipes (approved):
:+1:

PR #26, opened 2021-03-02
add support for generic callback hooks

jaypipes:
This patch adds support for a generic callback system into the generator
config that allows controller implementors to specify some code that
should be injected at specific named points in a template. I expect that
eventually this generic hook system will be more useful, flexible and
extensible than the hodge-podge of custom callback methods and overrides
currently in the generator config.

The `pkg/generate/config.ResourceConfig` struct now has a `Hooks` field
of type `map[string]*HookConfig`, with the map keys being named hook
points, e.g. "sdk_update_pre_build_request".

There are two ways to inject code at hook points: inline and via a
template path. The inline method uses the `HookConfig.Code` field which
should contain the Go code that gets injected at a named hook point. The
`HookConfig.TemplatePath` field is used to refer to a template file at a
specific path. The template file is searched for in any of the
TemplateSet's base template paths.

Here's an example of a generator config snippet that uses the inline
code injection method (`HookConfig.Code`) to add a piece of custom code
to be executed in the sdk_update.go.tpl right before the code in the
resource manager's `sdkUpdate` method calls the
`newUpdateRequestPayload()` function:

```yaml
resources:
  Broker:
    hooks:
      sdk_update_pre_build_request:
        code: if err := rm.requeueIfNotRunning(latest); err != nil { return nil, err }
```

Here is the snippet from the templates/pkg/resource/sdk_update.go.tpl
file that shows how we can add these generic named hooks into our
templates at various places:

```
{{- if $hookCode := .CRD.HookCode sdk_update_pre_build_request }}
{{ $hookCode }}
{{ end -}}
```

The controller implementor need only implement the little
`requeueIfNotRunning` function, with the function signature described
in the generator config code block. We no longer need to have function
signatures match for custom callback code since the function signature
for callback code is up to the dev writing the generator.yaml config
file.

Here is an example of the `HookConfig.TemplatePath` being used to refer
to a template file containing some code that is injected at a named hook
point:

```yaml
resources:
  Broker:
    hooks:
      sdk_update_pre_build_request:
        template_path: sdk_update_pre_build_request.go.tpl
```

A controller implementor would simply need to populate a
`sdk_update_pre_build_request.go.tpl` file with the code to be included
at the hook point.

This patch introduces the following hook points in the ACK controller
resource manager code paths:

* sdk_read_one_pre_build_request
* sdk_read_many_pre_build_request
* sdk_get_attributes_pre_build_request
* sdk_create_pre_build_request
* sdk_update_pre_build_request
* sdk_delete_pre_build_request
* sdk_read_one_post_request
* sdk_read_many_post_request
* sdk_get_attributes_post_request
* sdk_create_post_request
* sdk_update_post_request
* sdk_delete_post_request
* sdk_read_one_pre_set_output
* sdk_read_many_pre_set_output
* sdk_get_attributes_pre_set_output
* sdk_create_pre_set_output
* sdk_update_pre_set_output

The "pre_build_request" hooks are called BEFORE the call to construct
the Input shape that is used in the API operation and therefore BEFORE
any call to validate that Input shape.

The "post_request" hooks are called IMMEDIATELY AFTER the API operation
aws-sdk-go client call.  These hooks will have access to a Go variable
named `resp` that refers to the aws-sdk-go client response and a Go
variable named `respErr` that refers to any error returned from the
aws-sdk-go client call.

The "pre_set_output" hooks are called BEFORE the code that processes the
Outputshape (the pkg/generate/code.SetOutput function). These hooks will
have access to a Go variable named `ko` that represents the concrete
Kubernetes CR object that will be returned from the main method
(sdkFind, sdkCreate, etc). This `ko` variable will have been defined
immediately before the "pre_set_output" hooks as a copy of the resource
that is supplied to the main method, like so:

```go
	// Merge in the information we read from the API call above to the copy of
	// the original Kubernetes object we passed to the function
	ko := r.ko.DeepCopy()
```

### DIFF
--- a/pkg/generate/ack/hook.go
+++ b/pkg/generate/ack/hook.go
@@ -1,0 +1,145 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package ack
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	ttpl "text/template"
+
+	ackmodel "github.com/aws-controllers-k8s/code-generator/pkg/model"
+	ackutil "github.com/aws-controllers-k8s/code-generator/pkg/util"
+)
+
+/*
+The following hook points are supported in the ACK controller resource manager
+code paths:
+
+* sdk_read_one_pre_build_request
+* sdk_read_many_pre_build_request
+* sdk_get_attributes_pre_build_request
+* sdk_create_pre_build_request
+* sdk_update_pre_build_request
+* sdk_delete_pre_build_request
+* sdk_read_one_post_request
+* sdk_read_many_post_request
+* sdk_get_attributes_post_request
+* sdk_create_post_request
+* sdk_update_post_request
+* sdk_delete_post_request
+* sdk_read_one_pre_set_output
+* sdk_read_many_pre_set_output
+* sdk_get_attributes_pre_set_output
+* sdk_create_pre_set_output
+* sdk_update_pre_set_output
+
+The "pre_build_request" hooks are called BEFORE the call to construct
+the Input shape that is used in the API operation and therefore BEFORE
+any call to validate that Input shape.
+
+The "post_request" hooks are called IMMEDIATELY AFTER the API operation
+aws-sdk-go client call.  These hooks will have access to a Go variable
+named `resp` that refers to the aws-sdk-go client response and a Go
+variable named `respErr` that refers to any error returned from the
+aws-sdk-go client call.
+
+The "pre_set_output" hooks are called BEFORE the code that processes the
+Outputshape (the pkg/generate/code.SetOutput function). These hooks will
+have access to a Go variable named `ko` that represents the concrete
+Kubernetes CR object that will be returned from the main method
+(sdkFind, sdkCreate, etc). This `ko` variable will have been defined
+immediately before the "pre_set_output" hooks as a copy of the resource
+that is supplied to the main method, like so:
+
+```go
+	// Merge in the information we read from the API call above to the copy of
+	// the original Kubernetes object we passed to the function
+	ko := r.ko.DeepCopy()
+```
+*/
+
+// ResourceHookCode returns a string with custom callback code for a resource
+// and hook identifier
+func ResourceHookCode(
+	templateBasePaths []string,
+	r *ackmodel.CRD,
+	hookID string,
+) (string, error) {
+	resourceName := r.Names.Original
+	if resourceName == "" || hookID == "" {
+		return "", nil
+	}
+	c := r.Config()
+	if c == nil {
+		return "", nil
+	}
+	rConfig, ok := c.Resources[resourceName]
+	if !ok {
+		return "", nil
+	}
+	hook, ok := rConfig.Hooks[hookID]
+	if !ok {
+		return "", nil
+	}
+	if hook.Code != nil {
+		return *hook.Code, nil
+	}
+	if hook.TemplatePath == nil {
+		err := fmt.Errorf(
+			"resource %s hook config for %s is invalid. Need either code or template_path",
+			resourceName, hookID,
+		)
+		return "", err
+	}
+	for _, basePath := range templateBasePaths {
+		tplPath := filepath.Join(basePath, *hook.TemplatePath)
+		if !ackutil.FileExists(tplPath) {
+			continue
+		}
+		tplContents, err := ioutil.ReadFile(tplPath)
+		if err != nil {
+			err := fmt.Errorf(
+				"resource %s hook config for %s is invalid: error reading %s: %s",
+				resourceName, hookID, tplPath, err,
+			)
+			return "", err
+		}
+		t := ttpl.New(tplPath)
+		if t, err = t.Parse(string(tplContents)); err != nil {
+			err := fmt.Errorf(
+				"resource %s hook config for %s is invalid: error parsing %s: %s",
+				resourceName, hookID, tplPath, err,
+			)
+			return "", err
+		}
+		var b bytes.Buffer
+		// TODO(jaypipes): Instead of nil for template vars here, maybe pass in
+		// a struct of variables?
+		if err := t.Execute(&b, nil); err != nil {
+			err := fmt.Errorf(
+				"resource %s hook config for %s is invalid: error executing %s: %s",
+				resourceName, hookID, tplPath, err,
+			)
+			return "", err
+		}
+		return b.String(), nil
+	}
+	err := fmt.Errorf(
+		"resource %s hook config for %s is invalid: template_path %s not found",
+		resourceName, hookID, *hook.TemplatePath,
+	)
+	return "", err
+}
--- a/pkg/generate/config/resource.go
+++ b/pkg/generate/config/resource.go
@@ -29,7 +29,10 @@ type ResourceConfig struct {
 	// Found and other common error types for primary resources, and thus we
 	// need these instructions.
 	Exceptions *ExceptionsConfig `json:"exceptions,omitempty"`
-
+	// Hooks is a map, keyed by the hook identifier, of instructions for the
+	// the code generator about a custom callback hooks that should be injected
+	// into the resource's manager or SDK binding code.
+	Hooks map[string]*HooksConfig `json:"hooks"`
 	// Renames identifies fields in Operations that should be renamed.
 	Renames *RenamesConfig `json:"renames,omitempty"`
 	// ListOperation contains instructions for the code generator to generate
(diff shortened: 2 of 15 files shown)


templates/pkg/resource/sdk_update.go.tpl
```diff
@@ -5,6 +5,9 @@ func (rm *resourceManager) sdkUpdate(
 	latest *resource,
 	diffReporter *ackcompare.Reporter,
 ) (*resource, error) {
+{{- if $hookCode := .CRD.HookCode sdk_update_pre_build_request }}
+{{ $hookCode }}
+{{ end -}}
```
jaypipes, line 10: Obviously I've only added in a single `sdk_update_pre_build_request` hook here, into the `sdk_update.go.tpl` template. If folks this is a good strategy, we can go and add other hook points in the other templates.

vijtrip2 (approved):
I like it. 🚀 

-------

My only concern from first look was what if someone takes this hook too literally and writes bulky code blocks in generator config, that will make generator config less readable.
But since they control the signature of hook function and implementation too, this code block should ideally be a single function call encapsulating all the details. : )

jaypipes:
> I like it. rocket
> 
> My only concern from first look was what if someone takes this hook too literally and writes bulky code blocks in generator config, that will make generator config less readable.
> But since they control the signature of hook function and implementation too, this code block should ideally be a single function call encapsulating all the details. : )

Yeah, agreed. Plus, putting large blocks of code in YAML is actually annoying and fraught with peril, so it kind of self-selects that out ;)

templates/pkg/resource/sdk_update.go.tpl
panazzo: I believe we support x-ray distributed tracing, correct? If so, would we propagate the tracing to the hook?

A-Hilaly: > I believe we support x-ray distributed tracing, correct? If so, would we propagate the tracing to the hook?

I think we only have Prometheus metrics for now. Regarding tracing I have two questions:
- what if containers are running outside of AWS? It is still possible to propagate tracing?
- if yes, what if customers want to use a different solution of tracing (jager ....etc)?

A-Hilaly:
LGTM! 

i'm wondering if we should offer a set of hooks to use, instead of allowing users to pass plain code. something like:
```yaml
resources:
  Broker:
    hooks:
      sdk_update_pre_build_request:
        requeueIfNotRunning: true
```
or 

```yaml
resources:
  Broker:
    hooks:
      sdk_update_pre_build_request:
        - requeueIfNotRunning
        - anotherHook
```
wdyt @vijtrip2 @jaypipes ?

jaypipes:
> LGTM!
> 
> i'm wondering if we should offer a set of hooks to use, instead of allowing users to pass plain code. something like:
> 
> ```yaml
> resources:
>   Broker:
>     hooks:
>       sdk_update_pre_build_request:
>         requeueIfNotRunning: true
> ```
> 
> or
> 
> ```yaml
> resources:
>   Broker:
>     hooks:
>       sdk_update_pre_build_request:
>         - requeueIfNotRunning
>         - anotherHook
> ```
> 
> wdyt @vijtrip2 @jaypipes ?

We can think about that at some later time, yes.

templates/pkg/resource/sdk_update.go.tpl
jaypipes: > I believe we support x-ray distributed tracing, correct? If so, would we propagate the tracing to the hook?

@panazzo this is for injecting custom code into known places in the various templates. If we wanted to support x-ray tracing, we'd probably want to do that in a standardized way -- similar to how we currently have a Prometheus standard way of tracking API calls:

https://github.com/aws-controllers-k8s/runtime/blob/cbda0dfbbca384a56c863b3973d8e3c437a46088/pkg/metrics/metrics.go#L63-L90

Happy to have a conversation about x-ray support in a separate github issue!

surajkota:
Don't really like the gocode in yaml file. I can't see how hooks is different from custom_implementation except one doesn't need to match the function signature. Developer still need to implement the custom method

Could you provide more information? Or am I missing something?

I can see a pattern for a status field and value to check for before triggering update or successful reconciling but the name of field and value to check will be different for each service

jaypipes:
> Don't really like the gocode in yaml file. I can't see how hooks is different from custom_implementation except one doesn't need to match the function signature. Developer still need to implement the custom method

I would like to make `custom_implementation` a last resort. The hooks concept is more flexible and able to be fine-tuned so that we reduce the total amount of custom code over time.

I agree that adding Go code in YAML is icky :)

Would you like this PR better if I added a `HookConfig.Template` field that could be the path to a template file that contained the Go code to be injected at the hook point? For instance, you could create a template file named `sagemaker-controller/templates/my_custom_code.go.tpl` and then refer to the code in that template file by doing something like this in the sagemaker generator.yaml file:

```yaml
resources:
  Endpoint:
    hooks:
      sdk_update_pre_build_request:
        template: sagemaker-controller/templates/my_custom_code.go.tpl
```

> Could you provide more information? Or am I missing something?
> 
> I can see a pattern for a status field and value to check for before triggering update or successful reconciling but the name of field and value to check will be different for each service

Right, the reason this needs to be custom code for the immediate future is because each API is different in the way it checks for certain state-transition changes... I would love to get to a point where we can describe these state-transition change strategies in a declarative fashion in the generator.yaml file and have the Go code generated that does the checks, but due to the inconsistencies in the APIs, we are quite far away from that ability at the moment so this is a stop-gap measure.

jaypipes:
@surajkota please see https://github.com/aws-controllers-k8s/code-generator/pull/28 for a patch that enables the idea of having code in template files instead of specified as `HookConfig.Code`

pkg/generate/testdata/models/apis/mq/0000-00-00/generator.yaml
```diff
@@ -0,0 +1,9 @@
+ignore:
+  resources:
+    - Configuration
+    - User
+resources:
+  Broker:
+    hooks:
+      sdk_update_pre_build_request:
```
surajkota, line 8: Could you please add a test for sdk_delete_pre_request as well so I can pick up this change as soon as it's merged?

jaypipes: Yes, I will. I am currently modifying this patch to support `template_path` for hook configurations.

pkg/generate/config/resource.go
```diff
@@ -349,3 +373,29 @@ func (c *Config) ResourceShortNames(resourceName string) []string {
 	}
 	return rConfig.ShortNames
 }
+
+// ResourceHookCode returns a string with custom callback code for a resource
+// and hook identifier
+func (c *Config) ResourceHookCode(resourceName string, hookID string) string {
```
vijtrip2, line 379: is this function redundant now because there is a function with similar name in hook.go file?

vijtrip2 (approved):
LGTM.

pkg/generate/config/resource.go
jaypipes: Doh, nice catch :) Yes, I forgot to remote it. Doing that now...

jaypipes: done.

pkg/generate/testdata/models/apis/mq/0000-00-00/generator.yaml
jaypipes: @surajkota please note I've updated this PR to include support for specifying `template_path` in addition to raw `code`

surajkota (approved):
Thanks, Lgtm!
not sure if it was intentional, template changes are missing in sdk_find_read_many, sdk_find_get etc. 

Will try using this PR today

general question, if one is using `hook.code`, how will the indentation be handled, is it only good for one-liner code?

A-Hilaly (approved):
/lgtm 👍

A-Hilaly:
>general question, if one is using hook.code, how will the indentation be handled, is it only good for one-liner code?

Most indentation problems will be corrected by `gofmt`.

It is still possible to provide multi-line code with the correct indentation using something like:

```yaml
hooks:
  sdk_update_pre_build_request:
  code: |
    if err := rm.requeueIfNotRunning(latest); err != nil {
        return nil, err
    }
```

pkg/generate/ack/hook.go
```diff
@@ -0,0 +1,139 @@
...
+		if t, err = t.Parse(string(tplContents)); err != nil {
+			err := fmt.Errorf(
+				"resource %s hook config for %s is invalid: error parsing %s: %s",
+				resourceName, hookID, tplPath, err,
+			)
+			return "", err
+		}
+		var b bytes.Buffer
+		// TODO(jaypipes): Instead of nil for template vars here, maybe pass in
+		// a struct of variables?
```
RedbackThomson, line 124: I like this idea, but I couldn't come up with any variables that could be useful at compile-time? Possibly version information?

pkg/generate/config/resource.go
```diff
@@ -71,6 +74,27 @@ type ResourceConfig struct {
...
+// HooksConfig instructs the code generator how to inject custom callback hooks
+// at various places in the resource manager and SDK linkage code.
+//
+// Example usage from the AmazonMQ generator config:
+//
+// resources:
+//   Broker:
+//     hooks:
+//		sdk_update_pre_build_update_request:
+//        code: if err := rm.requeueIfNotRunning(latest); err != nil { return nil, err }
```
RedbackThomson, line 86: Github formatting or actually ugly?

pkg/util/file.go
```diff
@@ -0,0 +1,22 @@
...
+// permissions and limitations under the License.
+
+package util
+
+import "os"
+
+// FileExists returns True if the supplied file path exists, false otherwise
+func FileExists(path string) bool {
+	_, err := os.Stat(path)
+	return !os.IsNotExist(err)
```
RedbackThomson, line 21: https://golang.org/pkg/os/#IsExist ? lol

A-Hilaly: Not sure if we can replace `!IsNotExist` with `IsExist`. Because `os.IsExist` and `os.IsNotExist` are error checkers (wrappers of `errors.Is` function) - they both return false if the given error is `nil`.

pkg/generate/ack/hook.go
jaypipes: Yeah, I left it as a TODO because I couldn't think of anything off top of my head. We can always add later if we stumble on something that should be passed to the template...

pkg/generate/config/resource.go
jaypipes: actually ugly :) will correct.

jaypipes:
> Thanks, Lgtm!
> not sure if it was intentional, template changes are missing in `sdk_find_read_many`, `sdk_find_get` etc.

@surajkota nice catch. Added hooks for read_many and get_attributes.

pkg/util/file.go
RedbackThomson: Ah I see, they check error message conditions. How confusing. Thanks!

PR #27, opened 2021-03-06
remove aws/aws-controllers-k8s from code-generator

jaypipes:
The code-generator was still referencing the
github.com/aws-controllers-k8s/runtime@v0.0.2 version which itself referenced still
the original github.com/aws/aws-controllers-k8s library. This was
causing a number of issues for generated controllers, including leaving
old Go module dependencies around.

Also adds a call to `git fetch --all --tags` to the `cloneSDKRepo()`
function in `ack-generate` to ensure that the most up-to-date tags are
pulled for the aws-sdk-go local cached repository. Calling `git checkout
tags/v1.37.4` when the latest cached aws-sdk-go was at 1.35.5 was
causing a frustrating "error: returned 1" garbage to be returned by the
`scripts/build-controller.sh` because the `exec.Cmd()` was burying the
response from Git which was clearly stating no such git tag or file
could be located.


A-Hilaly (approved):
Nice! LGTM

PR #28, opened 2021-03-06
add multiple template base path functionality

jaypipes:
We want different teams to have the ability to specify template
overrides for their specific controller or generated codebase.

There are a number of use cases where having the ability to specify
multiple "template search paths" will be useful.

Imagine that the ElastiCache ACK controller team wants to change their
boilerplate.txt content to mention something specific to ElastiCache.
Instead of adding a bunch of conditional statements to the code
generator or new fields to the generator configuration object (which is
already complicated), what if the maintainer team could add their custom
boilerplate text to a `templates/boilerplate.txt` file in the
elasticache-controller source repository and call `ack-generate`
commands like this:

```bash
CODE_GEN_REPO=$GOPATH/src/github.com/aws-controllers-k8s/code-generator
CONTROLLER_REPO=$GOPATH/src/github.com/aws-controllers-k8s/elasticache-controller
ack-generate apis --template-dirs $CONTROLLER_REPO/templates,$CODE_GEN_REPO/templates`
```

and the code generator would simply pick up the boilerplate.txt file in
the elasticache-controller repository and use that instead of the one in
the code-generator repo.

Another use case might be for a service controller team to experiment
with some Go code changes to the standard template files without needing
to modify any code in the code-generator repo itself.

For example, let's say that the S3 controller requires a radically different
set of Go code for its update code path than all the other controllers
and we want to test some ideas but not update the code-generator repo
itself. We could add a `templates/pkg/resource/sdk_update.go.tpl` file
to the s3-controller repository, rebuild the controller and iterate more
quickly.

Finally, there is the use case of having templates containing "hook
code" that is specific to one service controller. Imagine the following
generator config snippet:

```yaml
resources:
  Broker:
    hooks:
      set_update_pre_build_update_request:
        templatePath: hooks/sdk_update_pre_build_update_request.go.tpl
```

And in the service controller, we had a file
`templates/hooks/sdk_update_pre_build_update_request.go.tpl` with the
contents:

```go
if err := rm.requeueIfNotRunning; err != nil {
        return nil, err
}
```

We could call the `ack-generate` passing in multiple template search
base paths like shown above and in this way, we could have
service-specific code placed inside the service controller source
repositories and have our code generation be much more flexible.

### DIFF
--- a/scripts/build-controller.sh
+++ b/scripts/build-controller.sh
@@ -27,8 +27,6 @@ ACK_GENERATE_BIN_PATH=${ACK_GENERATE_BIN_PATH:-$DEFAULT_ACK_GENERATE_BIN_PATH}
 ACK_GENERATE_API_VERSION=${ACK_GENERATE_API_VERSION:-"v1alpha1"}
 ACK_GENERATE_CONFIG_PATH=${ACK_GENERATE_CONFIG_PATH:-""}
 AWS_SDK_GO_VERSION=${AWS_SDK_GO_VERSION:-""}
-DEFAULT_TEMPLATES_DIR="$ROOT_DIR/templates"
-TEMPLATES_DIR=${TEMPLATES_DIR:-$DEFAULT_TEMPLATES_DIR}
 
 USAGE="
 Usage:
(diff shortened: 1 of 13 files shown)


scripts/build-controller.sh
```diff
@@ -98,6 +96,18 @@ if [[ ! -d $SERVICE_CONTROLLER_SOURCE_PATH ]]; then
     exit 1
 fi
 
+BOILERPLATE_TXT_PATH="$ROOT_DIR/templates/boilerplate.txt"
+DEFAULT_TEMPLATE_DIRS="$ROOT_DIR/templates"
+# If the service controller source repository has a templates/ directory, add
```
vijtrip2, line 101: very minor: Maybe adding a comment regarding the priority of template override here will be helpful. If script run does not produce desired results, the developer first comes to check this script rather than reading the priority related comment in go files.

vijtrip2 (approved):
LGTM!

scripts/build-controller.sh
jaypipes: @vijtrip2 great suggestion. added!

PR #31, opened 2021-03-12
crossplane: fix the controller template for attribute-based APIs

muvaf:
Description of changes: Fixes the controller template for attribute-based APIs

### DIFF
--- a/templates/crossplane/pkg/controller.go.tpl
+++ b/templates/crossplane/pkg/controller.go.tpl
@@ -180,22 +180,30 @@ func newExternal(kube client.Client, client svcsdkapi.{{ .SDKAPIInterfaceTypeNam
 	e := &external{
 		kube:           kube,
 		client:         client,
-		{{- if or .CRD.Ops.ReadOne .CRD.Ops.GetAttributes .CRD.Ops.ReadMany }}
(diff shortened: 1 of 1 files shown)


templates/crossplane/pkg/controller.go.tpl
```diff
@@ -180,22 +180,30 @@ func newExternal(kube client.Client, client svcsdkapi.{{ .SDKAPIInterfaceTypeNam
 	e := &external{
 		kube:           kube,
 		client:         client,
-		{{- if or .CRD.Ops.ReadOne .CRD.Ops.GetAttributes .CRD.Ops.ReadMany }}
+		{{- if or .CRD.Ops.ReadOne }}
```
jaypipes, line 183: I think this needs to be either this:

```
{{- if or .CRD.Ops.ReadOne .CRD.Ops.GetAttributes }}
```

or

```
{{- if .CRD.Ops.ReadOne }}
```

jaypipes (changes requested):

templates/crossplane/pkg/controller.go.tpl
muvaf: Good catch, fixed! Thanks!

jaypipes (approved):
:+1:

PR #34, opened 2021-03-19
Handle nil fields in response from describe API output - readOne and readMany

surajkota:
### Description of changes:

Currently, we create a [copy for the current resource state](https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/pkg/resource/sdk_find_read_one.go.tpl#L29) in sdkFind and merge the output of describe output if it is non-nil. This doesn't work for fields that can toggle between nil and non-nil. The latest resource should match the output from AWS service. This PR addresses this issue

### Testing
with sagemaker controller [aws-controllers-k8s/sagemaker-controller#7](https://github.com/aws-controllers-k8s/sagemaker-controller/pull/7)

### Notes to the reviewer
Change is not implemented for the `ko.Status.ACKResourceMetadata` field because the only case I see where it can toggle is if the user changes the name of resource they created, in which case `sdkFind` will throw an `ackerr.NotFound` error and sdkCreate should overwrite the ARN.

### DIFF
--- a/pkg/generate/code/set_resource.go
+++ b/pkg/generate/code/set_resource.go
@@ -289,6 +289,13 @@ func SetResource(
 			)
 		}
 		out += fmt.Sprintf(
+			"%s} else {\n", indent,
+		)
+		out += fmt.Sprintf(
+			"%s%s%s.%s = nil\n", indent, indent,
+			targetAdaptedVarName, f.Names.Camel,
+		)
+		out += fmt.Sprintf(
 			"%s}\n", indent,
 		)
 	}
@@ -340,6 +347,8 @@ func ListMemberNameInReadManyOutput(
 //      }
 //      if elem.AtRestEncryptionEnabled != nil {
 //          ko.Status.AtRestEncryptionEnabled = elem.AtRestEncryptionEnabled
+//      } else {
+//          ko.Status.AtRestEncryptionEnabled = nil
 //      }
 //      ...
 //      if elem.CacheClusterId != nil {
@@ -349,6 +358,8 @@ func ListMemberNameInReadManyOutput(
 //              }
 //          }
 //          r.ko.Spec.CacheClusterID = elem.CacheClusterId
+//      } else {
+//          r.ko.Spec.CacheClusterID = nil
 //      }
 //      found = true
 //  }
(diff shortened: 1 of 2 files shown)


pkg/generate/code/set_resource.go
```diff
@@ -284,7 +284,7 @@ func SetResource(
 			)
 		}
 		out += fmt.Sprintf(
-			"%s}\n", indent,
+			"%s} else {\n %s.%s = nil \n}\n", indent, targetAdaptedVarName, f.Names.Camel,
```
jaypipes, line 287: Please break into multiple `fmt.Sprintf()` calls, one for each line of output:

```go
out += fmt.Sprintf(
	"%s} else {\n", indent,
)
out += fmt.Sprintf(
	"%s%s%s.%s = nil\n", indent, indent,
	targetAdaptedVarName, f.Names.Camel,
)
out += fmt.Sprintf(
	"%s}\n", indent,
)
```

jaypipes (changes requested):
See inline for a requested change. Also, you will need to modify the unit tests in `pkg/generate`

surajkota:
@jaypipes ready to merge. Please check notes to the reviewer in the description

pkg/generate/code/set_resource.go
```diff
@@ -340,7 +347,9 @@ func ListMemberNameInReadManyOutput(
 //      }
 //      if elem.AtRestEncryptionEnabled != nil {
 //          ko.Status.AtRestEncryptionEnabled = elem.AtRestEncryptionEnabled
-//      }
+//      } else {
+//			ko.Status.AtRestEncryptionEnabled = nil
+//		}
```
jaypipes, line 352: nit, you can fix in a later PR... the indentation above is off, probably because gofmt doesn't replace TAB characters in comments with spaces and the other comments use 4 space indents...

jaypipes (approved):
@surajkota looks good to me. I'd like @vijtrip2 to review this first, though, and just have an extra set of eyes on this part of the generated codebase. Vijay, please double-check that by setting Spac and Status fields to `nil` for empty fields in a Output shape, we would not be overriding any important or defaulted values...

pkg/generate/code/set_resource.go
surajkota: nice catch, easy to forget, fixed it

vijtrip2:
> @surajkota looks good to me. I'd like @vijtrip2 to review this first, though, and just have an extra set of eyes on this part of the generated codebase. Vijay, please double-check that by setting Spac and Status fields to `nil` for empty fields in a Output shape, we would not be overriding any important or defaulted values...

Looks good to me.

* I double checked that We are only updating spec fields for ReadOne, ReadMany and not Create. This was the existing behavior and still being preserved. :+1:
* And In my knowledge of AWS APIs, the ReadOne/ReadMany/CreateOutput result are consistent with CreateInput. If field is being present in CreateInput, it is reflected similarly in ReadOne/CreateOutput. So I am not concerned about overriding default values.

vijtrip2 (approved):
LGTM!

surajkota:
Great! thanks folks

PR #38, opened 2021-03-23
handle renamed fields in SetResource generation

jaypipes:
If fields were renamed in some operations, the code generated for
`SetResource()` was not properly looking up the renamed field. This
patch updates `SetResource()` to first check if the field has 
renamed for the referred Operation and if so, look up the CRDField
object in the CRD.SpecFields collection instead of the StatusFields
collection.


jaypipes:
Note to reviewers: I've included a unit test of this functionality, but also note that the RDS controller has been generated with this code and the DBSubnetGroup resource is properly handling renamed Name and Description fields for that resource.

A-Hilaly (approved):

PR #40, opened 2021-04-05
Adding status reporting for RecoverableCondition

jkuruba:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/498

Description of changes: RecoverableCondition is used to report errors that are not marked as Terminal. Currently controllers sync silently and errors are not reported. Reporting will notify Recoverable errors that can likely be fixed without re-applying the spec. 

Testing: Ran reconciler unit tests
go test -v
=== RUN   TestReconcilerUpdate
2021-04-05T18:51:38.576Z	INFO	ackrt	updated resource	{"kind": "Book", "namespace": "default", "name": "mybook", "generation": 1, "is_adopted": false}
--- PASS: TestReconcilerUpdate (0.00s)
=== RUN   TestRegistry
--- PASS: TestRegistry (0.00s)
=== RUN   TestServiceController
--- PASS: TestServiceController (0.00s)
=== RUN   TestIsAdopted
--- PASS: TestIsAdopted (0.00s)
=== RUN   TestIsSynced
--- PASS: TestIsSynced (0.00s)
PASS
ok  	github.com/aws-controllers-k8s/runtime/pkg/runtime	0.016s

Also, tested if the errors are reported in SageMaker manually for AccessDeniedException. 

### DIFF
--- a/templates/pkg/resource/sdk.go.tpl
+++ b/templates/pkg/resource/sdk.go.tpl
@@ -164,16 +164,15 @@ func (rm *resourceManager) updateConditions (
 
 	// Terminal condition
 	var terminalCondition *ackv1alpha1.Condition = nil
-	var nonTerminalCondition *ackv1alpha1.Condition = nil
+	var recoverableCondition *ackv1alpha1.Condition = nil
 	for _, condition := range ko.Status.Conditions {
-		// If terminal condition exists, other errors will
-		// will not be reported
 		if condition.Type == ackv1alpha1.ConditionTypeTerminal {
 			terminalCondition = condition
 			break
 		}
-		if condition.Type == ackv1alpha1.ConditionTypeNonTerminal {
-			nonTerminalCondition = condition
+		// Continue to check if Terminal Condition exists
+		if condition.Type == ackv1alpha1.ConditionTypeRecoverable {
+			recoverableCondition = condition
 		}
 	}
 
@@ -194,25 +193,25 @@ func (rm *resourceManager) updateConditions (
 			terminalCondition.Status = corev1.ConditionFalse
 			terminalCondition.Message = nil
 		}
-		// Handling NonTerminal Conditions
+		// Handling Recoverable Conditions
 		if err != nil {
-			if nonTerminalCondition == nil {
+			if recoverableCondition == nil {
 				// Add a new Condition containing a non-terminal error
-				nonTerminalCondition = &ackv1alpha1.Condition{
-					Type:   ackv1alpha1.ConditionTypeNonTerminal,
+				recoverableCondition = &ackv1alpha1.Condition{
+					Type:   ackv1alpha1.ConditionTypeRecoverable,
 				}
-				ko.Status.Conditions = append(ko.Status.Conditions, nonTerminalCondition)
+				ko.Status.Conditions = append(ko.Status.Conditions, recoverableCondition)
 			}
-			nonTerminalCondition.Status = corev1.ConditionTrue
+			recoverableCondition.Status = corev1.ConditionTrue
 			awsErr, _ := ackerr.AWSError(err)
 			errorMessage := "Unknown Error"
 			if awsErr != nil {
 				errorMessage = awsErr.Message()
 			}
-			nonTerminalCondition.Message = &errorMessage
-		} else if nonTerminalCondition != nil {
-			nonTerminalCondition.Status = corev1.ConditionFalse
-			nonTerminalCondition.Message = nil
+			recoverableCondition.Message = &errorMessage
+		} else if recoverableCondition != nil {
+			recoverableCondition.Status = corev1.ConditionFalse
+			recoverableCondition.Message = nil
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


templates/pkg/resource/sdk.go.tpl
```diff
@@ -164,11 +164,17 @@ func (rm *resourceManager) updateConditions (
 
 	// Terminal condition
 	var terminalCondition *ackv1alpha1.Condition = nil
+	var nonTerminalCondition *ackv1alpha1.Condition = nil
 	for _, condition := range ko.Status.Conditions {
+		// If terminal condition exists, other errors will
+		// will not be reported
```
jaypipes, line 170: This isn't actually true, though, right? If a NonTerminalCondition is found first in the list of Condition structs, we will still end up "reporting" that NonTerminalCondition...

jaypipes (commented):
Apologies, @jkuruba, I changed my mind about naming this ConditionTypeNonTerminal... see the PR review on the runtime repo. Other than the name change, though, I'm good with this.

templates/pkg/resource/sdk.go.tpl
jkuruba: Yes, it was an oversight and I thought we were checking for the existing TerminalCondition below, but we are actually checking if the new error is Terminal. I fixed the comment to indicate that we continue to look for TerminalCondition even if we find a RecoverableError. TerminalCondition takes higher priority to display status as the sync stops.

surajkota: Good catch
right now if someone sets terminalCondition outside of customUpdateConditions, it is reset to False in this method based on `rm.terminalAWSError` check. Should we keep the same behaviour for Recoverable as well?

templates/pkg/resource/sdk.go.tpl
```diff
@@ -182,19 +188,43 @@ func (rm *resourceManager) updateConditions (
...
+		// Handling NonTerminal Conditions
+		if err != nil {
+			if nonTerminalCondition == nil {
+				// Add a new Condition containing a non-terminal error
+				nonTerminalCondition = &ackv1alpha1.Condition{
+					Type:   ackv1alpha1.ConditionTypeNonTerminal,
+				}
+				ko.Status.Conditions = append(ko.Status.Conditions, nonTerminalCondition)
+			}
+			nonTerminalCondition.Status = corev1.ConditionTrue
```
surajkota: `customUpdateConditions` should run before deciding if `Recoverable` can be set because one might set terminalCondition based on the error, it can be a service error or some other custom error

templates/pkg/resource/sdk.go.tpl
```diff
@@ -164,11 +164,16 @@ func (rm *resourceManager) updateConditions (
 
 	// Terminal condition
 	var terminalCondition *ackv1alpha1.Condition = nil
+	var recoverableCondition *ackv1alpha1.Condition = nil
 	for _, condition := range ko.Status.Conditions {
 		if condition.Type == ackv1alpha1.ConditionTypeTerminal {
 			terminalCondition = condition
 			break
```
surajkota, line 171: break should be removed

templates/pkg/resource/sdk.go.tpl
```diff
@@ -182,19 +187,43 @@ func (rm *resourceManager) updateConditions (
...
+		// Handling Recoverable Conditions
+		if err != nil {
+			if recoverableCondition == nil {
+				// Add a new Condition containing a non-terminal error
+				recoverableCondition = &ackv1alpha1.Condition{
+					Type:   ackv1alpha1.ConditionTypeRecoverable,
+				}
+				ko.Status.Conditions = append(ko.Status.Conditions, recoverableCondition)
+			}
+			recoverableCondition.Status = corev1.ConditionTrue
```
surajkota, line 205: need to let `customUpdateConditions` run before this can be set because terminal condition can be set there as well

surajkota (changes requested):

templates/pkg/resource/sdk.go.tpl
jaypipes: No worries, was just making sure I was reading the logic right :)

templates/pkg/resource/sdk.go.tpl
```diff
@@ -182,19 +188,43 @@ func (rm *resourceManager) updateConditions (
...
+			if nonTerminalCondition == nil {
+				// Add a new Condition containing a non-terminal error
+				nonTerminalCondition = &ackv1alpha1.Condition{
+					Type:   ackv1alpha1.ConditionTypeNonTerminal,
+				}
+				ko.Status.Conditions = append(ko.Status.Conditions, nonTerminalCondition)
+			}
+			nonTerminalCondition.Status = corev1.ConditionTrue
+			awsErr, _ := ackerr.AWSError(err)
+			errorMessage := "Unknown Error"
```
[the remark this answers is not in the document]
jkuruba: Done

templates/pkg/resource/sdk.go.tpl
jkuruba: Thanks for catching this. break causes the RecoverableError to be added twice. Removing it.

templates/pkg/resource/sdk.go.tpl
jkuruba: As we agreed, custom code overriding framework behavior in CustomUpdateCondition will also need to handle clearing of errors modified.

surajkota (approved):

jaypipes (approved):
:+1: thank you @jkuruba! :)

PR #41, opened 2021-04-06
always sort additional printer columns

jaypipes:
When returning AdditionalPrinterColumns for a CRD, we now always sort
the list of PrinterColumn structs by their Name. This prevents the
non-deterministic behaviour we've seen in Issue #720.

Also, the `pkg/generate:Generator.GetCRDs()` method was incorrectly only
adding PrinterColumns to a CRD when the field was specified as being
sourced from a different shape than the Create input or output shape (in
other words, only when the `pkg/generate/config.FieldConfig.From` field
was non-nil). This patch corrects this so that any field in a resource
may be marked as a printer column.

Fixes Issue aws-controller-k8s/community#720

### DIFF
--- a/pkg/model/printer_column.go
+++ b/pkg/model/printer_column.go
@@ -1,0 +1,142 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package model
+
+import (
+	"fmt"
+	"sort"
+)
+
+// PrinterColumn represents a single field in the CRD's Spec or Status objects
+type PrinterColumn struct {
+	CRD      *CRD
+	Name     string
+	Type     string
+	JSONPath string
+}
+
+// By can sort two PrinterColumns
+type By func(a, b *PrinterColumn) bool
+
+// Sort does an in-place sort of the supplied printer columns
+func (by By) Sort(subject []*PrinterColumn) {
+	pcs := printerColumnSorter{
+		cols: subject,
+		by:   by,
+	}
+	sort.Sort(pcs)
+}
+
+// printerColumnSorter sorts printer columns by name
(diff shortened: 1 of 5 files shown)


pkg/model/printer_column.go
```diff
@@ -0,0 +1,144 @@
...
+	By(byName).Sort(r.additionalPrinterColumns)
+	return r.additionalPrinterColumns
+}
+
+// addPrintableColumn adds an entry to the list of additional printer columns
+// using the given path and field types.
+func (r *CRD) addPrintableColumn(
+	field *Field,
+	jsonPath string,
+) *PrinterColumn {
```
A-Hilaly, line 77: nit: Maybe we should stop returning `*PrinterColumn` ? `AddStatusField` and `AddSpecField` don't use it

A-Hilaly (commented):
/lgtm overall :+1: I left two nit comments

pkg/model/printer_column.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// Less implements sort.Interface.Less
+func (pcs printerColumnSorter) Less(i, j int) bool {
+	return pcs.by(pcs.cols[i], pcs.cols[j])
+}
+
+// AdditionalPrinterColumns returns a sorted list of PrinterColumn structs for
+// the resource
+func (r CRD) AdditionalPrinterColumns() []*PrinterColumn {
```
RedbackThomson, line 64: Is this still accessible from the templates?

pkg/model/printer_column.go
jaypipes: nice callout. done.

pkg/model/printer_column.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	printColumnType, exists := acceptableColumnMaps[fieldColumnType]
+
+	if !exists {
+		msg := fmt.Sprintf(
+			"GENERATION FAILURE! Unable to generate a printer column for the field %s that has type %s.",
+			field.Names.Camel, fieldColumnType,
+		)
+		panic(msg)
+		return nil
```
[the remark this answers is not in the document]
jaypipes: fixed.

vijtrip2 (approved):
LGTM.

pkg/model/printer_column.go
jaypipes: yep. Go text templates silly-ish in this way. An exported field and an exported method with no arguments is accessed in an identical fashion.

In other words, this:

```
{{ .CRD.AdditionalPrinterColumns }}
```

is used regardless of whether `AdditionalPrinterColumns` is a field of CRD or a method with no arguments.

A-Hilaly (approved):

PR #47, opened 2021-04-28
allow "nested fields" to be configured

jaypipes:
When designing the `pkg/generate/config.FieldConfig` object, we really
only thought about (and tested for) "top-level fields". In other words,
fields that were on the CR's `Spec` or `Status` objects.

However, as we've added more configurability to the code-generator
regarding fields -- including marking fields as printable or secrets --
we're now limited substantially by our initial implementation that only
allows referring to top-level fields.

For example, we cannot mark a field as being a secret field if that
field is a subfield of one of these top-level fields. Concretely, this
means that with the existing generator configuration, we cannot support
configuring AmazonMQ's `Broker` CRD's `Spec.Users.Password` field as a
SecretKeyReference because `Password` is not a top-level field. It is
instead a field *of* a top-level field -- a "nested field" -- and
therefore our simplistic configuration system isn't able to tell the
code generator to replace the `string` Go type with
`*ackv1alpha1.SecretKeyReference` like it [does][secref-inject] for
top-level fields.

[secref-inject]: https://github.com/aws-controllers-k8s/code-generator/blob/94186d92e778792ccba11b5db3478e037256b36b/pkg/model/field.go#L61-L64

The solution to this dilemma is contained in this PR and involves
reworking our generator model in a couple ways:

1) The `pkg/model.CRD` struct now has a `Fields` attribute to go along
   with the existing `SpecFields` and `StatusFields` attributes.
   `CRD.Fields` is a `map[string]*Field` like `SpecFields` and
   `StatusFields`, but unlike those attributes, the `Fields` map keys
   are field *paths* and not the original aws-sdk-go Shape member names.
   When inspecting the AWS service API model as we do in
   `pkg/generate.Generator:GetCRDs()`, we now construct a field path for
   each `pkg/model.Field` object that corresponds to "where" in the CR the
   field is. For example, `Spec.Name` or `Status.BrokerInstances`

2) The `pkg/generate.Generator` has been updated with a
   `processNestedFields()` method to populate the new field-path-keyed
   `CRD.Fields` map with nested fields, making it possible to refer in our
   generator config, by field path, to fields that are not directly in the
   Status or Spec object.

A following patch will update the `Generator.GetTypeDefs()` method to
look up field configuration by this new field-path-based map and replace
`string` with `*ackv1alpha1.SecretKeyReference` when the field is a
secret field (Issue aws-controllers-k8s/community#743).

Issue aws-controllers-k8s/community#748

### DIFF
--- a/pkg/generate/generator.go
+++ b/pkg/generate/generator.go
@@ -226,6 +226,10 @@ func (g *Generator) GetCRDs() ([]*ackmodel.CRD, error) {
 	sort.Slice(crds, func(i, j int) bool {
 		return crds[i].Names.Camel < crds[j].Names.Camel
 	})
+	// This is the place that we build out the CRD.Fields map with
+	// `pkg/model.Field` objects that represent the non-top-level Spec and
+	// Status fields.
+	g.processNestedFields(crds)
 	g.crds = crds
 	return crds, nil
 }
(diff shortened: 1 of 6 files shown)


vijtrip2 (approved):
/lgtm

I like the distinction you made using '..' for list and Map members. Haven't seen that before.

A-Hilaly (approved):
/lgtm

pkg/generate/generator.go
```diff
@@ -401,6 +405,145 @@ func (g *Generator) GetTypeDefs() ([]*ackmodel.TypeDef, map[string]string, error
...
+func (g *Generator) processNestedListField(
+	crd *ackmodel.CRD,
+	baseFieldPath string,
+	baseField *ackmodel.Field,
+) {
+	baseFieldShape := baseField.ShapeRef.Shape
+	elementFieldShape := baseFieldShape.MemberRef.Shape
+	if elementFieldShape.Type != "structure" {
+		return
+	}
```
RedbackThomson, line 493: Q: What does this conditional shield against?

RedbackThomson (approved):

pkg/generate/generator.go
jaypipes: There's no need to recurse into list fields when it's a list of strings, for example. Only when it's a list of structs is it necessary to dig deeper into the nesting.

PR #49, opened 2021-04-29
replace nested field SecretKeyReference GoType

jaypipes:
In this patch, we post-process the `pkg/model.TypeDef` objects that are
constructed by the code generator for nested fields of type "structure".
This post-processing is used to replace the GoType of an Attr
representing a Secret field in a nested parent field.

For example, the AmazonMQ `Broker` resource has a `Spec.Users` top-level
field. This top-level field is of type `[]*User`. The `User` Go struct
type is constructed from a `pkg/model.TypeDef` object that describes the
struct's attributes. One of those struct attributes is called
`Password`, and we want to replace the Go type for this `Attr` from
`string` to `*ackv1alpha1.SecretKeyReference`. In order to do this Go
type replacement, we need to search through the TypeDef objects, looking
for the parent field TypeDef, then look through that TypeDef's Attr
collection for the Attr with the same name as our Field that has a
FieldConfig.IsSecret setting of `true`.

Issue aws-controllers-k8s/community#743


vijtrip2 (approved):
/lgtm

A-Hilaly:
I was trying to regenerate ecr with a fake secret field - I have two comments regarding the sdk and crd templates:

1. Since the nested types used in `apis/{{CRD.Name}}.go` are defined in `apis/types.go` (for example `User` type). I think that we also need a hack to avoid import errors [similar the one we have in `sdk.go`](https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/pkg/resource/sdk.go.tpl#L20-L29)

2. Currently `SetSDK` only sets secret values for [top level fields](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/code/set_sdk.go#L194-L207). I guess we should add the same for CRD nested types ?

jaypipes:
> I was trying to regenerate ecr with a fake secret field - I have two comments regarding the sdk and crd templates:
> 
>     1. Since the nested types used in `apis/{{CRD.Name}}.go` are defined in `apis/types.go` (for example `User` type). I think that we also need a hack to avoid import errors [similar the one we have in `sdk.go`](https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/pkg/resource/sdk.go.tpl#L20-L29)

Yes, good callout. I will update in the morning.

>     2. Currently `SetSDK` only sets secret values for [top level fields](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/code/set_sdk.go#L194-L207). I guess we should add the same for CRD nested types ?

Yep. ++

jaypipes:
@A-Hilaly great callouts on your review. I've addressed both of them in the latest commits on this PR. Thanks for your review again!

pkg/generate/code/set_sdk.go
```diff
@@ -243,14 +226,37 @@ func SetSDK(
...
+		if r.IsSecretField(memberName) {
+			out += setSDKForSecret(
+				cfg, r,
+				memberName,
+				targetVarName,
+				sourceAdaptedVarName,
+				indentLevel,
+			)
+			continue
+		}
```
A-Hilaly, line 259: Duplicate code :)

A-Hilaly (commented):
Neat! I like the 5a5300dd9b1379cff39077405a8b0221f9655941 commit :+1: 
I left two non blocking comments

pkg/generate/code/set_sdk.go
jaypipes: oh, wow... copy pasta. Will fix ASAP. thank you!

A-Hilaly (approved):
LGTM - let's 🚢 it!

PR #51, opened 2021-05-03
Use camel case name in SpecIdentifierField

nmvk:
In `crd.go.tpl` template file `field.Names.Camel` is used this results in `replicationGroupID` to be used as spec param rather than `replicationGroupId`.

Updating this method to be consistent with `crd.go.tpl` and to unblock elasticache code generation. 

Without fix

```
	if identifier.NameOrID == nil {
		return ackerrors.MissingNameIdentifier
	}
	r.ko.Spec.ReplicationGroupId = identifier.NameOrID
	return nil
}
```

With fix

```
	if identifier.NameOrID == nil {
		return ackerrors.MissingNameIdentifier
	}
	r.ko.Spec.ReplicationGroupID = identifier.NameOrID
	return nil
}
```

### DIFF
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -365,7 +365,7 @@ func (r *CRD) SpecIdentifierField() *string {
 	}
 	for memberName := range r.SpecFields {
 		if util.InStrings(memberName, lookup) {
-			return &memberName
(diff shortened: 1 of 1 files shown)


jaypipes (approved):
:+1:

PR #52, opened 2021-05-03
Add check for CRD list types in model.HasConflictingTypeName

ulucinar:
Issue #, if available:

Description of changes:
When generating controllers for the CloudFront API using the crossplane pipeline, I realized that we do not account for CRD List types (e.g., `DistributionList`) in `model.HasConflictingTypeName`, which results in such types being declared multiple times in generated code. This PR attempts to prevent this.


jaypipes (approved):
:+1: thank you @ulucinar!

ulucinar:
Thank you @jaypipes.

PR #54, opened 2021-05-06
delta: Add map types support for field comparison

A-Hilaly:
Issue https://github.com/aws-controllers-k8s/community/issues/757

- Add code to generate maps comparison helper functions
- Modify delta code generation to use the generated helper functions
- Add unit tests for `CompareResourceHelpers` and `CompareResource`

[Example](https://github.com/A-Hilaly/ack-apigatewayv2-controller/blob/delta-maps/pkg/resource/stage/delta.go) of the new `apigatewayv2` delta file for the `Stage` CRD

Description of changes:

Add a new code generation function (`CompareResourceHelpers`) that outputs Go function code that helps in the comparison of string to struct/list/map maps. The helper functions return true if two maps are equal and false otherwise.
This patch also updates `code.CompareResource` to call the generated helper functions without having to compare the maps manually.

For string to struct maps the generated code looks like:

```go
func equalRouteParameters(a, b map[string]*svcapitypes.ParameterConstraints) bool {
	if len(a) != len(b) {
		return false
	}
	for ka := range a {
		_, ok := b[ka]
		if !ok {
			return false
		}
	}
	for kb := range b {
		_, ok := a[kb]
		if !ok {
			return false
		}
	}
	for ka, aX := range a {
		bY := b[ka]
		if ackcompare.HasNilDifference(aX.Required, bY.Required) {
			return false
		} else if aX.Required != nil && bY.Required != nil {
			if *aX.Required != *bY.Required {
				return false
			}
		}
	}
	return true
}

```

/do-not-merge wip


vijtrip2:
I think this following part is redundant because you already check that length of a and b are equal And every element of a exists in b.

```
for kb := range b {
		_, ok := a[kb]
		if !ok {
			return false
		}
	}
```

A-Hilaly:
> I think this following part is redundant because you already check that length of a and b are equal And every element of a exists in b.
> 
> ```
> for kb := range b {
> 		_, ok := a[kb]
> 		if !ok {
> 			return false
> 		}
> 	}
> ```

Good point and nice catch! thanks @vijtrip2

pkg/generate/code/compare_helpers.go
```diff
@@ -0,0 +1,460 @@
...
+		}
+		specFieldNames = append(specFieldNames, fieldName)
+	}
+	sort.Strings(specFieldNames)
+
+	for _, fieldName := range specFieldNames {
+		specField := r.SpecFields[fieldName]
+		memberShapeRef := specField.ShapeRef
+		memberShape := memberShapeRef.Shape
+		shapes := collectSubShapes(memberShape)
```
A-Hilaly, line 39: Thinking that it might be better to collect the shapes that needs a helper method in `CompareResource` (during the `newResourceDelta` generation)

pkg/generate/code/compare_helpers.go
```diff
@@ -0,0 +1,460 @@
...
+	// Number of levels of indentation to use
+	indentLevel int,
+) string {
+	out := "\n"
+
+	specFieldNames := []string{}
+	for fieldName := range r.SpecFields {
+		specField := r.SpecFields[fieldName]
+		if expectHelperFunction(specField.ShapeRef.Shape) {
+		}
```
RedbackThomson, line 30: Empty conditional?

pkg/generate/code/compare_helpers.go
```diff
@@ -0,0 +1,460 @@
...
+		case "map":
+			valueShape := valueShape.ValueRef.Shape
+			shapes = append(shapes, collectSubShapes(valueShape)...)
+		case "list":
+			memberShape := valueShape.MemberRef.Shape
+			shapes = append(shapes, collectSubShapes(memberShape)...)
+		case "structure":
+			for _, memberName := range valueShape.MemberNames() {
+				memberShape := valueShape.MemberRefs[memberName].Shape
+				shapes = append(shapes, collectSubShapes(memberShape)...)
```
vijtrip2, line 126: Should duplicates be taken care of ?

pkg/generate/code/compare_helpers.go
RedbackThomson: I really do like that they're completely decoupled hmm... Would it be too crude to pass around a `[string]shape` map? I fear that using a partial `CRD` could cause issues if later on we accidentally try to use the other attributes.

pkg/generate/code/compare_helpers.go
```diff
@@ -0,0 +1,460 @@
...
+			}
+		}
+	case shape.Type == "list":
+		memberShape := shape.MemberRef.Shape
+		switch memberShape.Type {
+		case "map":
```
vijtrip2, line 132: nit: a little duplication of code, but not concerning at all.

pkg/generate/code/compare_helpers.go
```diff
@@ -0,0 +1,460 @@
...
+//
+//  if len(a) != len(b) {
+//  	return false
+//  }
+//  for ka := range a {
+//  	_, ok := b[ka]
+//  	if !ok {
+//  		return false
+//  	}
+//  }
```
RedbackThomson, line 169: nit: Remove references to iterators in code comments

pkg/generate/code/compare_helpers.go
```diff
@@ -0,0 +1,460 @@
...
+			"%s\n", indent,
+		)
+	case "structure":
+		out += equalStruct(
+			memberValueShape,
+			firstVarAdaptedName,
+			secondVarAdaptedName,
+			indentLevel+1,
+		)
+	case "list", "blop":
```
vijtrip2, line 285: what a 'blop' shape?

vijtrip2 (commented):
I will review remaining part very soon.

pkg/generate/code/compare_helpers.go
A-Hilaly: I was thinking about appending helper functions Go code to an array while generating the `newResourceDelta`. But i totally agree it will be very crude - especially if we conciser nested helper functions... Let's keep it decoupled :)

pkg/generate/code/compare_helpers.go
A-Hilaly: Yup this comment section is a little bit too large. Fixed

pkg/generate/code/compare_helpers.go
A-Hilaly: Yes the "map" and "list" cases are very similar, but for maps we need to look for `shape.ValueRef` and for lists we need to look for `shape.MemberRef`... maybe we could do something like this:

```go
	// set target shape
	var targetShape *awssdkmodel.Shape
	if shape.Type == "map" {
		targetShape = shape.ValueRef.Shape
	} else if shape.Type == "list" {
		targetShape = shape.MemberRef.Shape
	}

	switch shape.Type {
	case "map", "list":
		switch targetShape.Type {
		case "map":
			valueShape := targetShape.ValueRef.Shape
			shapes = append(shapes, collectSubShapes(valueShape)...)
		case "list":
			memberShape := targetShape.MemberRef.Shape
			shapes = append(shapes, collectSubShapes(memberShape)...)
		case "structure":
			for _, memberName := range targetShape.MemberNames() {
				memberShape := targetShape.MemberRefs[memberName].Shape
				shapes = append(shapes, collectSubShapes(memberShape)...)
			}
		}
	}
```

WDYT @vijtrip2 ?

pkg/generate/code/compare_helpers.go
A-Hilaly: Excellent point, i totally missed this. I'm pretty sure we can have duplicates in some cases..

pkg/generate/code/compare_helpers.go
A-Hilaly: Yikes! I meant "blob"... "blop" is a dungeon in a game i used to play long time ago ^^'

PR #58, opened 2021-05-10
Use ExportedName property of operation calls instead of Name

muvaf:
Description of changes: `Name` can return the name of the call with a date as we experienced in `CloudFront` service (like `GetDistribution2020_05_31WithContext`). Tested this statement with `rds` service and no change was seen.

I made the change in both ACK and Crossplane pipelines. It shouldn't affect the existing controllers.

### DIFF
--- a/templates/crossplane/apis/types.go.tpl
+++ b/templates/crossplane/apis/types.go.tpl
@@ -5,18 +5,15 @@
 package {{ .APIVersion }}
 
 import (
-	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
-	"github.com/aws/aws-sdk-go/aws"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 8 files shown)


templates/crossplane/apis/types.go.tpl
```diff
@@ -5,18 +5,15 @@
...
 import (
-	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
-	"github.com/aws/aws-sdk-go/aws"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // Hack to avoid import errors during build...
 var (
 	_ = &metav1.Time{}
-	_ = &aws.JSONValue{}
```
jaypipes, line 16: You won't need the ackv1alpha1 import, yes. But just double-check you won't need this one above...

jaypipes (approved):
Nice catch @muvaf! Crazy this isn't really documented (or reasonablly understandle default naming...) :)

PR #62, opened 2021-05-12
Add additional keys as identifiers

RedbackThomson:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/766

Description of changes:
When setting identifiers on a given resource, dynamically add all additional keys according to their json tag names.

### DIFF
--- a/templates/pkg/resource/resource.go.tpl
+++ b/templates/pkg/resource/resource.go.tpl
@@ -3,9 +3,14 @@
 package {{ .CRD.Names.Snake }}
 
 import (
+	"reflect"
+	"strings"
+
 	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
 	acktypes "github.com/aws-controllers-k8s/runtime/pkg/types"
+{{- if $idField := .CRD.SpecIdentifierField }}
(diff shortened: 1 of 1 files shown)


templates/pkg/resource/resource.go.tpl
```diff
@@ -67,15 +72,38 @@ func (r *resource) SetObjectMeta(meta metav1.ObjectMeta) {
...
+		return nil
+	}
+
+	specRef := reflect.Indirect(reflect.ValueOf(&r.ko.Spec))
+	specType := specRef.Type()
+
+	// Iterate over spec fields and associate corresponding json tags
+	for i := 0; i < specRef.NumField(); i++ {
+		// Get only the first field in the json tag (the field name)
+		jsonTag := strings.Split(specType.Field(i).Tag.Get("json"), ",")[0]
```
jaypipes, line 100: I'm OK with using the `reflect` package (even though it's incredibly ugly to work with, IMHO) here.

In the future, we may want to take an approach similar to how we handle TerminalConditions and simply generate a `switch` case lookup here...

jaypipes (changes requested):

A-Hilaly:
@RedbackThomson is gonna rework this PR, sometime next week

/hold

PR #65, opened 2021-05-17
Add code generation for type comparison

A-Hilaly:
Issue https://github.com/aws-controllers-k8s/community/issues/757

Description of changes:

Adds new code generation functions to `pkg/generate/code` that output Go code that checks the equality of two types. The generated code returns false at the first observed difference. The generated code re-uses the comparison functions for all nested fields of struct type.

This package will be imported and used by the generated `delta.go` file
that helps in comparing CRD fields.

For field of scalar type code looks like:
```go
	if ackcompare.HasNilDifference(a.RegistryID, b.RegistryID) {
		return false
	} else if a.RegistryID != nil && b.RegistryID != nil {
		if *a.RegistryID != *b.RegistryID {
			return false
		}
	}
```

For fields of type struct code look like:

```go
	if ackcompare.HasNilDifference(a.ImageScanningConfiguration, b.ImageScanningConfiguration) {
		return false
	} else if a.ImageScanningConfiguration != nil && b.ImageScanningConfiguration != nil {
		if !IsEqualImageScanningConfiguration(a.ImageScanningConfiguration, b.ImageScanningConfiguration) {
			return false
		}
	}
```

This patch doesn't implement comparison logic for slices and maps.

### DIFF
--- a/pkg/generate/code/equal.go
+++ b/pkg/generate/code/equal.go
@@ -1,0 +1,263 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package code
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+
+	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
+
+	ackmodel "github.com/aws-controllers-k8s/code-generator/pkg/model"
+	"github.com/aws-controllers-k8s/code-generator/pkg/names"
+)
+
+// IsEqualTypeDef returns Go code that checks the equality of two struct types.
+// The generated code returns false at the first spotted difference.
+//
+// Output code will look something like this:
+//
+//  if ackcompare.HasNilDifference(a.CreatedAt, b.CreatedAt) {
+//  	return false
+//  } else if a.CreatedAt != nil && b.CreatedAt != nil {
+//  	if *a.CreatedAt != *b.CreatedAt {
+//  		return false
+//  	}
+//  }
+//  if ackcompare.HasNilDifference(a.EncryptionConfiguration, b.EncryptionConfiguration) {
+//  	return false
+//  } else if a.EncryptionConfiguration != nil && b.EncryptionConfiguration != nil {
+//  	if !IsEqualEncryptionConfiguration(a.EncryptionConfiguration, b.EncryptionConfiguration) {
+//  		return false
+//  	}
+//  }
+func IsEqualTypeDef(
+	typeDef *ackmodel.TypeDef,
+	// String representing the name of the variable that represents the first
+	// object under comparison. This will typically be something like "a"
+	firstVarName string,
+	// String representing the name of the variable that represents the second
+	// object under comparison. This will typically be something like "b".
+	secondVarName string,
+	// Number of levels of indentation to use
+	indentLevel int,
+) string {
+	out := ""
+
+	// We need a deterministic order to loop over attributes
+	attrsNames := []string{}
+	for attrName := range typeDef.Attrs {
+		attrsNames = append(attrsNames, attrName)
+	}
+	sort.Strings(attrsNames)
+
+	// For each attribute generate the equality check logic
+	for _, attrName := range attrsNames {
+		attr := typeDef.Attrs[attrName]
+		out += isEqualField(attrName, attr.Shape, firstVarName, secondVarName, indentLevel)
+	}
+	return out
+}
+
+// isEqualField outputs Go code that compares two similar fields from different objects.
+// The generated code first checks the nility of the fields and then procede on comparing
+// their value.
+//
+// Output code will look something like this:
+//
+//	if ackcompare.HasNilDifference(a.RepositoryName, b.RepositoryName) {
+//		return false
+//	} else if a.RepositoryName != nil && b.RepositoryName != nil {
+//		if *a.RepositoryName != *b.RepositoryName {
+//			return false
+//		}
+//	}
+func isEqualField(
+	// fieldName is the field to generate the comparison logic for
+	fieldName string,
+	shape *awssdkmodel.Shape,
+	// String representing the name of the variable that represents the object
+	// under comparison.
+	firstVarName string,
+	// String representing the name of the variable that represents the second
+	// object under comparison.
+	secondVarName string,
+	// Number of levels of indentation to use
+	indentLevel int,
+) string {
+	out := ""
+
+	memberNames := names.New(fieldName)
+	memberNameClean := memberNames.Camel
+	firstAdaptedVarName := firstVarName + "." + memberNameClean
+	secondAdaptedVarName := secondVarName + "." + memberNameClean
+
+	nilCode := isEqualNil(
+		shape,
+		firstAdaptedVarName,
+		secondAdaptedVarName,
+		indentLevel,
+	)
+	if nilCode != "" {
+		out += fmt.Sprintf(
+			"%s else if %s != nil && %s != nil {\n",
+			nilCode, firstAdaptedVarName, secondAdaptedVarName,
+		)
+		indentLevel++
+	} else {
+		out += "\n"
+	}
+
+	indent := strings.Repeat("\t", indentLevel)
+
+	switch shape.Type {
+	case "structure":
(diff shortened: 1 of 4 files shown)


pkg/generate/code/equal.go
```diff
@@ -0,0 +1,236 @@
...
+
+	// For each attribute generate the equality check logic
+	for _, attrName := range attrsNames {
+		attr := typeDef.Attrs[attrName]
+		out += checkEqualField(attrName, attr.Shape, firstVarName, secondVarName, indentLevel)
+	}
+	return out
+}
+
+// equalStruct outputs Go code that compares two similar fields from different objects.
```
vijtrip2, line 57: documentation should start with method name "checkEqualField", correct?

pkg/generate/code/equal_test.go
```diff
@@ -0,0 +1,173 @@
...
+		return false
+	} else if a.CreatedAt != nil && b.CreatedAt != nil {
+		if *a.CreatedAt != *b.CreatedAt {
+			return false
+		}
+	}
+	if ackcompare.HasNilDifference(a.ImageScanningConfiguration, b.ImageScanningConfiguration) {
+		return false
+	} else if a.ImageScanningConfiguration != nil && b.ImageScanningConfiguration != nil {
+		if !ImageScanningConfiguration(a.ImageScanningConfiguration, b.ImageScanningConfiguration) {
```
vijtrip2, line 98: "!ImageScanningConfiguration(a.ImageScanningConfiguration, b.ImageScanningConfiguration)"

this looks incorrect. Is there a function with same name as shape?

pkg/generate/code/equal.go
A-Hilaly: Yep ++

pkg/generate/code/equal_test.go
A-Hilaly: >Is there a function with same name as shape?

Yes, for each type we generate a function (with same name as the type) that will be used to check the equality of two instances. The reason why we don't add any prefix to the function names, is that the package it self is called `equal`, so when importing this package in `delta.go`, we will be using it with something like `equal.ImageScanningConfiguration(a, b)`

RedbackThomson (approved):

pkg/generate/code/equal.go
```diff
@@ -0,0 +1,236 @@
...
+	switch shape.Type {
+	case "structure":
+		// We just re-use the generated functions to compare field of `struct` type
+		out += fmt.Sprintf(
+			"%sif !%s(%s, %s) {\n",
+			indent,
+			shape.ShapeName,
+			firstAdaptedVarName,
+			secondAdaptedVarName,
+		)
```
jaypipes, line 107: go ahead and add another parameter to this function, called `comparePackageAlias` and add this variable into the output Go code here, so that if this `comparePackageAlias` parameter was `svccompare` and the Shape name was `ImageScanningConfiguration`, the output Go code would be:

```go
if !svccompare.IsEqualImageScanningConfiguration(a.ImageScanningConfiguration, b.ImageScanningConfiguration) {
```

pkg/generate/code/equal.go
```diff
@@ -0,0 +1,236 @@
...
+
+	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
+
+	ackmodel "github.com/aws-controllers-k8s/code-generator/pkg/model"
+	"github.com/aws-controllers-k8s/code-generator/pkg/names"
+)
+
+// CheckEqualType returns Go code that checks the equality of two struct types.
+// The generated code returns false at the first spotted difference.
+func CheckEqualType(
```
jaypipes, line 29: Maybe rename this function to `IsEqualTypeDef`. See my comment at the bottom of this PR...

templates/pkg/equal/struct.go.tpl
```diff
@@ -0,0 +1,19 @@
+{{ template "boilerplate" }}
+
+package equal
```
jaypipes, line 3: After further thought, I think it would be best to name the package `pkg/compare` and have the functions comparing struct equality be named `IsEqual{ShapeName}`.

We can then alias `pkg/compare` to `svccompare` and then in the `pkg/resource/{resource}/delta.go` file's `newDelta` function, we would use both `ackcompare` general utility comparison functions as well as `svcompare` service-type-specific comparison functions.

The reason I would like to name the package `pkg/compare` instead of `pkg/equal` is because I think eventually we will want to generate comparison functions that evaluate more than just equality. For instance, we might have comparison functions that return whether an enum status string field has one of a set of valid values...

templates/pkg/equal/struct.go.tpl
```diff
@@ -0,0 +1,19 @@
...
+
+	svcapitypes "github.com/aws-controllers-k8s/{{ .ServiceIDClean }}-controller/apis/v1alpha1"
+)
+
+{{- range $typeDef := .TypeDefs }}
+
+// {{ .Names.Camel }} compares two {{ .Names.Camel }} object pointers. 
+func {{ .Names.Camel }}(a, b *svcapitypes.{{ .Names.Camel -}}) bool {
+{{ GoCodeEqual $typeDef "a" "b" 1 }}
+    return true
```
jaypipes, line 16: Make sure you `:%s/    /\t/g` :)

jaypipes (changes requested):
@A-Hilaly great start here, thank you! I changed my mind after further thought about what to name the generated package and the generated struct equality comparison functions (please see comment at the bottom of the PR).

pkg/generate/code/equal.go
A-Hilaly: `IsEqualImageScanningConfiguration` is in this same package and doesn't need to be imported.
I think this is something we should do in `pkg/generate/code/compare.go` (code that generates `newResourceDelta`).

pkg/generate/code/equal.go
```diff
@@ -0,0 +1,236 @@
...
+	"strings"
+
+	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
+
+	ackmodel "github.com/aws-controllers-k8s/code-generator/pkg/model"
+	"github.com/aws-controllers-k8s/code-generator/pkg/names"
+)
+
+// CheckEqualType returns Go code that checks the equality of two struct types.
+// The generated code returns false at the first spotted difference.
```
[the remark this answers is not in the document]
A-Hilaly: Done

pkg/generate/code/equal.go
```diff
@@ -0,0 +1,236 @@
...
+	for _, attrName := range attrsNames {
+		attr := typeDef.Attrs[attrName]
+		out += checkEqualField(attrName, attr.Shape, firstVarName, secondVarName, indentLevel)
+	}
+	return out
+}
+
+// checkEqualField outputs Go code that compares two similar fields from different objects.
+// The generated code first checks the nility of the fields and then procede on comparing
+// their value.
```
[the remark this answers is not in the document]
A-Hilaly: Done

templates/pkg/equal/struct.go.tpl
A-Hilaly: ++ for `pkg/compare` and import as `svccompare`

pkg/generate/code/equal.go
jaypipes: > `IsEqualImageScanningConfiguration` is in this same package and doesn't need to be imported.

Ahhhh, yes, correct you are!

> I think this is something we should do in `pkg/generate/code/compare.go` (code that generates `newResourceDelta`).

:+1:

pkg/generate/code/equal.go
```diff
@@ -0,0 +1,263 @@
...
+//  	return false
+//  } else if a.CreatedAt != nil && b.CreatedAt != nil {
+//  	if *a.CreatedAt != *b.CreatedAt {
+//  		return false
+//  	}
+//  }
+//  if ackcompare.HasNilDifference(a.EncryptionConfiguration, b.EncryptionConfiguration) {
+//  	return false
+//  } else if a.EncryptionConfiguration != nil && b.EncryptionConfiguration != nil {
+//  	if !EncryptionConfiguration(a.EncryptionConfiguration, b.EncryptionConfiguration) {
```
jaypipes, line 42: The above would be `if !IsEqualEncryptionConfiguration(...) {`, right?

jaypipes (approved):
:+1: nice work on this @A-Hilaly. One little mistake in the code comment for output code, but no big deal. :)

pkg/generate/code/equal.go
A-Hilaly: Yepp ++

jaypipes (approved):
muy bueno.

surajkota:
this PR has a bug
you can reproduce it by generating code for sagemaker-controller

notice VPC vs Vpc
```
// IsEqualVPCConfig compares two VPCConfig object pointers.
func IsEqualVPCConfig(a, b *svcapitypes.VPCConfig) bool {
```

```
else if a.VPCConfig != nil && b.VPCConfig != nil {
		if !IsEqualVpcConfig(a.VPCConfig, b.VPCConfig) {
			return false
		}
```

PR #69, opened 2021-05-20
helm chart: add log and accountid

surajkota:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/787

Tested using sagemaker-controller

```
k get pods -n ack-system -ojson

"spec": {
                "containers": [
                    {
                        "args": [
                            "--aws-account-id",
                            "$(AWS_ACCOUNT_ID)",
                            "--aws-region",
                            "$(AWS_REGION)",
                            "--enable-development-logging",
                            "$(ACK_ENABLE_DEVELOPMENT_LOGGING)",
                            "--log-level",
                            "$(ACK_LOG_LEVEL)",
                            "--resource-tags",
                            "$(ACK_RESOURCE_TAGS)",
                            "--watch-namespace",
                            "$(ACK_WATCH_NAMESPACE)"
                        ],
                        "command": [
                            "./bin/controller"
                        ],
                        "env": [
                            {
                                "name": "K8S_NAMESPACE",
                                "valueFrom": {
                                    "fieldRef": {
                                        "apiVersion": "v1",
                                        "fieldPath": "metadata.namespace"
                                    }
                                }
                            },
                            {
                                "name": "AWS_ACCOUNT_ID",
                                "value": "xxx"
                            },
                            {
                                "name": "AWS_REGION",
                                "value": "us-west-2"
                            },
                            {
                                "name": "ACK_WATCH_NAMESPACE"
                            },
                            {
                                "name": "ACK_ENABLE_DEVELOPMENT_LOGGING",
                                "value": "true"
                            },
                            {
                                "name": "ACK_LOG_LEVEL",
                                "value": "INFO"
                            },
                            {
                                "name": "ACK_RESOURCE_TAGS",
                                "value": "services.k8s.aws/managed=true,services.k8s.aws/created=%UTCNOW%,services.k8s.aws/namespace=%KUBERNETES_NAMESPACE%"
                            },
                            {
                                "name": "AWS_ROLE_ARN",
                                "value": "arn:aws:iam::xxx:role/ack-sage-role-ack-bug-bash-may18"
                            },
                            {
                                "name": "AWS_WEB_IDENTITY_TOKEN_FILE",
                                "value": "/var/run/secrets/eks.amazonaws.com/serviceaccount/token"
                            }
                        ],
```

### DIFF
--- a/templates/helm/values.yaml.tpl
+++ b/templates/helm/values.yaml.tpl
@@ -27,6 +27,12 @@ resources:
 aws:
   # If specified, use the AWS region for AWS API calls
   region: ""
+  account_id: ""
+
+# log level for the controller
+log:
+  enable_development_logging: false
+  level: info
 
 # If specified, the service controller will watch for object creation only in the provided namespace
 watchNamespace: ""
--- a/go.mod
+++ b/go.mod
@@ -17,5 +17,4 @@ require (
 	github.com/stretchr/testify v1.7.0
 	golang.org/x/mod v0.4.1
 	k8s.io/apimachinery v0.20.1
-	k8s.io/utils v0.0.0-20201110183641-67b214c5f920
 )
--- a/go.sum
+++ b/go.sum
@@ -61,8 +61,8 @@ github.com/armon/go-metrics v0.0.0-20180917152333-f0300d1749da/go.mod h1:Q73ZrmV
 github.com/armon/go-radix v0.0.0-20180808171621-7fddfc383310/go.mod h1:ufUuZ+zHj4x4TnLV4JWEpy2hxWSpsRywHrMgIH9cCH8=
 github.com/asaskevich/govalidator v0.0.0-20180720115003-f9ffefc3facf/go.mod h1:lB+ZfQJz7igIIfQNfa7Ml4HSf2uFQQRzpGGRXenZAgY=
 github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a/go.mod h1:lB+ZfQJz7igIIfQNfa7Ml4HSf2uFQQRzpGGRXenZAgY=
-github.com/aws-controllers-k8s/runtime v0.1.0 h1:XTBQf7hn0792yuCRy58IZpbZoGMrGCpSK2mZXkof5W4=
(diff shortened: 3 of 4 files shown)


templates/helm/values.yaml.tpl
```diff
@@ -27,6 +27,12 @@ resources:
 aws:
   # If specified, use the AWS region for AWS API calls
   region: ""
+  account_id: ""
+
+# log level for the controller
+log:
+  enable_development_logging: true
+  log_level: info
```
surajkota, line 35: `DEBUG` did not work, hence I kept it as `info` and not `INFO`

RedbackThomson: The `./kind-build-test.sh` script has this set to `debug`, but I don't think that's appropriate for production use anyway. I think this is good.

RedbackThomson (approved):

templates/helm/values.yaml.tpl
surajkota: Actually, I am inclined to keep default as debug because info logs are very less and hard to relate with state of resource. I think even the delta is not printed in info. Let's see what others have to say.

A-Hilaly:
@RedbackThomson PTAL at my comments on https://github.com/aws-controllers-k8s/test-infra/pull/27

/hold

templates/helm/values.yaml.tpl
jaypipes, line 35: I just realized that:

```yaml
log:
  log_level:
```

is redundant. Can we use this instead please?

```yaml
log:
  level: info
```

jaypipes (changes requested):

templates/helm/values.yaml.tpl
jaypipes: > Actually, I am inclined to keep default as debug because info logs are very less and hard to relate with state of resource. I think even the delta is not printed in info. Let's see what others have to say.

Delta is shown in Info, actually: https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/runtime/reconciler.go#L251

I would prefer to use info level for default.

jaypipes: @surajkota is correct. We are logging the diff/delta using V(1) (debug) not info...

That said, I would prefer to merge this with info as the default log level and address specific log messages (and moving some to info vs debug, etc) in a separate PR.

surajkota: Makes sense, I can create a separate issue/pr for that

PR #70, opened 2021-05-20
Immutable fields advisory

nmvk:
Add an advisory condition if an immutable field is modified.

```
  Conditions:
    Status:     False
    Type:       ACK.Terminal
    Status:     True
    Type:       ACK.ResourceSynced
    Status:     False
    Type:       ACK.Recoverable
    Message:    Immutable Spec fields have been modified : AtRestEncryptionEnabled,TransitEncryptionEnabled
    Status:     True
    Type:       ACK.Advisory
  Description:  Reinvent replication group
```

Clears the status upon fixing the field in the manifest

```
  Conditions:
    Status:     False
    Type:       ACK.Terminal
    Status:     True
    Type:       ACK.ResourceSynced
    Status:     False
    Type:       ACK.Recoverable
  Description:  Reinvent replication group
```

This commit allows one to specify immutable fields in `generator.yaml`.

```
      TransitEncryptionEnabled:
        is_immutable: true
      AtRestEncryptionEnabled:
        is_immutable: true
```

### DIFF
--- a/pkg/generate/config/field.go
+++ b/pkg/generate/config/field.go
@@ -155,6 +155,9 @@ type FieldConfig struct {
 	// IsSecret instructs the code generator that this field should be a
 	// SecretKeyReference.
 	IsSecret bool `json:"is_secret"`
+	// IsImmutable instructs the code generator to add advisory conditions
+	// if user modifies the spec field after resource was created.
+	IsImmutable bool `json:"is_immutable"`
 	// From instructs the code generator that the value of the field should
 	// be retrieved from the specified operation and member path
 	From *SourceFieldConfig `json:"from,omitempty"`
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -303,6 +303,32 @@ func (r *CRD) IsSecretField(path string) bool {
 	return false
 }
 
+// GetImmutableFieldPaths returns list of immutable field paths present in CRD
+func (r *CRD) GetImmutableFieldPaths() []string {
+	fConfigs := r.cfg.ResourceFields(r.Names.Original)
+	var immutableFields []string
+
+	for field, fieldConfig := range fConfigs {
+		if fieldConfig.IsImmutable {
+			immutableFields = append(immutableFields, field)
+		}
+	}
+
+	return immutableFields
+}
+
+// HasImmutableFieldChanges helper function that return true if there are any immutable field changes
+func (r *CRD) HasImmutableFieldChanges() bool {
+	fConfigs := r.cfg.ResourceFields(r.Names.Original)
+
+	for _, fieldConfig := range fConfigs {
+		if fieldConfig.IsImmutable {
+			return true
+		}
+	}
+	return false
+}
+
 // SetOutputCustomMethodName returns custom set output operation as *string for
 // given operation on custom resource, if specified in generator config
 func (r *CRD) SetOutputCustomMethodName(
--- a/templates/pkg/resource/sdk.go.tpl
+++ b/templates/pkg/resource/sdk.go.tpl
@@ -273,3 +273,63 @@ func (rm *resourceManager) terminalAWSError(err error) bool {
 	return false
 {{- end }}
 }
+
+{{- if .CRD.HasImmutableFieldChanges }}
+// getImmutableFieldChanges returns list of immutable fields from the
+func (rm *resourceManager) getImmutableFieldChanges(
+	delta *ackcompare.Delta,
+) []string {
+	var fields []string;
+
+	{{- range $immutableField := .CRD.GetImmutableFieldPaths }}
+		if delta.DifferentAt("{{$immutableField}}") {
+			fields = append(fields,"{{$immutableField}}")
+		}
+	{{- end }}
+
+	return fields
+}
+
+// handleImmutableFieldsChangedCondition validates the immutable fields and set appropriate condition
+func (rm *resourceManager) handleImmutableFieldsChangedCondition(
+	r *resource,
+	delta *ackcompare.Delta,
+) *resource {
+
+	fields := rm.getImmutableFieldChanges(delta)
+	ko := r.ko.DeepCopy()
+	var advisoryCondition *ackv1alpha1.Condition = nil
+	for _, condition := range ko.Status.Conditions {
+		if condition.Type == ackv1alpha1.ConditionTypeAdvisory {
+			advisoryCondition = condition
+			break
+		}
+	}
+
+	// Remove the advisory condition if issue is no longer present
+	if len(fields) == 0 && advisoryCondition != nil{
+		var newConditions []*ackv1alpha1.Condition
+		for _, condition := range ko.Status.Conditions {
+			if condition.Type != ackv1alpha1.ConditionTypeAdvisory {
+				newConditions = append(newConditions,condition)
+			}
+		}
+		ko.Status.Conditions = newConditions
+	}
+
+	if len(fields) > 0 {
+		if advisoryCondition == nil {
+			advisoryCondition = &ackv1alpha1.Condition{
+				Type: ackv1alpha1.ConditionTypeAdvisory,
+			}
+			ko.Status.Conditions = append(ko.Status.Conditions, advisoryCondition)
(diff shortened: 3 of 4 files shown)


pkg/model/crd.go
```diff
@@ -303,6 +303,20 @@ func (r *CRD) IsSecretField(path string) bool {
 	return false
 }
 
+// GetImmutableFields returns list of immutable fields present in CRD
+func (r *CRD) GetImmutableFields() []string {
+	fConfigs := r.cfg.ResourceFields(r.Names.Original)
+	var immutableFields []string
+
+	for field,fieldConfig := range fConfigs {
```
jaypipes, line 311: nit: need a space between `field,` and `fieldConfig` above. (This is usually fixed with `go fmt`, so @nmvk please make sure you run `go fmt` before `git commit`'ing :) or use an autoformatter plugin in your text editor or IDE)

pkg/model/crd.go
```diff
@@ -303,6 +303,20 @@ func (r *CRD) IsSecretField(path string) bool {
 	return false
 }
 
+// GetImmutableFields returns list of immutable fields present in CRD
```
jaypipes, line 306: Technically, what this is returning is a list of **field paths** that are marked immutable, not fields...

vijtrip2 (approved):

templates/pkg/resource/sdk.go.tpl
```diff
@@ -273,3 +273,55 @@ func (rm *resourceManager) terminalAWSError(err error) bool {
 	return false
 {{- end }}
 }
+
+{{- if .CRD.GetImmutableFields }}
+// validateImmutableFields validates the immutable fields and set appropriate condition
+func (rm *resourceManager) validateImmutableFields(ko *svcapitypes.{{ .CRD.Names.Camel }},
+	delta *ackcompare.Delta,) {
```
jaypipes, line 280: Please pass a `*resource` pointer instead of `*svcapitypes.{{ .CRD.Names.Camel }}`, similar to the `updateConditions()`  method above, which has this function signature:

```go
func (rm *resourceManager) updateConditions (
        r *resource,
        err error,
) (*resource, bool) {
```

You can then get the `ko` object from `r` like so:

```go
        ko := r.ko
```

Also, please style the function signature like so:

```go
func (rm *resourceManager) validateImmutableFields(
        r *resource,
        delta *ackcompare.Delta,
) {
```

templates/pkg/resource/sdk.go.tpl
```diff
@@ -273,3 +273,55 @@ func (rm *resourceManager) terminalAWSError(err error) bool {
...
+		if diff.Path.Contains("Spec") {
+		{{- range $immutableField := .CRD.GetImmutableFields }}
+			if diff.Path.Contains("{{$immutableField}}") {
+				fields = append(fields,"{{$immutableField}}")
+			}
+		{{- end }}
+		}
+	}
+
+	// Warning condition
```
jaypipes, line 293: nix the above comment, it's not necessary.

templates/pkg/resource/sdk.go.tpl
```diff
@@ -273,3 +273,55 @@ func (rm *resourceManager) terminalAWSError(err error) bool {
...
+		for _, condition := range ko.Status.Conditions {
+			if condition.Type != ackv1alpha1.ConditionTypeWarning {
+				newConditions = append(newConditions,condition)
+			}
+		}
+		ko.Status.Conditions = newConditions
+	}
+
+	if len(fields) > 0 {
+		// Add warning condition
```
jaypipes, line 314: I'd nix the above comment as well, since it's obvious that is what we're doing here.

templates/pkg/resource/sdk.go.tpl
```diff
@@ -273,3 +273,55 @@ func (rm *resourceManager) terminalAWSError(err error) bool {
...
+
+	for _, diff := range delta.Differences {
+		if diff.Path.Contains("Spec") {
+		{{- range $immutableField := .CRD.GetImmutableFields }}
+			if diff.Path.Contains("{{$immutableField}}") {
+				fields = append(fields,"{{$immutableField}}")
+			}
+		{{- end }}
+		}
+	}
```
jaypipes, line 291: Couple things about the above....

a) Recommend putting it into a separate helper function called `hasImmutableFieldChanges`
b) The [Update code paths will never fire if the Spec hasn't been modified](https://github.com/aws-controllers-k8s/runtime/blob/0ff9ffa7634a156ab23b4102e200df50e9ece56b/pkg/runtime/reconciler.go#L248), so I think you should be able to simplify the above to just this:

```
{{- range $immutableField := .CRD.GetImmutableFields }}
if delta.DifferentAt("{{$immutableField}}") {
        fields = append(fields,"{{$immutableField}}")
}
{{- end }}
```

surajkota:
how about making the message available in `ACK.Recoverable` itself?

because it looks like the controller will never be able to reach the desired state/spec so customer should revert the fields

templates/pkg/resource/sdk_update.go.tpl
```diff
@@ -38,6 +38,9 @@ func (rm *resourceManager) sdkUpdate(
 {{- if $hookCode := Hook .CRD "sdk_update_pre_set_output" }}
 {{ $hookCode }}
 {{- end }}
+{{- if .CRD.GetImmutableFields }}
+    rm.validateImmutableFields(ko, delta)
```
jaypipes, line 42: Need to replace the four space characters above with a `\t` :)

jaypipes (changes requested):

jaypipes:
> how about making the message available in `ACK.Recoverable` itself?
> 
> because it looks like the controller will never be able to reach the desired state/spec so customer should revert the fields

`ACK.ConditionTypeRecoverable` was mainly for communicating with the user that an external dependency or failure happened and that the controller would be able to reconcile the resource if those external things were fixed. In this case, we want to indicate to the user that they've made a modification to the desired state that we're going to ignore and they should revert their changes to get rid of the message. Subtly different, IMHO.

templates/pkg/resource/sdk.go.tpl
```diff
@@ -273,3 +273,55 @@ func (rm *resourceManager) terminalAWSError(err error) bool {
 	return false
 {{- end }}
 }
+
+{{- if .CRD.GetImmutableFields }}
+// validateImmutableFields validates the immutable fields and set appropriate condition
+func (rm *resourceManager) validateImmutableFields(ko *svcapitypes.{{ .CRD.Names.Camel }},
```
jaypipes, line 279: Also, since this isn't actually validating anything (and is really just identifying changes to immutable fields and adding a condition to the resource), I'd rename this function to `handleImmutableFieldsChangedCondition` or something like that...

pkg/generate/config/field.go
```diff
@@ -155,6 +155,9 @@ type FieldConfig struct {
 	// IsSecret instructs the code generator that this field should be a
 	// SecretKeyReference.
 	IsSecret bool `json:"is_secret"`
+	// IsImmutable instructs the code generator to add warning conditions
```
RedbackThomson, line 158: nit
```suggestion
	// IsImmutable instructs the code generator to add advisory conditions
```

RedbackThomson (approved):

templates/pkg/resource/sdk.go.tpl
```diff
@@ -273,3 +273,62 @@ func (rm *resourceManager) terminalAWSError(err error) bool {
...
+
+	{{- range $immutableField := .CRD.GetImmutableFieldPaths }}
+		if delta.DifferentAt("{{$immutableField}}") {
+			fields = append(fields,"{{$immutableField}}")
+		}
+	{{- end }}
+
+	return fields
+}
+// handleImmutableFieldsChangedCondition validates the immutable fields and set appropriate condition
```
A-Hilaly, line 292: nit: space between functions needed

A-Hilaly (approved):
LGTM @nmvk :+1:

Can you please change the PR description? @ack-bot will squash/commit using the PR TItle/description

PR #73, opened 2021-05-27
Adding ClusterIP service in helm chart to expose controller metrics.

vijtrip2:
Description of changes:

Adding ClusterIP service in helm chart to expose controller metrics.

It is much simpler and easier to update prometheus scraping configuration with a static endpoint.

### DIFF
--- a/templates/helm/templates/service.yaml
+++ b/templates/helm/templates/service.yaml
@@ -1,0 +1,30 @@
+{{- if .Values.service.create }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "app.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: {{ include "app.name" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: Helm
+    app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+    k8s-app: {{ include "app.name" . }}
+    helm.sh/chart: {{ include "chart.name-version" . }}
+    control-plane: controller
+spec:
+  selector:
+    app.kubernetes.io/name: {{ include "app.name" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: Helm
+    k8s-app: {{ include "app.name" . }}
+{{- range $key, $value := .Values.deployment.labels }}
+    {{ $key }}: {{ $value | quote }}
+{{- end }}
+  type: {{ .Values.service.type }}
+  ports:
+  - name: metricsport
+    port: 8080
+    targetPort: 8080
+    protocol: TCP
+{{- end }}
--- a/templates/helm/values.yaml.tpl
+++ b/templates/helm/values.yaml.tpl
@@ -16,6 +16,10 @@ deployment:
   labels: {}
   containerPort: 8080
 
+service:
+  create: false
+  type: "ClusterIP"
+
 resources:
   requests:
     memory: "64Mi"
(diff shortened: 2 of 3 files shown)


A-Hilaly (approved):
Great! :+1:

RedbackThomson (changes requested):
I don't think this is necessary for any of the default installations we offer. There is no reason to access the controller pods through a service as they offer no network APIs. I would rather see a separate service spec added to cluster gitops alongside the helm chart if a customer wants to expose metrics in this way.

A-Hilaly:
> I don't think this is necessary for any of the default installations we offer. There is no reason to access the controller pods through a service as they offer no network APIs. I would rather see a separate service spec added to cluster gitops alongside the helm chart if a customer wants to expose metrics in this way.

There is a metrics api exposed at :8080 in every controller https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/config/config.go#L60-L64

A-Hilaly:
> If this is a must, then at least gating the service behind a create: true/false in the helm values would be required. (Default being false)

LGTM, If we're going with solution, we could also gate the metrics API using service controller flags, thinking `--enable-metrics-server`

RedbackThomson:
> > I don't think this is necessary for any of the default installations we offer. There is no reason to access the controller pods through a service as they offer no network APIs. I would rather see a separate service spec added to cluster gitops alongside the helm chart if a customer wants to expose metrics in this way.
> 
> There is a metrics api exposed at :8080 in every controller https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/config/config.go#L60-L64

I understand that, but does Prometheus not detect all metrics servers in every pod by default?

vijtrip2:
>"If this is a must, then at least gating the service behind a create: true/false in the helm values would be required. (Default being false)"

Good Idea. I will definitely do this.

>"I understand that, but does Prometheus not detect all metrics servers in every pod by default?"

While prototyping, I did not find this to be the case. I can double check.
a) First I tried to use prometheus Kubernetes Service Discovery for finding these metrics but was not able to successfully do it. 
b) Then I tried this static config method where we ask Prometheus to scrape the metric from a static endpoint. It was very simple, straight forward and clean.
Hence I went ahead with method (b).

RedbackThomson (approved):

A-Hilaly:
>I understand that, but does Prometheus not detect all metrics servers in every pod by default?

AFAIK Prometheus does not detect any metrics at all by default - it needs to be configured for pod discovery

templates/helm/values.yaml.tpl
```diff
@@ -16,6 +16,10 @@ deployment:
   labels: {}
   containerPort: 8080
 
+service:
+  create: false
+  type: "ClusterIP"
```
jaypipes, line 21: @vijtrip2 @RedbackThomson @A-Hilaly

Can we rename this to specify that it has to do with the metrics server and not the service controller itself?

Perhaps something like this?

```yaml
metrics:
  service:
    # Set to true to automatically create a Kubernetes Service resource for the
    # Prometheus metrics server endpoint in controller
    create: false
    # Which Type to use for the Kubernetes Service?
    # See: https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types
    type: ClusterIP
```

templates/helm/templates/service.yaml
```diff
@@ -0,0 +1,30 @@
+{{- if .Values.service.create }}
```
jaypipes, line 1: Can we rename this file to `templates/metrics-service.yaml` please?

templates/helm/values.yaml.tpl
vijtrip2: Yep, I will get it done.

PR #76, opened 2021-06-02
Patch the correct observed state after a `rm.ReadOne` call

A-Hilaly:
Issue https://github.com/aws-controllers-k8s/community/issues/807

For some controllers we noticed an incorrect reconciliation behaviour
when a custom hook code was returning a `RequeueNeededAfter` error.
More precisely `resourceManager.ReadOne` was not returning the correct
observed state when `sdkFind` returns with an error. Which was causing
 the reconciler to patch to wrong state (`desird` instead `observed`).

This patch fixes this issue by passing the correct observed state to `onError`.

### DIFF
--- a/templates/pkg/resource/manager.go.tpl
+++ b/templates/pkg/resource/manager.go.tpl
@@ -74,6 +74,9 @@ func (rm *resourceManager) ReadOne(
 	}
 	observed, err := rm.sdkFind(ctx, r)
 	if err != nil {
+		if observed != nil {
+			return rm.onError(observed, err)
+		}
 		return rm.onError(r, err)
 	}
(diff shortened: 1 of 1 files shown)


A-Hilaly:
i'm currently testing this fix with `ecr` and `dynamodb` controllers
/hold

jaypipes:
@A-Hilaly you may want to update the PR/commit message to mention that `RequeueNeededAfter` is actually an `error` and is the reason why we're noticing some of this behaviour in some of the custom hook code.

A-Hilaly:
@jaypipes Updated,  PTAL again.

jaypipes (approved):
:+1:

PR #78, opened 2021-06-02
Add a new operation config to override the `SetOutput` wrapper field

A-Hilaly:
Issue N/A

In some AWS APIs `Describe` operations output contains the
wanted information in a wrapper field like `TableDescription` and
`GlobalTableDescription` ... However in some other APIs for the same
operation type, the wanted information is wrapped in different nested
object, like `BackupDescription.BackupDetails` for example.

Description of changes:
This patch adds a new operation configuration that instructs the
code-generator to use a specific field path, in order to find the
struct that contains the information we want to set the output from.

### DIFF
--- a/pkg/generate/code/set_resource.go
+++ b/pkg/generate/code/set_resource.go
@@ -118,14 +118,29 @@ func SetResource(
 		return ""
 	}
 
+	var err error
 	// We might be in a "wrapper" shape. Unwrap it to find the real object
-	// representation for the CRD's createOp. If there is a single member
-	// shape and that member shape is a structure, unwrap it.
-	if outputShape.UsedAsOutput && len(outputShape.MemberRefs) == 1 {
-		for memberName, memberRef := range outputShape.MemberRefs {
-			if memberRef.Shape.Type == "structure" {
-				sourceVarName += "." + memberName
-				outputShape = memberRef.Shape
+	// representation for the CRD's createOp/DescribeOP.
+
+	// Use the wrapper field path if it's given in the ack-generate config file.
+	wrapperFieldPath := r.GetOutputWrapperFieldPath(op)
+	if wrapperFieldPath != nil {
+		outputShape, err = GetWrapperOutputShape(outputShape, *wrapperFieldPath)
+		if err != nil {
+			msg := fmt.Sprintf("Unable to unwrap the output shape: %v", err)
+			panic(msg)
+		}
+		sourceVarName += "." + *wrapperFieldPath
+	} else {
+		// If the wrapper field path is not specified in the config file and if
+		// there is a single member shape and that member shape is a structure,
+		// unwrap it.
+		if outputShape.UsedAsOutput && len(outputShape.MemberRefs) == 1 {
+			for memberName, memberRef := range outputShape.MemberRefs {
+				if memberRef.Shape.Type == "structure" {
+					sourceVarName += "." + memberName
+					outputShape = memberRef.Shape
+				}
 			}
 		}
 	}
@@ -300,6 +315,41 @@ func SetResource(
 		)
 	}
 	return out
+}
+
+// GetWrapperOutputShape returns the shape of the last element of a given field
+// Path. It carefully unwraps the output shape and verifies that every element
+// of the field path exists in their correspanding parent shape and that they are
+// structures.
+func GetWrapperOutputShape(
+	shape *awssdkmodel.Shape,
+	fieldPath string,
+) (*awssdkmodel.Shape, error) {
+	if fieldPath == "" {
+		return shape, nil
+	}
+	fieldPathParts := strings.Split(fieldPath, ".")
+	for x, wrapperField := range fieldPathParts {
+		for memberName, memberRef := range shape.MemberRefs {
+			if memberName == wrapperField {
+				if memberRef.Shape.Type != "structure" {
+					// All the mentionned shapes must be structure
+					return nil, fmt.Errorf(
+						"Expected SetOutput.WrapperFieldPath to only contain fields of type 'structure'."+
+							" Found %s of type '%s'",
+						memberName, memberRef.Shape.Type,
+					)
+				}
+				remainPath := strings.Join(fieldPathParts[x+1:], ".")
+				return GetWrapperOutputShape(memberRef.Shape, remainPath)
+			}
+		}
+		return nil, fmt.Errorf(
+			"Incorrect SetOutput.WrapperFieldPath. Could not find %s in Shape %s",
+			wrapperField, shape.ShapeName,
+		)
+	}
+	return shape, nil
 }
 
 func ListMemberNameInReadManyOutput(
--- a/pkg/generate/testdata/models/apis/dynamodb/0000-00-00/generator.yaml
+++ b/pkg/generate/testdata/models/apis/dynamodb/0000-00-00/generator.yaml
@@ -4,3 +4,12 @@ resources:
       errors:
         404:
           code: ResourceNotFoundException
+operations:
+  DescribeBackup:
+    # DescribeBackupOutput is an unsual shape because it contains information for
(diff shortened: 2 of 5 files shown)


pkg/generate/code/set_resource.go
```diff
@@ -119,13 +119,46 @@ func SetResource(
...
-	// shape and that member shape is a structure, unwrap it.
-	if outputShape.UsedAsOutput && len(outputShape.MemberRefs) == 1 {
-		for memberName, memberRef := range outputShape.MemberRefs {
-			if memberRef.Shape.Type == "structure" {
-				sourceVarName += "." + memberName
-				outputShape = memberRef.Shape
+	// representation for the CRD's createOp/DescribeOP.
+
+	// Use the wrapper field path if it's given in the ack-generate config file.
+	wrapperFieldPath := r.SetOutputWrapperFieldPath(op)
```
RedbackThomson, line 125: I understand this field comes from the `set_output_wrapper_field_path` JSON key, but it's pretty confusing to see a getter with the name `Set*`.

pkg/generate/code/set_resource.go
```diff
@@ -119,13 +119,46 @@ func SetResource(
...
+	if wrapperFieldPath != nil {
+		wrapperFieldParts := strings.Split(*wrapperFieldPath, ".")
+	mainLoop:
```
RedbackThomson, line 128: Interested to hear opinions about use of goto labels. I understand you want to continue the outer loop, but I imagine this could be achieved with booleans - although messier.

pkg/generate/code/set_resource.go
A-Hilaly: Yes, i agree it's very confusing.. I'm not sure if `GetSetOutputWrapperFieldPath` is a better option.

A-Hilaly: We also have a similar function is crd.go: `SetOutputCustomMethodName`

pkg/generate/code/set_resource.go
A-Hilaly: I don't have a strong opinion about `goto` statements - I personally only use them sometimes for `continue`/`break` statements. I'll push a different approach using using booleans

pkg/generate/code/set_resource.go
RedbackThomson: I understand using `set_*` for the JSON key, but the getter for this value could just be `GetOutputWrapperFieldPath`? 

```go
	OutputWrapperFieldPath string `json:"set_output_wrapper_field_path,omitempty"`
```

then:

```go
 	wrapperFieldPath := r.OutputWrapperFieldPath(op) 
```

Perhaps?

pkg/generate/code/set_resource.go
RedbackThomson: I don't believe that boolean logic here would be any cleaner. I think this is a fair use of `goto` statements.

pkg/generate/code/set_resource.go
A-Hilaly: ++ in this case I prefer to remove the `set_` from the JSON Key as well - just to keep fields names and json tags synced

pkg/generate/code/set_resource.go
A-Hilaly: Indeed it will be a little bit messier :)
Thoughts @vijtrip2 @jaypipes ?

jaypipes: How about breaking the below code out into a helper function?

pkg/generate/code/set_resource.go
A-Hilaly: Renamed to `OutputWrapperFieldPath`

pkg/generate/code/set_resource.go
A-Hilaly: Done, and added unit tests for the helper function

pkg/generate/testdata/models/apis/dynamodb/0000-00-00/generator.yaml
```diff
@@ -4,3 +4,12 @@ resources:
...
         404:
           code: ResourceNotFoundException
+operations:
+  DescribeBackup:
+    # DescribeBackupOutput is an unsual shape because it contains information for
+    # the backup it self (BackupDetails), the table details when the backup was
+    # created (SourceTableDetails) and the table features (SourceTableFeatureDetails).
+    # If not specified the code generator will try to determine the wrapper field by
+    # selecting for the output shape that only have a single member, which is incorrect
+    # in this case.
```
jaypipes, line 14: :+1:

pkg/generate/code/set_resource.go
```diff
@@ -302,6 +317,39 @@ func SetResource(
...
+	shape *awssdkmodel.Shape,
+	fieldPath string,
+) (*awssdkmodel.Shape, error) {
+	fieldPathParts := strings.Split(fieldPath, ".")
+mainLoop:
+	for _, wrapperField := range fieldPathParts {
+		for memberName, memberRef := range shape.MemberRefs {
+			if memberName == wrapperField {
+				if memberRef.Shape.Type != "structure" {
+					// All the mentionned shapes must a be of type 'structure'
```
jaypipes, line 334: s/must a be of type 'structure'/must be structures/

pkg/generate/code/set_resource.go
```diff
@@ -302,6 +317,39 @@ func SetResource(
...
+				continue mainLoop
+			}
+		}
+		return nil, fmt.Errorf(
+			"Incorrect SetInput.WrapperFieldPath. Could not find %s in Shape %s",
+			wrapperField, shape.ShapeName,
+		)
+	}
+	return shape, nil
+}
```
jaypipes, line 351: I'm with @RedbackThomson on the goto label use here being a bit of a code smell...

How about using recursion instead?

```go
func GetWrapperOutputShape(
    shape *awssdkmodel.Shape,
    fieldPath string,
) (*awssdkmodel.Shape, error) {
    fieldPathParts := strings.Split(fieldPath, ".")
    for x, field := range fieldPathParts {
        for memberName, memberRef := range shape.MemberRefs {
            if memberName == wrapperField {
                if memberRef.Shape.Type != "structure" {
                    // handle error...
                }
                remainPath := strings.Join(fieldPathParts[x+1:], ".")
                return GetWrapperOutputShape(memberRef.Shape, remainPath)
        }
    }
    return shape, nil
}
```

jaypipes (changes requested):
@A-Hilaly left a little suggestion to replace the goto with a simple recursion.

pkg/generate/code/set_resource.go
```diff
@@ -302,6 +317,39 @@ func SetResource(
 	return out
 }
 
+// GetWrapperOutputShape returns the shape of the last element of a given field
+// Path. It carefully unwrapps the output shape and verifies that every element
```
jaypipes, line 321: s/unwrapps/unwraps/

pkg/generate/code/set_resource.go
```diff
@@ -302,6 +317,39 @@ func SetResource(
...
+) (*awssdkmodel.Shape, error) {
+	fieldPathParts := strings.Split(fieldPath, ".")
+mainLoop:
+	for _, wrapperField := range fieldPathParts {
+		for memberName, memberRef := range shape.MemberRefs {
+			if memberName == wrapperField {
+				if memberRef.Shape.Type != "structure" {
+					// All the mentionned shapes must a be of type 'structure'
+					return nil, fmt.Errorf(
+						"Expected SetInput.WrapperFieldPath to only contain fields of type 'structure'."+
```
jaypipes, line 336: s/SetInput/SetOutput/?

A-Hilaly: Yikes.. x)

jaypipes (approved):
merci beaucoup, monsieur :)

PR #79, opened 2021-06-02
Add a new config to override kubebuilder `printcolumn:name`

A-Hilaly:
Issue N/A

In some CRDs like `GlobalTable` adding a `is_printable: true` to the `GlobalTableStatus` causes `kubectl get` to print very large/redundant column names.

Description of changes:
- This patch adds a new config to allow overriding kubebuilder `printcolumn:name`
- Adds `metav1.Time` to the `acceptableColumnMap`

### DIFF
--- a/pkg/generate/config/field.go
+++ b/pkg/generate/config/field.go
@@ -135,6 +135,10 @@ type FieldConfig struct {
 	// AdditionalPrinterColumns list to be included in the `kubectl get`
 	// response.
 	IsPrintable bool `json:"is_printable"`
+	// PrintName instructs the code generator to override the column name used
+	// to include the field in `kubectl get` response. If `IsPrintable` is false
+	// this field is ignored.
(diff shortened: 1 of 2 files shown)


RedbackThomson:
Hmm I wish I had the foresight to add this initially, because I'd probably have made the config a struct containing the boolean and the string - rather than them both being on the parent resource level. Regardless, without breaking backwards compatibility, this is good.

/approve

A-Hilaly:
>Hmm I wish I had the foresight to add this initially, because I'd probably have made the config a struct containing the boolean and the string - rather than them both being on the parent resource level. Regardless, without breaking backwards compatibility, this is good.

Agreed. I hesitated to introduce a struct config for it.. but as you said it will break the compatibility .. maybe one day a generator config v2 :)

jaypipes (approved):
Nice, Amine. :)

PR #81, opened 2021-06-03
crossplane: Mark spec as required, status.atProvider as optional

negz:
This is required to fix https://github.com/crossplane/provider-aws/issues/697

In the vast majority of resources a spec is required, so I'd like to mark it as such. I believe that in the edge cases where the spec really contains only optional fields it's possible to write `spec: {}`.

We prefer not to mark status fields as required, since resources will always exist at some point without their status and we consider 'requiredness' to be more of a human-facing constraint than a software-facing constraint that our managed resource controllers should be subject to.


negz:
CC @muvaf

muvaf (approved):
@negz thanks for catching this!

A-Hilaly:
Thanks @negz !
/approve

PR #82, opened 2021-06-08
Move "printer column" options under the same generator config struct

a-hilaly:
Issue N/A

Initially we used to have one config field ("is_printable") under
`ackconfig.FieldConfig` to instruct the code generator to generate
kubebuilder marker comments in order to include the field in `kubectl get` 
response.

Time passed and we are in a situation where want to tune more the way
we print these fields and extend the `kubebuilder:printcolumns` comment
markers to include options like `priority`, `order` and default columns.

This patch reworks the field configuration structure and prepare the
laying ground for future print config fields.

This patch will simplify the realization of the following issues:
* https://github.com/aws-controllers-k8s/community/issues/821
* https://github.com/aws-controllers-k8s/community/issues/822
* https://github.com/aws-controllers-k8s/community/issues/823

### DIFF
--- a/pkg/generate/codedeploy_test.go
+++ b/pkg/generate/codedeploy_test.go
@@ -113,12 +113,12 @@ func TestCodeDeploy_Deployment(t *testing.T) {
 
 	// We marked the fields, "ApplicationName", "DeploymentGroupName",
 	// "DeploymentConfigName and "Description" as printer columns in the
-	// generator.yaml. Let's make sure that they are always returned in sorted
-	// order.
+	// generator.yaml and trimmed the "Name" suffix. Let's make sure that
+	// they are always returned in sorted order.
 	expPrinterColNames := []string{
-		"ApplicationName",
-		"DeploymentConfigName",
-		"DeploymentGroupName",
+		"Application",
+		"DeploymentConfig",
+		"DeploymentGroup",
 		"Description",
 	}
 	gotPrinterCols := crd.AdditionalPrinterColumns()
--- a/pkg/generate/config/field.go
+++ b/pkg/generate/config/field.go
@@ -116,6 +116,16 @@ type CompareFieldConfig struct {
 	NilEqualsZeroValue bool `json:"nil_equals_zero_value"`
 }
 
+// PrintFieldConfig instructs the code generator how to handle kubebuilder:printcolumn
+// comment marker generation. If this struct is not nil, the field will be added to the
+// columns of `kubectl get` response.
+type PrintFieldConfig struct {
+	// Name instructs the code generator to override the column name used to
+	// include the field in `kubectl get` response. This field is generally used
+	// to override very long and redundant columns names.
+	Name string `json:"name"`
+}
+
 // FieldConfig contains instructions to the code generator about how
 // to interpret the value of an Attribute and how to map it to a CRD's Spec or
 // Status field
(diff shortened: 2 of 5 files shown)


A-Hilaly:
@jaypipes @RedbackThomson i'm aware that this is a breaking change, but TBH this is necessary to minimize damage (very verbose and non-understandable config files) in the future.
I made a research and it looks like only `sagemaker-controller` uses "is_printable" field - i'll make sure to make the necessary changes there.

pkg/generate/codedeploy_test.go
```diff
@@ -113,12 +113,12 @@ func TestCodeDeploy_Deployment(t *testing.T) {
 
 	// We marked the fields, "ApplicationName", "DeploymentGroupName",
 	// "DeploymentConfigName and "Description" as printer columns in the
-	// generator.yaml. Let's make sure that they are always returned in sorted
-	// order.
+	// generator.yaml and trimed the "Name" suffix. Let's make sure that
```
jaypipes, line 116: s/trimed/trimmed/ :)

pkg/generate/config/field.go
```diff
@@ -168,4 +170,8 @@ type FieldConfig struct {
 	// Compare instructs the code generator how to produce code that compares
 	// the value of the field in two resources
 	Compare *CompareFieldConfig `json:"compare,omitempty"`
+	// Print instructs the code generator how to generate comment markers that
+	// influence how field are printed in `kubectl get` response. If this field
```
jaypipes, line 174: s/how field/how fields/

pkg/generate/testdata/models/apis/codedeploy/0000-00-00/generator.yaml
```diff
@@ -7,10 +7,13 @@ resources:
...
-        is_printable: true
+        print:
+          name: Application
       DeploymentConfigName:
-        is_printable: true
+        print:
+          name: DeploymentConfig
       Description:
-        is_printable: true
+        print: {}
```
jaypipes, line 19: I'm not a huge fan of this, as I feel it is overly verbose and easy to mess up :)

How about we just remove the `IsPrintable` field and keep the `PrintName` field, but make it a `*string` instead of `string` and just require any printable column to have `PrintName` set to a string value (even if that string value is the same name as the field).

A-Hilaly: The problem is that there other configuration we want to add the `PrinterColumn` generator. Like `order`, `priority`, `type` and probably more in the future.

So long term, instead of having something like:
```yaml
resources:
  Deployment:
    fields:
      DeploymentGroupName:
        is_printable: true
        print_name: DeploymentGroup
        print_order: 3
        print_priority: 1
      DeploymentConfigName:
        is_printable: true
        print_name: DeploymentConfig
        print_order: 2
        print_priority: 1
        print_type: string

```

we can have something like:

```yaml
resources:
  Deployment:
    fields:
      DeploymentGroupName:
        print:
          name: DeploymentGroup
          order: 3
          priority: 1
      DeploymentConfigName:
        print:
          name: DeploymentConfig
          order: 2
          priority: 1
          type: string
```

The second option way less verbose and easier to clearer to read/understand IMO.

jaypipes: OK, you convinced me. :)

jaypipes (approved):

PR #86, opened 2021-06-09
Add a new config to override `kubebuilder:printcolumn` priority

A-Hilaly:
Issue https://github.com/aws-controllers-k8s/community/issues/821

Description of changes:
This patch introduces a new config field called "Standard" under 
`PrintFieldConfig` to indicates whether the column is of priority 0 or 1.

### DIFF
--- a/pkg/generate/config/field.go
+++ b/pkg/generate/config/field.go
@@ -124,6 +124,10 @@ type PrintFieldConfig struct {
 	// include the field in `kubectl get` response. This field is generally used
 	// to override very long and redundant columns names.
 	Name string `json:"name"`
+	// Priority differentiates between fields/columns shown in standard view or wide
+	// view (using the -o wide flag). Fields with priority 0 are shown in standard view.
+	// Fields with priority greater than 0 are only shown in wide view. Default is 0
+	Priority int `json:"priority"`
 }
 
 // FieldConfig contains instructions to the code generator about how
--- a/pkg/model/printer_column.go
+++ b/pkg/model/printer_column.go
@@ -24,6 +24,7 @@ type PrinterColumn struct {
 	CRD      *CRD
 	Name     string
 	Type     string
+	Priority int
 	JSONPath string
 }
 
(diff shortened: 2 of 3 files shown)


pkg/generate/config/field.go
```diff
@@ -124,6 +124,9 @@ type PrintFieldConfig struct {
 	// include the field in `kubectl get` response. This field is generally used
 	// to override very long and redundant columns names.
 	Name string `json:"name"`
+	// Standard indicates whether the column/field should be shown in standard view or
+	// wide view (using the -o wide flag).
+	Standard bool `json:"standard"`
```
A-Hilaly, line 129: This means that users should set `standard: true` in order to make the field printable with a simple `kubectl get`. Maybe i should rename it to `WideView` and inverse the default behavior ?

jaypipes: How about calling the field `Priority` and making it an `*int`. If missing, default is `1` to align with the upstream kubebuilder behaviour.

A-Hilaly: Done

A-Hilaly: I just realized that default in kubebuilder is `0`. But i'm thinking that maybe keeping it to `1` is OK :thinking:

pkg/generate/config/field.go
```diff
@@ -124,6 +124,10 @@ type PrintFieldConfig struct {
 	// include the field in `kubectl get` response. This field is generally used
 	// to override very long and redundant columns names.
 	Name string `json:"name"`
+	// Priority differentiates between fields/columns shown in standard view or wide
+	// view (using the -o wide flag). Fields with priority 0 are shown in standard view.
+	// Fields with priority greater than 0 are only shown in wide view. Default is 1
```
jaypipes, line 129: s/Default is 1/Default is 0/

pkg/model/printer_column.go
```diff
@@ -136,10 +137,18 @@ func (r *CRD) addPrintableColumn(
 		name = field.FieldConfig.Print.Name
 	}
 
+	// Fields/Columns with priority 0 are shown in standard view, and those with
+	// priority greater than 0 are shown only in wide view.
+	priority := 1
```
jaypipes, line 142: Let's use the default of 0, which makes sense... if the user specifies a printcolumn tag, they mean for the column to be shown in standard views.

A-Hilaly: Done by switching to simple `int` type

jaypipes (approved):
:+1:

PR #87, opened 2021-06-10
Extend ack-generate to generate slice with elements of type K8s Secret

kumargauravsharma:
Issue: https://github.com/aws-controllers-k8s/community/issues/828

Description of changes:
With this code change, when a field of type slice
is configured as secret type inside Generator config,
then the generated CRD yaml allows specifying multiple k8s secret values
for that field in input yaml and generated sdk code supplies these values
to the resource API input field as slice type.

Added unit tests to test the scenario.

Details on resultant generated code:
Taking ElastiCache User API has `passwords` filed as example:
ElastiCache User API has `passwords` filed of type `[]*string` in the API input. [Reference API docs](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/elasticache/create-user.html)

When this field is configured as secret type inside Generator config:
```
resources:
  User:
    fields:
      Passwords:
        is_secret: true
```
and service controller code is generated with changes from this PR then,
the generated CRD yaml for password field is:
```
              passwords:
                description: Passwords used for this user. You can create up to two
                  passwords for each user.
                items:
                  description: SecretKeyReference combines a k8s corev1.SecretReference
                    with a specific key within the referred-to Secret
                  properties:
                    key:
                      description: Key is the key within the secret
                      type: string
                    name:
                      description: Name is unique within a namespace to reference
                        a secret resource.
                      type: string
                    namespace:
                      description: Namespace defines the space within which the secret
                        name must be unique.
                      type: string
                  required:
                  - key
                  type: object
                type: array
```
Observe that the password field is `array` type with K8s Secreret as element type

and User resource sdk.go file contain following code for password field while setting the inputs:
```
	if r.ko.Spec.Passwords != nil {
		f3 := []*string{}
		for _, f3iter := range r.ko.Spec.Passwords {
			var f3elem string
			if f3iter != nil {
				tmpSecret, err := rm.rr.SecretValueFromReference(ctx, f3iter)
				if err != nil {
					return nil, err
				}
				if tmpSecret != "" {
					f3elem = tmpSecret
				}
			}
			f3 = append(f3, &f3elem)
		}
		res.SetPasswords(f3)
	}
```
Observe that it supplies the secrets values as slice type to the input API.

Testing:
1. ```make test``` passed for `code-generator` which contained already existing tests for secret fields code generation scenarios for MQ, ElastiCache ReplicationGroup.
2. ```make test``` passed for ElastiCache ACK controller generated code.
3. Generated ElastiCache controller code did not have any unrelated/unexpected changes.

### DIFF
--- a/pkg/generate/code/set_sdk.go
+++ b/pkg/generate/code/set_sdk.go
@@ -236,17 +236,6 @@ func SetSDK(
 		sourceAdaptedVarName += "." + f.Names.Camel
 		sourceFieldPath := f.Names.Camel
 
-		if r.IsSecretField(memberName) {
-			out += setSDKForSecret(
-				cfg, r,
-				memberName,
-				targetVarName,
-				sourceAdaptedVarName,
-				indentLevel,
-			)
-			continue
-		}
-
 		memberShapeRef, _ := inputShape.MemberRefs[memberName]
 		memberShape := memberShapeRef.Shape
 
@@ -339,16 +328,26 @@ func SetSDK(
 				)
 			}
 		default:
-			out += setSDKForScalar(
-				cfg, r,
-				memberName,
-				targetVarName,
-				inputShape.Type,
-				sourceFieldPath,
-				sourceAdaptedVarName,
-				memberShapeRef,
-				indentLevel+1,
-			)
+			if r.IsSecretField(memberName) {
+				out += setSDKForSecret(
+					cfg, r,
+					memberName,
+					targetVarName,
+					sourceAdaptedVarName,
+					indentLevel,
+				)
+			} else {
+				out += setSDKForScalar(
+					cfg, r,
+					memberName,
+					targetVarName,
+					inputShape.Type,
+					sourceFieldPath,
+					sourceAdaptedVarName,
+					memberShapeRef,
+					indentLevel+1,
+				)
+			}
 		}
 		out += fmt.Sprintf(
 			"%s}\n", indent,
@@ -770,6 +769,25 @@ func setSDKForContainer(
 			indentLevel,
 		)
 	default:
+		if r.IsSecretField(sourceFieldPath) {
+			indent := strings.Repeat("\t", indentLevel)
+			// if ko.Spec.MasterUserPassword != nil {
+			out := fmt.Sprintf(
+				"%sif %s != nil {\n",
+				indent, sourceVarName,
+			)
+			out += setSDKForSecret(
+				cfg, r,
+				"",
+				targetVarName,
+				sourceVarName,
+				indentLevel,
+			)
+			// }
+			out += fmt.Sprintf("%s}\n", indent)
+			return out
+		}
+
 		return setSDKForScalar(
 			cfg, r,
 			targetFieldName,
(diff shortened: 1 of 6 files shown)


nmvk (approved):
LGTM

pkg/generate/code/set_sdk.go
```diff
@@ -833,10 +846,17 @@ func setSDKForSecret(
 	//         res.SetMasterUserPassword(tmpSecret)
 	//     }
```
jaypipes, line 847: please update the above code comment to say something like this:

```go
//     if tmpSecret != "" {
//         res.SetMasterUserPassword(tmpSecret)
//     }
//
//    or:
//
//     if tmpSecret != "" {
//         f3elem = tmpSecret
//     }
//
// The second case is used when the SecretKeyReference field
// is a slice of `[]*string` in the original AWS API Input shape.
```

pkg/generate/code/set_sdk.go
```diff
@@ -974,14 +994,16 @@ func setSDKForSlice(
 	//
 	//  f0elem.SetMyField(*f0iter)
 	containerFieldName := ""
+	sourceAttributePath := sourceFieldPath
 	if targetShape.MemberRef.Shape.Type == "structure" {
 		containerFieldName = targetFieldName
+		sourceAttributePath = sourceFieldPath+"."
```
jaypipes, line 1000: do we have a unit test that can stress this code path? I don't believe the unit test about User.Passwords will hit this.

pkg/model/types.go
```diff
@@ -48,20 +49,26 @@ func cleanGoType(
...
+		gtwp = "[]" + mgtwp
 	} else if shape.Type == "timestamp" {
 		// time.Time needs to be converted to apimachinery/metav1.Time
 		// otherwise there is no DeepCopy support
 		gtwp = "*metav1.Time"
 		gte = "metav1.Time"
 		gt = "*metav1.Time"
+	} else if fieldCfg != nil && fieldCfg.IsSecret {
+		gt = "*ackv1alpha1.SecretKeyReference"
+		gte = "SecretKeyReference"
```
jaypipes, line 69: good fix here. ++

jaypipes (changes requested):
This is awesomesauce, @kumargauravsharma :) I have one question and a small suggestion inline, but overall this looks great.

pkg/generate/code/set_sdk.go
```diff
@@ -236,7 +236,13 @@ func SetSDK(
 		sourceAdaptedVarName += "." + f.Names.Camel
 		sourceFieldPath := f.Names.Camel
 
-		if r.IsSecretField(memberName) {
+		memberShapeRef, _ := inputShape.MemberRefs[memberName]
+		memberShape := memberShapeRef.Shape
+
+		if r.IsSecretField(memberName) &&
```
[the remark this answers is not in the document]
kumargauravsharma: Thanks for the suggestion.
I have refactored the code in next revision.

pkg/generate/code/set_sdk.go
kumargauravsharma: Updated code comments.

pkg/generate/code/set_sdk.go
kumargauravsharma: This path is unit tested in `TestSetSDK_MQ_Broker_Create` test for Users field.

jaypipes (approved):
Rock on :)

PR #92, opened 2021-06-17
Add CRD method to get OutputShape

kumargauravsharma:
Issue #, if available:

Description of changes:

This PR also adds a method `GetOutputShape` to crd which provides output shape

For example, `CreateReplicationGroupOutput`, `ModifyReplicationGroupOutput` wrap `ReplicationGroup` object.
This method is useful in creating reusable code which populates the fields from wrapped object to `ko`.

### DIFF
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -14,6 +14,7 @@
 package model
 
 import (
+	"errors"
 	"fmt"
 	"sort"
 	"strings"
@@ -385,6 +386,81 @@ func (r *CRD) GetOutputWrapperFieldPath(
 		return nil
 	}
 	return &opConfig.OutputWrapperFieldPath
+}
+
+// GetOutputShape returns the Output shape for given operation.
+func (r *CRD) GetOutputShape(
+	// The operation to look for the Output shape
+	op *awssdkmodel.Operation,
+) (*awssdkmodel.Shape, error) {
+	if op == nil {
+		return nil, errors.New("no output shape for nil operation")
+	}
+
+	outputShape := op.OutputRef.Shape
+	if outputShape == nil {
+		return nil, errors.New("output shape not found")
+	}
+
+	// We might be in a "wrapper" shape. Unwrap it to find the real object
+	// representation for the CRD's createOp/DescribeOP.
+
+	// Use the wrapper field path if it's given in the ack-generate config file.
+	wrapperFieldPath := r.GetOutputWrapperFieldPath(op)
+	if wrapperFieldPath != nil {
+		wrapperOutputShape, err := r.GetWrapperOutputShape(outputShape, *wrapperFieldPath)
+		if err != nil {
+			return nil, fmt.Errorf("unable to unwrap the output shape: %v", err)
+		}
+		outputShape = wrapperOutputShape
+	} else {
+		// If the wrapper field path is not specified in the config file and if
+		// there is a single member shape and that member shape is a structure,
+		// unwrap it.
+		if outputShape.UsedAsOutput && len(outputShape.MemberRefs) == 1 {
+			for _, memberRef := range outputShape.MemberRefs {
+				if memberRef.Shape.Type == "structure" {
+					outputShape = memberRef.Shape
+				}
+			}
+		}
+	}
+	return outputShape, nil
+}
+
+// GetWrapperOutputShape returns the shape of the last element of a given field
+// Path. It carefully unwraps the output shape and verifies that every element
+// of the field path exists in their correspanding parent shape and that they are
+// structures.
+func (r *CRD) GetWrapperOutputShape(
+	shape *awssdkmodel.Shape,
+	fieldPath string,
+) (*awssdkmodel.Shape, error) {
+	if fieldPath == "" {
+		return shape, nil
+	}
+	fieldPathParts := strings.Split(fieldPath, ".")
+	for x, wrapperField := range fieldPathParts {
+		for memberName, memberRef := range shape.MemberRefs {
+			if memberName == wrapperField {
+				if memberRef.Shape.Type != "structure" {
+					// All the mentionned shapes must be structure
+					return nil, fmt.Errorf(
+						"Expected SetOutput.WrapperFieldPath to only contain fields of type 'structure'."+
+							" Found %s of type '%s'",
+						memberName, memberRef.Shape.Type,
+					)
+				}
+				remainPath := strings.Join(fieldPathParts[x+1:], ".")
+				return r.GetWrapperOutputShape(memberRef.Shape, remainPath)
+			}
+		}
+		return nil, fmt.Errorf(
+			"Incorrect SetOutput.WrapperFieldPath. Could not find %s in Shape %s",
+			wrapperField, shape.ShapeName,
+		)
+	}
+	return shape, nil
 }
 
 // GetCustomImplementation returns custom implementation method name for the
--- a/pkg/generate/code/set_resource.go
+++ b/pkg/generate/code/set_resource.go
@@ -125,7 +125,7 @@ func SetResource(
 	// Use the wrapper field path if it's given in the ack-generate config file.
 	wrapperFieldPath := r.GetOutputWrapperFieldPath(op)
 	if wrapperFieldPath != nil {
-		outputShape, err = GetWrapperOutputShape(outputShape, *wrapperFieldPath)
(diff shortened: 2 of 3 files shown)


kumargauravsharma:
Related PR: aws-controllers-k8s/elasticache-controller#36 where dynamic hook template is used to auto generate reusable setOutput method.

pkg/model/crd.go
```diff
@@ -387,6 +387,34 @@ func (r *CRD) GetOutputWrapperFieldPath(
...
+	// shape and that member shape is a structure, unwrap it.
+	if outputShape.UsedAsOutput && len(outputShape.MemberRefs) == 1 {
+		for _, memberRef := range outputShape.MemberRefs {
+			if memberRef.Shape.Type == "structure" {
+				outputShape = memberRef.Shape
+				break
+			}
+		}
+	}
+	return &outputShape.ShapeName
```
A-Hilaly, line 415: what if `memberRef.Shape.Type` is not "structure" ? L415 will panic
Maybe we can return an error when `outputShape` is not found ?

pkg/model/crd.go
```diff
@@ -387,6 +387,34 @@ func (r *CRD) GetOutputWrapperFieldPath(
...
+	}
+
+	// We might be in a "wrapper" shape. Unwrap it to find the real object
+	// representation for the CRD's op. If there is a single member
+	// shape and that member shape is a structure, unwrap it.
+	if outputShape.UsedAsOutput && len(outputShape.MemberRefs) == 1 {
+		for _, memberRef := range outputShape.MemberRefs {
+			if memberRef.Shape.Type == "structure" {
+				outputShape = memberRef.Shape
+				break
```
A-Hilaly, line 411: nit: unnecessary break

templates/pkg/resource/sdk.go.tpl
```diff
@@ -335,4 +335,7 @@ func (rm *resourceManager) handleImmutableFieldsChangedCondition(
 
 	return &resource{ko}
 }
-{{- end }}
\ No newline at end of file
+{{- end }}
+{{- if $hookCode := Hook .CRD "sdk_addon" }}
+{{ $hookCode }}
+{{- end }}
```
A-Hilaly, line 341: Can you put this in a separate PR or commit ?

pkg/model/crd.go
```diff
@@ -387,6 +387,34 @@ func (r *CRD) GetOutputWrapperFieldPath(
 	return &opConfig.OutputWrapperFieldPath
 }
 
+// OutputShapeName returns name of the Output shape for given operation.
+func (r *CRD) UnwrappedOutputShapeName(
+	// The operation to look for the Output shape
+	op *awssdkmodel.Operation,
+) *string {
```
A-Hilaly, line 394: Thinking more about this, i believe we should return the `outputShape` instead of the `outputShape.Name`. This way we can reuse this function other places as well like: https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/code/set_resource.go#L138-L145

pkg/model/crd.go
```diff
@@ -387,6 +387,34 @@ func (r *CRD) GetOutputWrapperFieldPath(
 	return &opConfig.OutputWrapperFieldPath
 }
 
+// OutputShapeName returns name of the Output shape for given operation.
```
jaypipes, line 390: mismatched function name here... plus, I agree with @A-Hilaly about making the function return a Shape instead of a string name. Maybe call this method `GetOutputShape()`?

pkg/model/crd.go
jaypipes: > what if `memberRef.Shape.Type` is not "structure" ? L415 will panic
> Maybe we can return an error when `outputShape` is not found ?

Agreed. This should be `return nil` and the above line 410 should be `return memberRef.Shape`

templates/pkg/resource/sdk.go.tpl
jaypipes: agree with @A-Hilaly this should go in a separate PR (and I'm not sold on the idea of having a generic "sdk_addon" here. The hooks code was meant for code injection in specific places in the processing pipeline, not really as a way to add custom code outside of the processing pipeline...

jaypipes (changes requested):

pkg/model/crd.go
kumargauravsharma: Thanks for the comments, I have updated the code.
I referred latest set_resource.go logic and moved reusable method into CRD.

pkg/model/crd.go
kumargauravsharma: Updated the code.

pkg/model/crd.go
kumargauravsharma: Changed the return type. However, did not update the set_resource.go as that also sets `sourceVarName` var thus would need to keep its logic.

pkg/model/crd.go
kumargauravsharma: Updated code.

templates/pkg/resource/sdk.go.tpl
kumargauravsharma: Created separate PR for it: https://github.com/aws-controllers-k8s/code-generator/pull/94

pkg/model/crd.go
```diff
@@ -387,6 +387,82 @@ func (r *CRD) GetOutputWrapperFieldPath(
...
+	// representation for the CRD's createOp/DescribeOP.
+
+	// Use the wrapper field path if it's given in the ack-generate config file.
+	wrapperFieldPath := r.GetOutputWrapperFieldPath(op)
+	if wrapperFieldPath != nil {
+		wrapperOutputShape, err := r.GetWrapperOutputShape(outputShape, *wrapperFieldPath)
+		if err != nil {
+			msg := fmt.Sprintf("Unable to unwrap the output shape: %v", err)
+			panic(msg)
+		}
```
A-Hilaly, line 414: Can we wrap and return an error instead? All the panic calls are made inside code generators functions - IMO CRD methods should just send simple errors

kumargauravsharma: Updated the code. Code now returns `nil`.

A-Hilaly: Why does it return `nil` @kumargauravsharma ? Unless i'm missing something i think it should just return a wrapped error like: `fmt.Errorf("unable to unwrap the output shape: %v", err)`

A-Hilaly: Ohhh i'm missing the fact that this function doesn't return an error.. I think that it would be nice to return an error anyways for debuging purposes

kumargauravsharma: Changed method return type to include error as well.

pkg/model/crd.go
```diff
@@ -387,6 +388,81 @@ func (r *CRD) GetOutputWrapperFieldPath(
...
 }
 
+// GetOutputShape returns the Output shape for given operation.
+func (r *CRD) GetOutputShape(
+	// The operation to look for the Output shape
+	op *awssdkmodel.Operation,
+) (*awssdkmodel.Shape, error) {
+	if op == nil {
+		return nil, nil
+	}
```
A-Hilaly, line 398: We don't need this block - this means that even if `err` is nil we should perform a "nillity" test on the shape as well. We just have to make sure that either we return and nil `error` or a nil `shape` but never both at the same in the logic bellow.

pkg/model/crd.go
```diff
@@ -387,6 +388,81 @@ func (r *CRD) GetOutputWrapperFieldPath(
...
+
+	// We might be in a "wrapper" shape. Unwrap it to find the real object
+	// representation for the CRD's createOp/DescribeOP.
+
+	// Use the wrapper field path if it's given in the ack-generate config file.
+	wrapperFieldPath := r.GetOutputWrapperFieldPath(op)
+	if wrapperFieldPath != nil {
+		wrapperOutputShape, err := r.GetWrapperOutputShape(outputShape, *wrapperFieldPath)
+		if err != nil {
+			return nil, err
```
A-Hilaly, line 413:
```suggestion
			return nil, fmt.Errorf("unable to unwrap the output shape: %v", err)
```

pkg/model/crd.go
kumargauravsharma: Thanks for the comment, I have updated the code.

pkg/model/crd.go
kumargauravsharma: Thanks for the comment, I have updated the code.

A-Hilaly (approved):
NIce, thanks @kumargauravsharma !:+1:

PR #94, opened 2021-06-17
Add sdk addon hook

kumargauravsharma:
Issue #, if available:

Ability for service controllers to add code (additional functions, variables etc as needed) to sdk.go file using hooks as extension mechanism.

Description of changes:

This PR adds `sdk_addon` hook to `sdk.go.tpl`. Using this hook, service controllers can append code to a custom resource's `sdk.go` file. It will help in placing reusable code/functions which are generated dynamically using hook templates.

Following PR contains changes showcasing how this facility helps service controllers auto generate reusable code: https://github.com/aws-controllers-k8s/elasticache-controller/pull/36

### DIFF
--- a/pkg/generate/ack/hook.go
+++ b/pkg/generate/ack/hook.go
@@ -56,6 +56,7 @@ code paths:
 * sdk_get_attributes_post_set_output
 * sdk_create_post_set_output
 * sdk_update_post_set_output
+* sdk_file_end
 * delta_pre_compare
 * delta_post_compare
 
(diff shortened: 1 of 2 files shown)


kumargauravsharma:
> Maybe we can optionally generate a `sdk_addons.go` files?

I think it would complicate the code generator implementation as it would need to decide if this file needs to be generated or skipped.
Currently, the list of files for CR is a fixed list: [code reference](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/ack/controller.go#L151-L159)

jaypipes (changes requested):
Two things I'm requesting:

1) Let's call this `sdk_file_end` instead of `sdk_addon` please. The reason is twofold: all the other hook points describe where in the file or code flow they are. Secondly, the term "addon" [means something specififc in Kubernetes-speak](https://kubernetes.io/docs/concepts/cluster-administration/addons/) so I want to avoid confusion with that.

2) Please add a callout to this new hook point in the docstring right below here:

https://github.com/aws-controllers-k8s/code-generator/blob/c77aa9c75d944952dee198029ba9822691cd82b0/pkg/generate/ack/hook.go#L58

That basically states `sdk_file_end` is a generic hook point that occurs outside the scope of any specific `AWSResourceManager` method and can be used to place commonly-generated code inside the `sdk.go` file.

Thank you @kumargauravsharma!

kumargauravsharma:
Thanks @jaypipes for your comments.
I have updated the code files per the comments.

jaypipes (approved):
Sweet. :+1:

PR #97, opened 2021-06-18
Support suffix exception

ryansteakley:
Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/840

Description of changes:
Add support for suffix exception checking. Also added Sagemaker tests for prefix and suffix exceptions.

### DIFF
--- a/pkg/generate/code/check.go
+++ b/pkg/generate/code/check.go
@@ -24,16 +24,17 @@ import (
 	"github.com/aws-controllers-k8s/code-generator/pkg/names"
 )
 
-// CheckExceptionMessagePrefix returns Go code that contains a condition to
-// check if the message_prefix specified for a particular HTTP status code in
+// CheckExceptionMessage returns Go code that contains a condition to
+// check if the message_prefix/message_suffix specified for a particular HTTP status code in
 // generator config is a prefix for the exception message returned by AWS API.
-// If message_prefix field was not specified for this HTTP code in generator
+// If message_prefix/message_suffix field was not specified for this HTTP code in generator
 // config, we return an empty string
 //
 // Sample Output:
 //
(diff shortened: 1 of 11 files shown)


pkg/generate/code/check.go
```diff
@@ -45,6 +46,10 @@ func CheckExceptionMessagePrefix(
 			return fmt.Sprintf("&& strings.HasPrefix(awsErr.Message(), \"%s\") ",
 				*excConfig.MessagePrefix)
 		}
+		if ok && excConfig.MessageSuffix != nil {
```
RedbackThomson, line 49: nit: Replace the `if ok` in each of these conditional blocks with an early return at the top:
```go
if !ok {
  return ""
}
```

RedbackThomson (approved):
Yeah, really like these changes!

pkg/generate/code/check.go
ryansteakley: replaced

pkg/generate/code/check.go
```diff
@@ -24,27 +24,35 @@ import (
...
-// If message_prefix field was not specified for this HTTP code in generator
+// If message_prefix/message_suffix field was not specified for this HTTP code in generator
 // config, we return an empty string
 //
 // Sample Output:
 //
 // && strings.HasPrefix(awsErr.Message(), "Could not find model")
-func CheckExceptionMessagePrefix(
+// && strings.HasSuffix(awsErr.Message(), "does not exist.")
+func CheckExceptionMessageAffix(
```
jaypipes, line 37: We could also just call this `CheckExceptionMessage` :)

jaypipes (approved):
Great stuff @ryansteakley :)

jaypipes (approved):

PR #98, opened 2021-06-20
trace sdk.go resource manager methods

jaypipes:
Updates the sdk.go-specific templates to take advantage of the new
ACK runtime logger that is embedded in all Context structs passed from
the ACK runtime v0.2.3 release and after.

In working these log traces into the various sdkXXX methods, I took this
opportunity to clean up some of the ugly template conditionals in these methods
that checked to see whether an Output shape was being referenced as a
variable after calling the aws-sdk-go API calls and setting the
associated variable name to `_` in order to prevent `unused variable
declared` compilation failures.

The template code for handling this used to look like this:

```
{{ $createCode := GoCodeSetCreateOutput .CRD "resp" "ko" 1 false }}
	{{ if not ( Empty $createCode ) }}resp{{ else }}_{{ end }}, respErr := rm.sdkapi.{{ .CRD.Ops.Create.ExportedName }}WithContext(ctx, input)
	rm.metrics.RecordAPICall("CREATE", "{{ .CRD.Ops.Create.ExportedName }}", respErr)
	if respErr != nil {
		return nil, respErr
	}
	// Merge in the information we read from the API call above to the copy of
	// the original Kubernetes object we passed to the function
	ko := r.ko.DeepCopy()
{{ $createCode }}
```

but now is simpler and easier to read:

```
	var resp {{ .CRD.GetOutputShapeGoType .CRD.Ops.Create }}
	resp, err = rm.sdkapi.{{ .CRD.Ops.Create.ExportedName }}WithContext(ctx, input)
	rm.metrics.RecordAPICall("CREATE", "{{ .CRD.Ops.Create.ExportedName }}", err)
	if err != nil {
		return nil, err
	}
	// Merge in the information we read from the API call above to the copy of
	// the original Kubernetes object we passed to the function
	ko := desired.ko.DeepCopy()
{{ GoCodeSetCreateOutput .CRD "resp" "ko" 1 false }}
```

### DIFF
--- a/templates/pkg/resource/sdk.go.tpl
+++ b/templates/pkg/resource/sdk.go.tpl
@@ -9,12 +9,14 @@ import (
 	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
 	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
 	ackerr "github.com/aws-controllers-k8s/runtime/pkg/errors"
+	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
 	"github.com/aws/aws-sdk-go/aws"
 	svcsdk "github.com/aws/aws-sdk-go/service/{{ .ServiceIDClean }}"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	svcapitypes "github.com/aws-controllers-k8s/{{.ServiceIDClean }}-controller/apis/{{ .APIVersion }}"
+	svcsdkapi "github.com/aws/aws-sdk-go/service/{{ .ServiceIDClean }}"
 )
 
 // Hack to avoid import errors during build...
@@ -26,6 +28,7 @@ var (
 	_ = &svcapitypes.{{ .CRD.Names.Camel }}{}
 	_ = ackv1alpha1.AWSAccountID("")
 	_ = &ackerr.NotFound
+	_ = svcsdkapi.New
 )
 
 // sdkFind returns SDK-specific information about a supplied resource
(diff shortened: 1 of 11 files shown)


templates/pkg/resource/sdk.go.tpl
```diff
@@ -26,6 +28,7 @@ var (
 	_ = &svcapitypes.{{ .CRD.Names.Camel }}{}
 	_ = ackv1alpha1.AWSAccountID("")
 	_ = &ackerr.NotFound
+	_ {{ .CRD.GetOutputShapeGoType .CRD.Ops.Create }} = nil
```
A-Hilaly, line 31: nit: `_ = svcsdkapi.New` could work as well

A-Hilaly (approved):
:+1:

A-Hilaly:
We can also start bumping [`VERSION`](https://github.com/aws-controllers-k8s/code-generator/blob/main/Makefile#L9) whenever we bring in a new runtime version in `go.mod`

templates/pkg/resource/sdk.go.tpl
jaypipes: Nice suggestion. Done.

templates/pkg/resource/sdk.go.tpl
```diff
@@ -40,52 +43,57 @@ var (
 {{- end }}
 
 // sdkCreate creates the supplied resource in the backend AWS service API and
-// returns a new resource with any fields in the Status field filled in
+// returns a copy of the resource with resource fields (in both Spec and
```
vijtrip2, line 46: We discussed this earlier that currently the spec fields are not updated for sdkCreate. We plan to allow it but currently this godoc is inconsistent with behavior.

"{{ GoCodeSetCreateOutput .CRD "resp" "ko" 1 false }}" -> the false parameter here does not allow spec update.

vijtrip2: As part of conversation on meeting, this godoc can be updated later.

surajkota:
Nice!!

PR #101, opened 2021-06-22
prevent unused variable error

jaypipes:
A [previous commit][0] that changed the ACK pkg/resource templates
introduced an error for controllers that have CREATE or UPDATE
implementations where the Output shape from the operation sets no fields
on the CR.

In these cases, attempting to build these controllers was resulting in a
build failure, like this for the RDS controller:

```
pkg/resource/db_parameter_group/sdk.go:205:6: resp declared but not used
```

This corrects that by using a blanking technique immediately after the
variable declaration:

```
var resp *SomeSDKShape; _ = resp;
```

[0]: a42b4a1a9081824ff60a29b8a5f31ccf79026372

### DIFF
--- a/templates/pkg/resource/sdk.go.tpl
+++ b/templates/pkg/resource/sdk.go.tpl
@@ -70,7 +70,7 @@ func (rm *resourceManager) sdkCreate(
 {{ $hookCode }} 
 {{- end }}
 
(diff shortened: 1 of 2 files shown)


A-Hilaly (approved):
Nice trick!

PR #109, opened 2021-06-25
Copy `generator.yaml` to apis/version directory

A-Hilaly:
Part of aws-controllers-k8s/community#835

Adding some logic to create copies of the `generator.yaml` within the
apis versions directories. These generator.yaml files we be reused by
the conversion webhook generator to understand the difference between
each two different API versions.

This patch Also adds a new utility function to copy files from a
source to a destination path.

### DIFF
--- a/cmd/ack-generate/command/apis.go
+++ b/cmd/ack-generate/command/apis.go
@@ -24,6 +24,7 @@ import (
 	generate "github.com/aws-controllers-k8s/code-generator/pkg/generate"
 	ackgenerate "github.com/aws-controllers-k8s/code-generator/pkg/generate/ack"
 	"github.com/aws-controllers-k8s/code-generator/pkg/model"
+	"github.com/aws-controllers-k8s/code-generator/pkg/util"
 )
 
 type contentType int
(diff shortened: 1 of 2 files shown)


cmd/ack-generate/command/apis.go
```diff
@@ -65,7 +66,20 @@ func saveGeneratedMetadata(cmd *cobra.Command, args []string) error {
...
+
+	copyDest := filepath.Join(
+		optOutputPath, "apis", optGenVersion, "generator.yaml",
+	)
+	err = util.CopyFile(optGeneratorConfigPath, copyDest)
+	if err != nil {
+		return fmt.Errorf("cannot copy generator configuration file: %v", err)
+	}
+
+	fmt.Println("copied to", copyDest)
```
RedbackThomson, line 81: Debugging mark?

RedbackThomson (changes requested):

cmd/ack-generate/command/apis.go
A-Hilaly: Yes :) :facepalm:

PR #112, opened 2021-06-29
Add new type to OperationConfig so user can implement CustomCheckRequiredFieldsMissingMethod 

ryansteakley:
Issue #, if available:
[#850](https://github.com/aws-controllers-k8s/community/issues/850)

Description of changes:
Adds new `pkg/generate/config.OperationConfig.CustomCheckRequiredFieldsMissingMethod `field of type *string  so that user can implement a `CustomCheckRequiredFieldsMissingMethod` method as code-generator cannot handle certain cases. 

ex. For the `SageMaker` resource `ModelPackage`. There are versioned `modelPackage` and unversioned `modelPackage`, they both have a different required field being `ModelPackageGroupName` and `ModelPackageName`. `requiredFieldsMissingFromReadOneInput` will by default auto-generate to check for `ModelPackageName` and will infinitely recreate since it will never find the required field when a user attempts to make a versioned `modelPackage` which is `modelPackageGroupName` as it does not contain a `ModelPackageName`. ModelPackage has one or the other for its required field but not both.

This new field allows the user to create their own custom implementation for `requiredFieldsMissingFromReadOneInput`,  `requiredFieldsMissingFromSetAttributesInput`, and  `requiredFieldsMissingFromGetAttributesInput`so that in cases where the auto-generated default required fields do not handle cases where it may be complicated ex: Have an exclusive-or of required fields.


RedbackThomson:
/ok-to-test

Hey Ryan, would you mind expanding on what cases this is hoping to cover - please update the PR description. Some examples about how hooks can/will be used is always useful when other services may require them in the future.

jaypipes (changes requested):
I should have been more explicit in my comments on aws-controllers-k8s/community#850.

The generic hook code locations are for code that does not *replace* the existing code paths but rather allow for individual service controllers to modify the behaviour of those existing code paths (either before or after important steps in the processing pipeline).

The generic hook code locations are not appropriate for this specific use case, since what is needed is literally to *replace* the existing code path for the `checkRequiredFieldsMissing` implementation with an entirely different implementation.

What is needed instead of the generic hook code location is something more akin to the `pkg/generate/config.OperationConfig:CustomImplementation` field:

https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/config/operation.go#L25

that we can use to entirely override the implementation of the `checkRequiredFieldsMissing` function.

Basically, instead of adding a new generic hook code location called `custom_required_fields_missing`, I am asking that you add a new `pkg/generate/config.OperationConfig.CustomCheckRequiredFieldsMissingMethod` field of type `*string` that, if present, you can use like you've done in the `sdk_find_read_one.go.tpl` template file below, but like this:

```
{{- if $customCheckMethod := .CRD.CustomMethodCheckRequiredFieldsMissing }}
return rm.{{ $customCheckMethod }}(r)
{{- else }}
{{ GoCodeRequiredFieldsMissingFromReadOneInput .CRD "r.ko" 1 }}
{{- end }}
```

pkg/model/crd.go
```diff
@@ -507,6 +507,21 @@ func (r *CRD) UpdateConditionsCustomMethodName() string {
 	return resGenConfig.UpdateConditionsCustomMethodName
 }
 
+// CustomCheckRequiredFieldsMissingMethod returns custom check required fields missing method
+// as *string for custom resource, if specified in generator config
```
RedbackThomson, line 511: nit: `as string`

pkg/model/crd.go
```diff
@@ -507,6 +507,21 @@ func (r *CRD) UpdateConditionsCustomMethodName() string {
...
+func (r *CRD) CustomCheckRequiredFieldsMissingMethod(op *awssdkmodel.Operation) string {
```
RedbackThomson, line 512: nit: Other methods don't add `Custom` prefix to this method, but use `Get` instead. `GetCheckRequiredFieldsMissingMethod` for example.

RedbackThomson (approved):

pkg/model/crd.go
ryansteakley: I see `customimplementation` which inspired this is written as `GetCustomImplementation`. What do you think about `GetCustomCheckRequiredFieldsMissingMethod` ?

pkg/model/crd.go
ryansteakley: thanks for spotting this

pkg/model/crd.go
RedbackThomson: Oh I see, yes!

RedbackThomson (approved):

jaypipes (approved):
Excellent, thanks @ryansteakley!

PR #114, opened 2021-07-02
Returning the latest resource from the Delete() call

kumargauravsharma:
Issue #, if available:

aws-controllers-k8s/community#836

Description of changes:

This PR is to related to https://github.com/aws-controllers-k8s/runtime/pull/21
It makes the code-generator changes for delete api as discussed under: https://github.com/aws-controllers-k8s/runtime/pull/21#discussion_r655729649

Testing:
* `make test` passed.

### DIFF
--- a/templates/pkg/resource/manager.go.tpl
+++ b/templates/pkg/resource/manager.go.tpl
@@ -129,11 +129,12 @@ func (rm *resourceManager) Update(
 }
 
 // Delete attempts to destroy the supplied AWSResource in the backend AWS
-// service API.
+// service API, returning an AWSResource representing the
+// resource being deleted (if delete is asynchronous and takes time)
 func (rm *resourceManager) Delete(
 	ctx context.Context,
 	res acktypes.AWSResource,
-) error {
+) (acktypes.AWSResource, error) {
 	r := rm.concreteResource(res)
 	if r.ko == nil {
 		// Should never happen... if it does, it's buggy code.
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ GO111MODULE=on
 AWS_SERVICE=$(shell echo $(SERVICE) | tr '[:upper:]' '[:lower:]')
 
 # Build ldflags
-VERSION ?= "v0.4.0"
+VERSION ?= "v0.5.0"
 GITCOMMIT=$(shell git rev-parse HEAD)
 BUILDDATE=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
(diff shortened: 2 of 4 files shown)


surajkota (approved):

templates/pkg/resource/manager.go.tpl
```diff
@@ -129,11 +129,12 @@ func (rm *resourceManager) Update(
...
 
 // Delete attempts to destroy the supplied AWSResource in the backend AWS
-// service API.
+// service API, returning an AWSResource representing the
+// resource being deleted (if delete is asynchronous and takes time)
 func (rm *resourceManager) Delete(
 	ctx context.Context,
 	res acktypes.AWSResource,
-) error {
+) (acktypes.AWSResource, error) {
```
jaypipes, line 137: you cannot change this method signature before changing the interface in ACK runtime:

https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/types/aws_resource_manager.go#L61-L63

jaypipes (changes requested):
See inline... we will need a change to ACK runtime first (the interface for AWSResourceManager needs to be updated before the templates here).

Also, please change the PR summary to reflect what this is doing (returning the latest resource from the Delete() call, not "Provide custom resource delete progress for long-running delete")

Thanks!
-jay

templates/pkg/resource/manager.go.tpl
kumargauravsharma: PR: https://github.com/aws-controllers-k8s/runtime/pull/21 contained changes in runtime.

kumargauravsharma: Updated the PR summary as: "Returning the latest resource from the Delete() call"

jaypipes (changes requested):
OK, I cut https://github.com/aws-controllers-k8s/runtime/releases/tag/v0.5.0 with the changes to the `AWSResourceManager.Delete` interface method. Please update code-gen's `go.mod` file accordingly in this PR.

kumargauravsharma:
Updated `code-gen` `go.mod` file to refer `runtime` `0.5.0`

jaypipes (approved):
:+1:

A-Hilaly:
@kumargauravsharma @jaypipes we also need to set the version to 0.5.0 in the Makefile https://github.com/aws-controllers-k8s/code-generator/blob/main/Makefile#L9

kumargauravsharma:
@A-Hilaly thanks. I have updated the Makefile.

PR #117, opened 2021-07-05
handle renamed fields for resource identifiers

jaypipes:
The code that outputs Go code that sets `Spec` or `Status` fields to a
resource identifier field was not accounting for cases where we renamed
the field. Such is the case for many of the RDS controller's resources,
since all the identifier fields for all resources are "de-stuttured" by
renaming, for example, `DBSubnetGroupName` to just `Name` or
`DBClusterParameterGroupName` to just `Name`.

This patch simply ensures we account for renamed fields in the Spec and
Status during calls to `pkg/generate/code.SetResourceIdentifiers` and
adds a unit test to ensure we don't regress in the future.

Issue aws-controllers-k8s/community#852

### DIFF
--- a/pkg/generate/code/set_resource.go
+++ b/pkg/generate/code/set_resource.go
@@ -922,21 +922,25 @@ func SetResourceIdentifiers(
 			continue
 
 		}
+		// Check that the field has potentially been renamed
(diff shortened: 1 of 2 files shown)


A-Hilaly (approved):
:+1:

PR #118, opened 2021-07-05
pin code-gen to ACK runtime v0.4.0

jaypipes:
Adds a simple compile-time test for pinning code-gen to the ACK runtime
v0.4.0 so that we can bring the code generated by these templates
in-line with the ACK runtime v0.4.0 (which added support for
AWSIdentifiers.AdditionalKeys and the pkg/webhook stuff referred to from
templates/cmd/controller/main.go.tpl)


A-Hilaly (approved):
:+1:

PR #119, opened 2021-07-05
Support nested Ignore.FieldPaths

chlunde:
The current implementation of IgnoreFields did not support nested
shapes. For Ignore.FieldPaths: Foo.Bar.Baz, it would ignore the whole of
Bar.

With this change, it supports multiple levels of nested fields.

A practical example for this is ses CreateTemplateInput.Template.TemplateName.


pkg/model/model.go
```diff
@@ -664,13 +664,23 @@ func (m *Model) ApplyShapeIgnoreRules() {
 		return
 	}
 	for sdkShapeID, shape := range m.SDKAPI.API.Shapes {
+	ignoreFields:
```
chlunde, line 667: I don't like this, I can extract this function later... :palm_tree:

A-Hilaly: @chlunde checkout this very similar function :) https://github.com/aws-controllers-k8s/code-generator/pull/78/files#diff-f5c09cb6b638098a0e513f710e901c9f7381165e4e2c49f2b3881613d2db518dR324

jaypipes: @chlunde ping! :) Did you have a chance to check out the code and comment from @A-Hilaly above?

chlunde: back from vacation today! I'll try to fix it by tomorrow

pkg/model/model_eks_test.go
```diff
@@ -0,0 +1,48 @@
...
+	crds, err := g.GetCRDs()
+	require.Nil(err)
+
+	crd := getCRDByName("Nodegroup", crds)
+	require.NotNil(crd)
+
+	// fields we have ignored via generator.yaml
+	ignoredFieldPaths := []string{
+		"Version",
+		"ScalingConfig.DesiredSize",
```
A-Hilaly, line 38: Should'nt these be more `Nodegroup.Version` and `CreateNodegroupInput.ScalingConfig.DesiredSize` ?

pkg/model/model.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/aws-controllers-k8s/code-generator/pkg/generate/templateset"
 	"github.com/aws-controllers-k8s/code-generator/pkg/names"
 	"github.com/aws-controllers-k8s/code-generator/pkg/util"
+	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
```
A-Hilaly, line 26: nit: `goimports` can help with separating internal from external imports

A-Hilaly (commented):
LGTM overall :+1: I left two comments below.

pkg/model/model.go
chlunde: I think `goimports` places it the same place, it was placed there by `gopls`. I tried deleting the line and by running `goimports -w ./pkg/model/model.go` I get the same result.

The sibling file
https://github.com/aws-controllers-k8s/code-generator/blob/2f11b23b1618466348fc850f9302c5ae3441580a/pkg/model/field.go#L19-L22 is also the same. But you're right, I think I've seen a convention where stdlib packages have one group, same-module has another group and all others have a third group.

pkg/model/model_eks_test.go
chlunde: This is already under `getCRDByName("Nodegroup", ..)`, so logically it is now `Nodegroup.ScalingConfig.DesiredSize`.

But yeah, if we could rewrite it to a unit test then maybe it would look like that, or maybe we should test it through another code path than `getCRDByName`?

pkg/testdata/models/apis/eks/0000-00-00/generator.yaml
```diff
@@ -1,3 +1,10 @@
+ignore:
+  field_paths:
+    # Test nested field path of multiple levels
+    - Nodegroup.Version
+    - CreateNodegroupInput.Version
+    - CreateNodegroupInput.ScalingConfig.DesiredSize
```
muvaf, line 6: Quoting from the docstring:
```
// Set of field paths to ignore. The name here should be the original name of
// the field as it appears in AWS SDK objects. You can refer to a field by
// giving its "<shape_name>.<field_name>". For example, "CreateApiInput.Name".
```

Would it be possible to use `ScalingConfig.DesiredSize` (assuming `ScalingConfig` is the name of the type) to achieve what you'd like?

chlunde: Yeah, that works :facepalm: Thanks! :D

I was trying to help @ikarldasan on slack to get progress on https://github.com/crossplane/provider-aws/pull/755 - but I didn't know how the generator config worked and dug into this hole.

So for that case, to ignore TemplateName, it would simply be:

```yaml
ignore:
  field_paths:
    - Template.TemplateName
```

I though that was tested but apparently not :)

PR #120, opened 2021-07-06
Refactor `pkg/generate` package

A-Hilaly:
Issue N/A

During `code-generator` early days, `generate.Generator` used to have
methods that generated controllers, apis and resources code. After few
refactoring iterations the code that outputs controllers Go code is
now situated in `pkg/generate/ack`, and `generate.Generator` only
contains methods to read/load CRDs and Type Definitions.

This PR contains two commits. The first one renames `pkg/generate.Generator`
to `pkg/model.Model` and the second moves `generator` unit tests to `pkg/model`
and `pkg/generate/testdata` to `pkg/testdata`

Description of changes:

- Rename `pkg/generate.Generator` to `pkg/model.Model`, also renames
all the variables and parameters that were used to represent a generator.    
- Move `pkg/generate/testdata` to `pkg/testdata`
- Move all `pkg/generate` unit tests to `pkg/model`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #121, opened 2021-07-06
Bring in `pkg/model/multiversion` package to help interacting with multiple API versions

A-Hilaly:
Part of https://github.com/aws-controllers-k8s/community/issues/835

Description of changes:
This patch brings a new package to `pkg/model`, essentially to
simplify accessing and interacting with multiple api versions.

The `pkg/model/multiversion` contains five main components:
- An `APIVersionManager` struct containing information about the hub 
and spokes apis, their corresponding `ackmodel.Model` and
`APIInfo`, plus a list of the deprecated and removed versions.
- A `FieldDelta`, `CRDDelta` and `FiledChangeTypes` which helps in
representing the difference between two different APIs.
- A `ComputeFieldsDiff` function which helps in computing the difference
between two given string to `ackmodel.Fields` maps.
- An `APIStatus` and `APIInfo` that holds information about a specific
api version (used sdk version, generator file path etc...)
- Add `ComputeRenamesDelta` to compte the correct field renames
between two api versions.

this PRs also changes `model.SDKHelper` methods and signatures to
support interacting with multiple "Service API Versions" and multiple sdk 
versions


A-Hilaly:
NOTE to reviews: This will need a rebase after #120.
/hold

pkg/model/multiversion/api_info.go
```diff
@@ -0,0 +1,38 @@
...
+// TODO(a-hilaly) store API status in `ack-generate-metadata.yaml`
+
+type APIStatus string
+
+const (
+	APIStatusUnknown    APIStatus = "unknown"
+	APIStatusAvailable            = "available"
+	APIStatusRemoved              = "removed"
+	APIStatusDeprecated           = "deprecated"
+)
```
RedbackThomson, line 28: Are these in line with the statuses used by AWS or K8s? I tried to do a bit of googling but didn't find anything useful. I did see `stable`, `deprecated` and `preview`.

pkg/model/multiversion/api_info.go
```diff
@@ -0,0 +1,38 @@
...
+
+// APIInfo contains information related a specific apiVersion.
+type APIInfo struct {
+	// The API status. Can be one of Available, Removed and Deprecated.
+	Status APIStatus
+	// the aws-sdk-go version used to generated the apiVersion.
+	AWSSDKVersion string
+	// Full path of the generator config file.
+	GeneratorConfigPath string
```
RedbackThomson, line 37: Store APIVersion in this as well?

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,248 @@
...
+	FieldChangeTypeAdded FieldChangeType = "added"
+	// FieldChangeTypeRemoved is used a when a field is removed from a CRD.
+	FieldChangeTypeRemoved FieldChangeType = "removed"
+	// FieldChangeTypeRenamed is used when a field is renamed.
+	FieldChangeTypeRenamed FieldChangeType = "renamed"
+	// FieldChangeTypeShapeChanged is used when a field shape has changed.
+	FieldChangeTypeShapeChanged FieldChangeType = "shape-changed"
+	// FieldChangeTypeShapeChangedToSecret is used when a field change to
+	// a k8s secret type.
+	FieldChangeTypeShapeChangedToSecret FieldChangeType = "shape-changed-to-secret"
```
RedbackThomson, line 45: We need discrete types for secret?

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,248 @@
...
+			ChangeType: FieldChangeTypeAdded,
+		})
+	}
+
+	return deltas, nil
+}
+
+// isEqualShape returns whether two awssdkmodel.ShapeRef are equal or not.
+// TODO(a-hilaly): this is very fragile - a simple docstring change will make the
+// result wrong - we'll need to slowly verify each member/key/value
```
RedbackThomson, line 235: Can we pass this through our existing generator methods to convert it to Go types before comparing?

pkg/model/multiversion/mapper.go
```diff
@@ -0,0 +1,194 @@
...
+	ackmodel "github.com/aws-controllers-k8s/code-generator/pkg/model"
+	"github.com/aws-controllers-k8s/code-generator/pkg/util"
+)
+
+var (
+	ErrAPIVersionNotFound   = errors.New("api version not found")
+	ErrAPIVersionDeprecated = errors.New("api version deprecated")
+)
+
+// ModelsMapper is a multi-version model mapper. It is containing the mapping
```
RedbackThomson, line 33: nit:
```suggestion
// ModelsMapper is a multi-version model mapper. It contains the mapping
```

pkg/model/multiversion/mapper.go
```diff
@@ -0,0 +1,194 @@
...
+// of each non-deprecated version with their correspending ackmodel.model
+// and APIInfos.
+type ModelsMapper struct {
+	gitRepo *git.Repository
+
+	hubVersion    string
+	spokeVersions []string
+
+	deprecatedVersions []string
+	removedVersions    []string
```
RedbackThomson, line 43: I think we need a repository metadata configuration file. It should contain general metadata about the service - full name, short name, link to docs, etc. - but also a list of each version and their support state (deprecated, stable, under development). This would serve as the basis for the generated documentation as well.

RedbackThomson (commented):
Amazing work! Your implementation of the field diffs feels very well thought out and I walked through it without any confusion. I have a suggestion about introducing a metadata configuration file that might help solve the deprecation status (and other stuff) we have. This can be a separate issue that I'd be happy to tackle.

pkg/model/multiversion/api_info.go
```diff
@@ -0,0 +1,38 @@
...
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package multiversion
+
+// TODO(a-hilaly) move this file outside of pkg/generate/multiversion. Ideally we
```
vijtrip2, line 16: nit: pkg/model/multiversion?

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,248 @@
...
+
+func max(a, b int) int {
+	if a > b {
+		return a
+	}
+	return b
+}
+
+// ComputeFieldsDiff computes the difference between two maps of fields. It returns a list
+// of FieldDelta's that contains the the ChangeType and at least one field reference.
```
vijtrip2, line 98: nit: "the the" typo

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,248 @@
...
+func ComputeFieldsDiff(
+	spokeCRDFields map[string]*ackmodel.Field,
+	hubCRDFields map[string]*ackmodel.Field,
+	// the hub renames
+	renames map[string]string,
+) ([]FieldDelta, error) {
+	// the smallest size of deltas we can get is max(len(spokeCRDFields), len(hubCRDFields))
```
vijtrip2, line 105: do you mean maximum size of delta, instead of smallest?

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,248 @@
...
+		// If a field is found in both arrays only three changes are possible:
+		// Intact, TypeChangeChange and ChangeTypeShapeChangedToSecret.
+		// NOTE(a-hilaly): carefull about X -> Y then Z -> X renames. It should
+		// not be allowed.
+		if ok {
+			// mark field as visited.
+			visitedFields[spokeFieldName] = nil
+			// check if field became secret
+			if (spokeField.FieldConfig == nil ||
+				(spokeField.FieldConfig != nil && !spokeField.FieldConfig.IsSecret)) &&
```
vijtrip2, line 139: Do we care about change in other direction for this SecretChangeType? IOW, should there be another if condition where "spokeField" is a secret but "hubField" is not.

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,248 @@
...
+				})
+				continue
+			}
+
+			// at this point we know that the fields kept the same name but have different
+			// shapes
+			deltas = append(deltas, FieldDelta{
+				Spoke:      spokeField,
+				Hub:        hubField,
+				ChangeType: FieldChangeTypeIntact,
```
vijtrip2, line 169: copy paste error? This should be "FieldChangeTypeShapeChanged". correct?

pkg/model/multiversion/mapper.go
```diff
@@ -0,0 +1,194 @@
...
+
+var (
+	ErrAPIVersionNotFound   = errors.New("api version not found")
+	ErrAPIVersionDeprecated = errors.New("api version deprecated")
+)
+
+// ModelsMapper is a multi-version model mapper. It is containing the mapping
+// of each non-deprecated version with their correspending ackmodel.model
+// and APIInfos.
+type ModelsMapper struct {
```
jaypipes, line 36: recommend calling this `APIVersionMapper` or `APIVersionManager`. You are mapping API Versions to models.

pkg/model/multiversion/mapper.go
```diff
@@ -0,0 +1,194 @@
...
+		modelsMapping: modelsMapping,
+	}
+	// TODO(hilalymh): Audit deprecated and removed versions
+	return model, nil
+}
+
+// GetModel returns the model of a given api version.
+func (m *ModelsMapper) GetModel(apiVersion string) (*ackmodel.Model, error) {
+	if err := m.VerifyAPIVersions(apiVersion); err != nil {
+		return nil, fmt.Errorf("cannot verify apiVersions %s: %v", apiVersion, err)
```
jaypipes, line 115: s/verify apiVersions/verify API version/

pkg/model/multiversion/mapper.go
```diff
@@ -0,0 +1,194 @@
...
+	return m.spokeVersions
+}
+
+// GetHubVersion returns the hub version.
+func (m *ModelsMapper) GetHubVersion() string {
+	return m.hubVersion
+}
+
+// CompareHubWith compares a given api version with the hub version and returns
+// slices of FieldDeltas representing the diff between CRDs status and spec fields.
```
jaypipes, line 131: docstring says the function returns slices of FieldDeltas, but it seems to return a map of string to CRDDelta.

pkg/model/multiversion/mapper.go
```diff
@@ -0,0 +1,194 @@
...
+	spokeVersions []string
+
+	deprecatedVersions []string
+	removedVersions    []string
+
+	apiInfos      map[string]APIInfo
+	modelsMapping map[string]*ackmodel.Model
+}
+
+// NewModelsMapper returns a new Model struct.
```
jaypipes, line 49: does it? :)

pkg/model/crd.go
```diff
@@ -618,6 +618,41 @@ func (r *CRD) ListOpMatchFieldNames() []string {
...
 }
 
+// GetAllRenames returns all the fields renames observed in the generator.yaml.
+// It will return two rename maps, the first is mapping the old names to the new
+// ones, the other is mapping the new names to the old ones.
+func (r *CRD) GetAllRenames() (
+	map[string]string,
+	map[string]string,
+	error,
+) {
```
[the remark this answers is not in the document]
jaypipes: Actually, I would just return a single `map[string]string` representing the old name to new name.

Looking at this PR, the only place where the map of new to old name is used is in a test case, and @A-Hilaly you can easily just reverse the map of old to new in a test utility method.

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,248 @@
...
+// ComputeFieldsDiff computes the difference between two maps of fields. It returns a list
+// of FieldDelta's that contains the the ChangeType and at least one field reference.
+func ComputeFieldsDiff(
+	spokeCRDFields map[string]*ackmodel.Field,
+	hubCRDFields map[string]*ackmodel.Field,
+	// the hub renames
+	renames map[string]string,
+) ([]FieldDelta, error) {
+	// the smallest size of deltas we can get is max(len(spokeCRDFields), len(hubCRDFields))
+	deltas := make([]FieldDelta, 0, max(len(spokeCRDFields), len(hubCRDFields)))
```
jaypipes, line 106: yeah, recommend just doing:

```go
deltas := []FieldDetla{}
```

and get rid of the max stuff.

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,248 @@
...
+
+	hubFieldsNames := []string{}
+	for name := range hubCRDFields {
+		hubFieldsNames = append(hubFieldsNames, name)
+	}
+	sort.Strings(hubFieldsNames)
+
+	// let's make sure we don't visit fields more than once - especially
+	// when fields are renamed.
+	visitedFields := map[string]*struct{}{}
```
jaypipes, line 123: typically I see `map[string]bool` when implementing sets in Go...

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,248 @@
...
+)
+
+// FieldChangeType represent the type of field modification.
+type FieldChangeType string
+
+const (
+	// FieldChangeTypeUnknown is used when ChangeType cannot be computed.
+	FieldChangeTypeUnknown FieldChangeType = "unknown"
+	// FieldChangeTypeIntact is used a field name and structure didn't change.
+	FieldChangeTypeIntact FieldChangeType = "intact"
```
jaypipes, line 34: Is this a field change then? Maybe call this `FieldChangeTypeNone`?

pkg/model/multiversion/delta.go
jaypipes, line 169: shouldn't this be `FieldChangeTypeShapeChanged`?

jaypipes (changes requested):

pkg/model/multiversion/api_info.go
A-Hilaly: These are k8s api versions statuses (picked from deprecation table https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-parts-of-the-api)

A-Hilaly: I'm thinking that maybe we can split `available` to `stable` and `preview`. WDYT @RedbackThomson ?

pkg/model/multiversion/delta.go
A-Hilaly: I believe that we need to handle `string` to `secret` changes differently than other type changes

pkg/model/multiversion/delta.go
A-Hilaly: @RedbackThomson  @jaypipes `cleanGoType` will always return the same type name even if the shape changes. I think that we need to minutely compare the shape members and their types.

pkg/model/multiversion/delta.go
A-Hilaly: I mean the minimum: if `[A, B]` changes to `[A, B, C]` the delta is `[A->A, B->B, ->C]`

A-Hilaly: I'm removed this size since it's not needed at all

pkg/model/multiversion/delta.go
A-Hilaly: Good point @vijtrip2 - it's very unlikely to happen but i should add a case for that.

pkg/model/multiversion/delta.go
A-Hilaly: Yup nice catch!

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,248 @@
...
+		return nil, fmt.Errorf("cannot compute status fields deltas: %s", err)
+	}
+
+	return &CRDDelta{
+		specDeltas,
+		statusDeltas,
+	}, nil
+}
+
+func max(a, b int) int {
```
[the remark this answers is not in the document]
A-Hilaly: Agreed it's useless to optimize here.

pkg/model/multiversion/mapper.go
A-Hilaly: I like the `APIVersionManager` - changed

pkg/model/multiversion/delta.go
A-Hilaly: Yepp ++

pkg/model/multiversion/delta.go
A-Hilaly: Changed

pkg/model/multiversion/mapper.go
A-Hilaly: Yikes :facepalm:

pkg/model/multiversion/api_info.go
A-Hilaly: Done

pkg/model/multiversion/api_info.go
jaypipes: > I'm thinking that maybe we can split `available` to `stable` and `preview`. WDYT @RedbackThomson ?

I'm against this. Our use of the phases `preview` and `stable` have nothing to do with API versions. They are an indication of the confidence of the team that owns that controller about the long-running stability of a particular controller image.

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,257 @@
...
+	return (src.FieldConfig == nil ||
+		(src.FieldConfig != nil && !src.FieldConfig.IsSecret)) &&
+		(dst.FieldConfig != nil && dst.FieldConfig.IsSecret)
+}
+
+// ComputeFieldsDiff computes the difference between two maps of fields. It returns a list
+// of FieldDelta's that contains the ChangeType and at least one field reference.
+func ComputeFieldsDiff(
+	spokeCRDFields map[string]*ackmodel.Field,
+	hubCRDFields map[string]*ackmodel.Field,
```
jaypipes, line 104: nit: might be easier to call these variables `from` and `to`. I know that upstream controller-runtime uses the terms spoke and hub to refer to the concept of multiple source versions to one target version, but in this function, what we are interested in is a *single* source (`from`) to a single target (`to`)

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,257 @@
...
+	// the hub renames
+	renames map[string]string,
+) ([]FieldDelta, error) {
+	deltas := []FieldDelta{}
+
+	// collect field names and sort them to ensure a determenistic output order.
+	spokeFieldsNames := []string{}
```
jaypipes, line 111: and similarly, this could be just `fromNames`.

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,257 @@
...
+	for name := range spokeCRDFields {
+		spokeFieldsNames = append(spokeFieldsNames, name)
+	}
+	sort.Strings(spokeFieldsNames)
+
+	hubFieldsNames := []string{}
```
jaypipes, line 117: and this could be `toNames`

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,257 @@
...
+			if fieldChangedToSecret(hubField, spokeField) {
+				deltas = append(deltas, FieldDelta{
+					Spoke:      spokeField,
+					Hub:        hubField,
+					ChangeType: FieldChangeTypeShapeChangedFromSecretToString,
+				})
+				continue
+			}
+
+			// TODO(a-hilaly) better function for equalizing shapes
```
jaypipes, line 158: s/equalizing/comparing/

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	// At this point we collected every type of changes except added fields.
+	// To find added fields we just look for fields that are in hubFieldsNames
+	// and were not visited before (are not in spokeFieldsNames).
+	for _, hubFieldName := range hubFieldsNames {
+		_, visited := visitedFields[hubFieldName]
+		if visited {
+			continue
+		}
```
jaypipes, line 229: you could shorten this to:

```go
if _, visited := visitedFields[hubFieldName]; visited {continue}
```

pkg/model/multiversion/delta.go
jaypipes: Yeah, I see what you're saying @A-Hilaly. ++

pkg/model/multiversion/delta_test.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+}
+
+func reverseMap(t *testing.T, m map[string]string) map[string]string {
+	reverseMap := make(map[string]string, len(m))
+	for k, v := range m {
+		reverseMap[v] = k
+	}
+	return reverseMap
+}
```
jaypipes, line 243: bingo. :+1:

pkg/model/multiversion/manager.go
```diff
@@ -0,0 +1,197 @@
...
+	gitRepo *git.Repository
+
+	hubVersion    string
+	spokeVersions []string
+
+	deprecatedVersions []string
+	removedVersions    []string
+
+	apiInfos      map[string]APIInfo
+	modelsMapping map[string]*ackmodel.Model
```
jaypipes, line 47: suggest naming this field just `models`. The fact it's a `map` is an implementation detail :) Just like above, you call your field `apiInfos` not `apiInfosMapping`. ;)

pkg/model/sdk_helper.go
```diff
@@ -59,6 +62,24 @@ func NewSDKHelper(basePath string) *SDKHelper {
...
+		}
+		h.gitRepository = gitRepository
+	}
+
+	err := util.CheckoutRepositoryTag(h.gitRepository, version)
+	if err != nil {
+		return fmt.Errorf("cannot checkout tag %s: %v", version, err)
+	}
+	return nil
+}
```
jaypipes, line 81: excellent. :+1:

jaypipes (changes requested):
@A-Hilaly a few suggestions for renaming some variables and fields, otherwise this is very close to merging. Great stuff!

pkg/model/multiversion/manager.go
A-Hilaly: Yep! agreed and changed to `apiInfosMapping`

pkg/model/multiversion/delta_test.go
A-Hilaly: Removing this utility function and adding `ComputeRenamesDelta` to properly detect the renames from a version to another.

pkg/model/multiversion/delta.go
A-Hilaly: Changed to `src` and `dst` (source and destination)

pkg/model/multiversion/delta.go
A-Hilaly: changed to `srcNames`

pkg/model/multiversion/delta.go
A-Hilaly: changed to `dstNames`

pkg/model/multiversion/delta.go
A-Hilaly: Removed `TODO` and implemented a proper function to compare two shapes `isEqualShape`

pkg/model/crd.go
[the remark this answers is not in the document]
A-Hilaly: @jaypipes @RedbackThomson 
I miss interpreted this problem when i started implementing `ack-generate webhooks`. The renames map should be computed separately, by comparing two rename maps (hub and spoke renames) and finding the "evolution" of renames from a source to destination API version.
For example if the renames didn't change from v1 to v2 our renames map should be empty.

I added `ComputeRenamesDelta` to properly compute the renames map in `delta_renames.go`

pkg/model/multiversion/api_info.go
A-Hilaly: Alright - keeping `available`/`removed`/`deprecated` for now.

pkg/model/crd.go
```diff
@@ -618,6 +618,32 @@ func (r *CRD) ListOpMatchFieldNames() []string {
...
+	resourceConfig, ok := r.cfg.Resources[r.Names.Original]
+	if !ok {
+		return nil, fmt.Errorf("resource not found")
+	}
+
+	renames := make(map[string]string)
+	opMap := r.sdkAPI.GetOperationMap(r.cfg)
+	createOps := (*opMap)[OpTypeCreate]
+
+	if resourceConfig.Renames != nil && resourceConfig.Renames.Operations != nil {
```
RedbackThomson, line 632: nit: Invert condition and early return to decrease nesting

pkg/model/crd.go
```diff
@@ -618,6 +618,32 @@ func (r *CRD) ListOpMatchFieldNames() []string {
...
+		opRenameConfigs := resourceConfig.Renames.Operations
+		for opName, opRenameConfigs := range opRenameConfigs {
+			for _, op := range createOps {
+				if opName == op.Name {
```
RedbackThomson, line 636: nit: Invert condition here too, use `continue`

pkg/model/crd.go
```diff
@@ -618,6 +618,32 @@ func (r *CRD) ListOpMatchFieldNames() []string {
...
+// GetAllRenames returns all the field renames observed in the generator.yaml.
+func (r *CRD) GetAllRenames() (map[string]string, error) {
+	resourceConfig, ok := r.cfg.Resources[r.Names.Original]
+	if !ok {
+		return nil, fmt.Errorf("resource not found")
+	}
+
+	renames := make(map[string]string)
+	opMap := r.sdkAPI.GetOperationMap(r.cfg)
+	createOps := (*opMap)[OpTypeCreate]
```
RedbackThomson, line 630: Is this necessary? I understand it's future proofing, but either we should iterate through all operations or only iterate through create.

pkg/model/crd.go
```diff
@@ -618,6 +618,32 @@ func (r *CRD) ListOpMatchFieldNames() []string {
 	return r.cfg.ListOpMatchFieldNames(r.Names.Original)
 }
 
+// GetAllRenames returns all the field renames observed in the generator.yaml.
```
RedbackThomson, line 621:
```suggestion
// GetAllRenames returns all the field renames observed in any `Create*` operations in the generator config
```

pkg/model/multiversion/api_info.go
```diff
@@ -0,0 +1,40 @@
...
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package multiversion
+
+// TODO(a-hilaly) move this file outside of pkg/model/multiversion. Idealy we
```
RedbackThomson, line 16: Why is this a TODO? Can we not do it in this PR?

pkg/model/multiversion/api_info.go
```diff
@@ -0,0 +1,40 @@
...
+// Should be able to access APIStatus and APIInfo to prevent regenerating removed or
+// deprecated APIs.
+// TODO(a-hilaly) store API status in `ack-generate-metadata.yaml`
```
RedbackThomson, line 19: nit: As part of discussions regarding the metadata config file, not true anymore. `metadata.yaml` will be source of truth for API statuses

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,289 @@
...
+	// FieldChangeTypeRenamed is used when a field is renamed.
+	FieldChangeTypeRenamed FieldChangeType = "renamed"
+	// FieldChangeTypeShapeChanged is used when a field shape has changed.
+	FieldChangeTypeShapeChanged FieldChangeType = "shape-changed"
+	// FieldChangeTypeShapeChangedFromStringToSecret is used when a field change to
+	// a k8s secret type.
+	FieldChangeTypeShapeChangedFromStringToSecret FieldChangeType = "shape-changed-from-string-to-secret"
+	// FieldChangeTypeShapeChangedFromSecretToString is used when a field changed from
+	// a k8s secret to a Go string.
+	FieldChangeTypeShapeChangedFromSecretToString FieldChangeType = "shape-changed-from-secret-to-string"
```
RedbackThomson, line 46: super nit: Could you tab indent these fields so they all line up in a nice fashion? :)

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,289 @@
...
+)
+
+// FieldDelta represent the delta between the same field in two different
+// CRD versions. If a field is removed in the Destination version the Destination value will
```
RedbackThomson, line 50: Could you clarify what the `same field` means? Is this identified by the field name in the operation shape?

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,289 @@
...
+
+	return &CRDDelta{
+		specDeltas,
+		statusDeltas,
+	}, nil
+}
+
+// fieldChangedToSecret returns true if field changed from string to secret.
+func fieldChangedToSecret(src, dst *ackmodel.Field) bool {
+	return (src.FieldConfig == nil ||
```
RedbackThomson, line 103: `src.FieldConfig == nil` implies this is a new field? Not that it changed from `string` to `Secret`? Is this the intended effect?

pkg/model/multiversion/delta_renames.go
```diff
@@ -0,0 +1,70 @@
...
+
+package multiversion
+
+import (
+	"errors"
+	"fmt"
+)
+
+var (
+	// ErrProhibitedRename is returned by ComputeRenames when a prohibited renaming
```
RedbackThomson, line 22: nit: `s/ComputeRenames/ComputeRenamesDelta`

pkg/model/multiversion/delta_renames.go
```diff
@@ -0,0 +1,70 @@
...
+
+// ComputeRenamesDelta returns a map representing the field renames map between two
+// distinguished api versions.
+//
+// Examples:
+//     if we rename X to Y in v1 and X to Y in v2 the map of renames is {}
+//     if we rename X to Y in v1 and X to Z in v2 the map of renames is {Y: Z}
+//     if we don't rename any field in v1 and we rename X to Y in v2 the map of renames if {X: Y}
+func ComputeRenamesDelta(srcRenames, dstRenames map[string]string) (map[string]string, error) {
+	// returns and error if we find any prohibitted or non-supported renaming pattern.
```
RedbackThomson, line 37: nit:
```suggestion
// returns an error if we find any prohibited or unsupported renaming pattern.
```

pkg/model/multiversion/manager.go
```diff
@@ -0,0 +1,197 @@
...
+var (
+	ErrAPIVersionNotFound   = errors.New("api version not found")
+	ErrAPIVersionRemoved    = errors.New("api version removed")
+	ErrAPIVersionDeprecated = errors.New("api version deprecated")
+)
+
+// APIVersionManager is a API versions manager. It contains the mapping
+// of each non-deprecated version with their correspending ackmodel.Model
+// and APIInfos.
+type APIVersionManager struct {
```
RedbackThomson, line 37: note to @RedbackThomson: Incorporate loading `ServiceMetadata` into `APIVersionManager`

pkg/model/multiversion/manager.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot read sdk git repository: %v", err)
+	}
+
+	SDKAPIHelper := ackmodel.NewSDKHelper(sdkCacheDir)
+
+	// create model for each non-deprecated api version
+	models := make(map[string]*ackmodel.Model, len(apisInfo))
+	for apiVersion, apiInfo := range apisInfo {
+		// TODO(a-hilaly) filter deprecated and removed api versions
```
RedbackThomson, line 73: I think we will continue to support all released versions of the API? Why would we ever deprecate an API version? It costs us nothing to generate the webhooks.

pkg/model/multiversion/manager.go
```diff
@@ -0,0 +1,197 @@
...
+) (*APIVersionManager, error) {
+	if len(apisInfo) == 0 {
+		return nil, fmt.Errorf("empty apisInfo")
+	}
+
+	spokeVersions := make([]string, 0, len(apisInfo)-1)
+	gitRepo, err := util.LoadRepository(sdkCacheDir)
+	if err != nil {
+		return nil, fmt.Errorf("cannot read sdk git repository: %v", err)
+	}
```
RedbackThomson, line 66: Do we have tests that we can identify and load the repository correctly? All the delta tests use the same AWS SDK version

pkg/testutil/schema_helper.go
```diff
@@ -24,7 +24,38 @@ import (
...
 
+// TestingModelOptions contains optional variables that are passed to
+// `NewModelForServiceWithOptions`.
+type TestingModelOptions struct {
+	// The CR API Version. Defaults to v1alpha1
+	APIVersion string
+	// The generator config file. Defaults to generator.yaml
+	GeneratorConfigFile string
+	// The AWS API version. Defaults to 00-00-0000
+	SDKAPIVersion string
```
RedbackThomson, line 35: nit: This is not the `SDKAPIVersion`, it's a versioned date string related to the API backend format (or something like that). It should not be confused for the AWS SDK version.

pkg/model/crd.go
A-Hilaly: Nice suggestion. Changing in next push

pkg/model/crd.go
A-Hilaly: We should be only interested by `CreateOp` renames when using this function. But that's a good point as well - i believe we should pass an op parameter to the function

pkg/model/multiversion/api_info.go
A-Hilaly: This is very related to `pkg/metadata` that you're working on. Just keeping here and moving it to the right place when we'll have a good idea where it should be placed.

pkg/model/multiversion/api_info.go
A-Hilaly: Yup ++, removing the TODO

pkg/model/multiversion/delta.go
A-Hilaly: Agreed this doesn't look very good x( ... `gofmt` will always relints these types

A-Hilaly: Maybe i can separate the constants and their comments 🤔

pkg/testutil/schema_helper.go
A-Hilaly: Yepp changed to `ServiceAPIVersion` to avoid confusion with AWS SDK Version.

A-Hilaly: In the `api-2.json` it is called `apiVersion` https://github.com/aws/aws-sdk-go/blob/main/models/apis/kafka/2018-11-14/api-2.json#L3

pkg/model/multiversion/delta.go
A-Hilaly: >src.FieldConfig == nil implies this is a new field? 

Not really, a field can exist and have no configuration in `generator.yaml`.

>  Not that it changed from string to Secret?

This function return true if a field changed from `string` to `Secret`. To find these fields we should verify that either the `src` Field have a `is_secret: false` or don't have a configuration in the `generator.yaml` (which is equivalent to `is_secret: false`), and that `dst` have a `is_secret: true`

pkg/model/multiversion/manager.go
A-Hilaly: We should continue to generate webhooks for `Deprecated` versions but not for the removed ones.
I think that at one point we should start deprecating versions and eventually remove them. I imagine that if one day we release a `v1` api version - all alpha/beta version should be at least "Deprecated" if not removed

A-Hilaly: According to https://kubernetes.io/docs/reference/using-api/deprecation-policy/ we should not have more than one "Available" version at onces - and at most two "Deprecated" versions

pkg/model/multiversion/manager.go
A-Hilaly: > Do we have tests that we can identify and load the repository correctly?

Unfortunately no.. the testdata we have is a collection of text files and they are not part of an `aws-sdk-go` git repository.

> All the delta tests use the same AWS SDK version

I just added new delta unit tests using multiple AWS SDK versions

A-Hilaly:
Current tests total coverage: 54.8%


```bash
go test -coverprofile test.out -tags codegen -v ./pkg/model/multiversion/...
go tool cover -func test.out

github.com/aws-controllers-k8s/code-generator/pkg/model/multiversion/delta.go:71:		ComputeCRDFieldsDeltas	68.8%
github.com/aws-controllers-k8s/code-generator/pkg/model/multiversion/delta.go:104:		fieldChangedToSecret	100.0%
github.com/aws-controllers-k8s/code-generator/pkg/model/multiversion/delta.go:112:		ComputeFieldsDelta	95.7%
github.com/aws-controllers-k8s/code-generator/pkg/model/multiversion/delta.go:251:		getFieldNameFromDelta	100.0%
github.com/aws-controllers-k8s/code-generator/pkg/model/multiversion/delta.go:261:		AreEqualShapes		66.7%
github.com/aws-controllers-k8s/code-generator/pkg/model/multiversion/delta_renames.go:36:	ComputeRenamesDelta	100.0%
github.com/aws-controllers-k8s/code-generator/pkg/model/multiversion/manager.go:51:		NewAPIVersionManager	0.0%
github.com/aws-controllers-k8s/code-generator/pkg/model/multiversion/manager.go:113:		GetModel		0.0%
github.com/aws-controllers-k8s/code-generator/pkg/model/multiversion/manager.go:121:		GetSpokeVersions	0.0%
github.com/aws-controllers-k8s/code-generator/pkg/model/multiversion/manager.go:126:		GetHubVersion		0.0%
github.com/aws-controllers-k8s/code-generator/pkg/model/multiversion/manager.go:132:		CompareHubWith		0.0%
github.com/aws-controllers-k8s/code-generator/pkg/model/multiversion/manager.go:137:		VerifyAPIVersions	0.0%
github.com/aws-controllers-k8s/code-generator/pkg/model/multiversion/manager.go:155:		CompareAPIVersions	0.0%
total:												(statements)		54.8%

```

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,307 @@
...
+		(src.FieldConfig != nil && !src.FieldConfig.IsSecret)) &&
+		(dst.FieldConfig != nil && dst.FieldConfig.IsSecret)
+}
+
+// ComputeFieldsDelta computes the difference between two maps of fields. It returns a list
+// of FieldDelta's that contains the ChangeType and at least one field reference.
+func ComputeFieldsDelta(
```
jaypipes, line 113: A couple more naming changes please:

1. Please name this function `ComputeFieldDeltas` not `ComputeFieldsDelta`
2. Please change `srcCRDFields` to `srcFields` and `dstCRDFields` to `dstFields`. A Field is part of a CRD so there's no reason to repeat `CRD` in the variable name.

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,307 @@
...
+// CRDDelta stores the spec and status deltas for a custom resource.
+type CRDDelta struct {
+	SpecDeltas   []FieldDelta
+	StatusDeltas []FieldDelta
+}
+
+// ComputeCRDFieldsDeltas compares two ackmodel.CRD instances and returns the
+// spec and status fields deltas. srcCRD is the CRD of the spoke (source) version
+// and dstCRD is the CRD of the hub (destination) version.
+func ComputeCRDFieldsDeltas(srcCRD, dstCRD *ackmodel.CRD) (*CRDDelta, error) {
```
jaypipes, line 72: Can we change these parameter names to just `src` and `dst` and leave off the `CRD` part? Would make it easier to read and it's obvious from the type what the parameters are.

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,307 @@
...
+	"sort"
+
+	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
+
+	ackmodel "github.com/aws-controllers-k8s/code-generator/pkg/model"
+)
+
+// FieldChangeType represent the type of field modification.
+//
+// Currently there 8 supported FieldChangeTypes
```
jaypipes, line 27: recommend removing the above line. It will quickly get out of date (the number `8`)

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,307 @@
...
+	FieldChangeTypeNone                           FieldChangeType = "none"
+	FieldChangeTypeAdded                          FieldChangeType = "added"
+	FieldChangeTypeRemoved                        FieldChangeType = "removed"
+	FieldChangeTypeRenamed                        FieldChangeType = "renamed"
+	FieldChangeTypeShapeChanged                   FieldChangeType = "shape-changed"
+	FieldChangeTypeShapeChangedFromStringToSecret FieldChangeType = "shape-changed-from-string-to-secret"
+	FieldChangeTypeShapeChangedFromSecretToString FieldChangeType = "shape-changed-from-secret-to-string"
+)
+
+// FieldDelta represent the delta between the same original field in two
```
jaypipes, line 51: s/represent/represents/

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,307 @@
...
+import (
+	"fmt"
+	"sort"
+
+	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
+
+	ackmodel "github.com/aws-controllers-k8s/code-generator/pkg/model"
+)
+
+// FieldChangeType represent the type of field modification.
```
jaypipes, line 25: s/represent/represents/

pkg/model/multiversion/delta.go
```diff
@@ -0,0 +1,307 @@
...
+		})
+	}
+
+	sort.Slice(deltas, func(i, j int) bool {
+		return getFieldNameFromDelta(deltas[i]) < getFieldNameFromDelta(deltas[j])
+	})
+	return deltas, nil
+}
+
+// getFieldNameFromDelta retrives the field name from a FieldDelta. If a field is
```
jaypipes, line 250: s/retrives/retrieves

jaypipes (approved):
minor variable rename suggestions and typos, but the code looks good, thank you Amine!

pkg/model/multiversion/delta.go
A-Hilaly: Agreed. Changed.

pkg/model/multiversion/delta.go
A-Hilaly: Changed

jaypipes (approved):
:+1:

PR #122, opened 2021-07-06
Introduce `ack-generate webhooks` sub command

A-Hilaly:
Part of https://github.com/aws-controllers-k8s/community/issues/835

This patch brings in a new sub command that can be used to generate
webhook functions. For now it can only generate conversion webhook
functions, but it can be extended to generate defaulting and validating
functions.

This patch also introduces a new function that generates conversion
functions. It can be used to generate `ConvertTo` and `ConvertFrom`
functions.


A-Hilaly:
NOTE to reviews: This will need a rebase after #120 and #121 merge.
/hold

RedbackThomson:
Could we hold off on this command until after I propose the metadata configuration? Options such as deprecated versions, etc. will all be available in that configuration file and it will not be necessary to pass them as command line arguments.

A-Hilaly:
> Could we hold off on this command until after I propose the metadata configuration? Options such as deprecated versions, etc. will all be available in that configuration file and it will not be necessary to pass them as command line arguments.

Sure. I will rebase on top of your work once it's available.

pkg/generate/ack/output.go
```diff
@@ -145,6 +145,23 @@ func CreateGenerationMetadata(
...
+	}
+
+	var generationMetadata GenerationMetadata
+	err = yaml.Unmarshal(b, &generationMetadata)
+	if err != nil {
+		return nil, err
+	}
+	return &generationMetadata, nil
+}
+
```
A-Hilaly, line 164: @RedbackThomson If you can implement a similar function in your PR, i'm taking it :)

A-Hilaly:
Need to be rebased, now that #126 is merged
/hold

pkg/generate/code/converter/converter.go
```diff
@@ -0,0 +1,718 @@
...
+//
+// Output code will look something like this:
+//
+//   elementCopy := &v1alpha2.WebhookFilter{}
+//   if element != nil {
+//   	webhookFilterCopy := &v1alpha2.WebhookFilter{}
+//   	webhookFilterCopy.ExcludeMatchedPattern = element.ExcludeMatchedPattern
+//   	webhookFilterCopy.Pattern = element.Pattern
+//   	webhookFilterCopy.Type = element.Type
+//   	elementCopy = webhookFilterCopy
```
jaypipes, line 137: This code doesn't look correct to me... Shouldn't it be like this?

```go
elementCopy := &v1alpha2.WebhookFilter{}
if element != nil {
    elementCopy.ExcludeMatchedPattern = element.ExcludeMatchedPattern
    elementCopy.Pattern = element.Pattern
    elementCopy.Type = element.Type
}
```

pkg/generate/code/converter/converter.go
```diff
@@ -0,0 +1,718 @@
...
+		varStructCopy,
+		c.srcVar,
+		false,
+	)
+
+	varIndex := "i"
+	varElement := "element"
+	// for i, element := range src.Spec.FilterGroups {
+	out += fmt.Sprintf(
+		"%s\tfor %s, %s := range %s {\n",
```
jaypipes, line 319: Why not just use `for _, %s := range %s {\n` here and get rid of the next line.

pkg/generate/code/converter/converter.go
```diff
@@ -0,0 +1,718 @@
...
+			)
+		} else {
+			out += c.withIndentLevel(c.indentLevel+2).
+				WithVariables(varElement, varElementCopy).
+				copyMap(
+					memberShape,
+				)
+		}
+	default:
+		panic(fmt.Sprintf("Unsupported shape type in generate.code.copyMap"))
```
jaypipes, line 384: s/copyMap/copyList`

pkg/generate/code/converter/converter.go
```diff
@@ -0,0 +1,718 @@
...
+	out += fmt.Sprintf(
+		"%s\t\t_ = %s // non-used value guard.\n",
+		indent,
+		varIndex,
+	)
+
+	memberShapeRef := shape.MemberRef
+	memberShape := memberShapeRef.Shape
+	// elementCopy := make([]*v1alpha2.WebhookFilter, 0, len(element))
+	varElementCopy := "element" + "Copy"
```
jaypipes, line 336: You are going to run into problems with nested list types here, because you will have variable name shadowing in effect. This is why the code generation that sets CR fields from an Output shape (the `pkg/generate/code.SetOutput` stuff) generates a variable name that represents the field's numeric index within the Shape. For instance, if I had a Shape like this:

```go
type Bar struct {
     Baz string
}

type Foo struct {
  Bars []*Bar
}

type MyShape struct {
  Foos []*Foo
}
```

The variable referring to `MyShape.Foos.Bars` would be called `f0f0` and the iterator over elements in that list of `Bar` structs would be called `f0f0iter`, with the element variable name of `f0f0elem`.

The way you are naming things here means that you would end up producing Go code like this:

```go
elementCopy := make([]*Foo, 0, len(src.Foos))
for _, element := range src.Foos {
    fooCopy := &Foo{}
    elementCopy := make([]*Bar, 0, len(element.Bars)
    for _, element := range element.Bars {
        barCopy := &Bar{}
        barCopy.Baz = element.Baz
        elementCopy = append(elementCopy, barCopy)
    }
    elementCopy = append(elementCopy, fooCopy)
}
```

which would be invalid since the variable name `elementCopy` is shadowed/hidden by the inner declaration of itself.

pkg/generate/code/converter/converter.go
```diff
@@ -0,0 +1,718 @@
...
+	out += fmt.Sprintf(
+		"%s\tfor %s, %s := range %s {\n",
+		indent,
+		keyVarName,
+		valueVarName,
+		c.srcVar,
+	)
+
+	memberShape := shape.ValueRef.Shape
+	varElementCopy := "element" + "Copy"
```
jaypipes, line 474: same comment as above :) you are going to run into variable naming conflicts.

pkg/generate/code/converter/converter.go
```diff
@@ -0,0 +1,718 @@
...
+		)
+	default:
+		msg := fmt.Sprintf("Unsupported shape type in generate.code.newShapeTypeInstance %s", shape.Type)
+		panic(msg)
+	}
+
+	return out
+}
+
+// storeVariableIn retruns go code that stores a value in a given variable.
```
jaypipes, line 676: s/retruns go/returns Go

jaypipes (changes requested):
@A-Hilaly wonderful start on this! You will need to make a number of changes, however, in order to properly deal with nested complex types. See inline for an explanation of why your variable-naming schema needs to be adapted.

cmd/ack-generate/command/webhook.go
```diff
@@ -0,0 +1,128 @@
...
+	apisInfos := map[string]multiversion.APIInfo{}
+	for _, f := range files {
+		metadata, err := ack.LoadGenerationMetadata(apisVersionPath, f.Name())
+		if err != nil {
+			return err
+		}
+		apisInfos[f.Name()] = multiversion.APIInfo{
+			// Setting the status to unknown for now.
+			// TODO(a-hilaly) read status from metadata.yaml
+			Status:              multiversion.APIStatusUnknown,
```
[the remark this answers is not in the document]
RedbackThomson: Even better, this can come directly from the `metadata` object.

pkg/generate/ack/output.go
RedbackThomson: These have been moved to `pkg/metadata`

PR #124, opened 2021-07-06
Add a new sort option for `kubebuilder:printcolumns` marker comments

A-Hilaly:
Issue N/A

This patch adds a new option `Index` for print column configuration.
Setting an index for a field and `resource.print.sortBy` to `index` will
intruct the code generator to sort the `kubebuilder:printcolumns` by the
`Index` value.

### DIFF
--- a/pkg/model/printer_column.go
+++ b/pkg/model/printer_column.go
@@ -26,6 +26,7 @@ type PrinterColumn struct {
 	Type     string
 	Priority int
 	JSONPath string
+	Index    int
 }
 
 // By can sort two PrinterColumns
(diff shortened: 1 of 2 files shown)


pkg/model/printer_column.go
```diff
@@ -77,8 +77,13 @@ func sortFunction(sortByField string) func(a, b *PrinterColumn) bool {
 		return func(a, b *PrinterColumn) bool {
 			return a.JSONPath < b.JSONPath
 		}
+	case "index":
+		return func(a, b *PrinterColumn) bool {
+			return a.Index < b.Index
+		}
```
RedbackThomson, line 83: Should this be the first case in the switch? Sorting by index should be preferred over any other?

A-Hilaly: I don't how moving this case could change the default/preferred value. `sortByField` will have one of the values.

A-Hilaly: Did you mean moving it to the default case? or maybe somehow set a default for `$resource.print.sortBy` ?

RedbackThomson: Oh no I totally misread this method. Sorry

PR #126, opened 2021-07-13
Add service metadata configuration file

RedbackThomson:
Description of changes:
Creates a new specification for a `metadata.yaml` file that resides in the root of the controller repository. The purpose of this file is to provide human-friendly display information about the service (long name, acronym, links) and to provide a source of truth for API versions.


pkg/metadata/service_metadata.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	return availableVersions[len(availableVersions)-1], nil
+}
+
+func (m *ServiceMetadata) GetDeprecatedAPIVersions() []string {
+	return m.getVersionsByStatus(APIStatusDeprecated)
+}
+
+func (m *ServiceMetadata) GetRemoveAPIVersions() []string {
```
jaypipes, line 65: s/GetRemove/GetRemoved/

pkg/metadata/service_metadata.go
```diff
@@ -0,0 +1,118 @@
...
+type ServiceVersion struct {
+	APIVersion string    `json:"api_version"`
+	Status     APIStatus `json:"status"`
+}
+
+func (m *ServiceMetadata) GetLatestAPIVersion() (string, error) {
+	availableVersions := m.GetAvailableAPIVersions()
+
+	if len(availableVersions) == 0 {
+		return "", fmt.Errorf("service metadata contains no available versions")
```
jaypipes, line 55: suggest moving the `fmt.Errorf` to being an `errors.New` and make it a package-level exported variable.

same for the other places you use `fmt.Errorf` in this file.

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/spf13/cobra v1.1.1
 	github.com/stretchr/testify v1.7.0
 	golang.org/x/mod v0.4.1
+	golang.org/x/tools v0.0.0-20200616195046-dc31b401abb5
```
jaypipes, line 19: where'd this come from? maybe remove it with `go mod tidy`?

pkg/metadata/api_info.go
```diff
@@ -11,19 +11,16 @@
...
 
 type APIStatus string
 
 const (
-	APIStatusUnknown    APIStatus = "unknown"
-	APIStatusAvailable            = "available"
-	APIStatusRemoved              = "removed"
-	APIStatusDeprecated           = "deprecated"
+	APIStatusUnknown       APIStatus = "unknown"
+	APIStatusInDevelopment           = "in_development"
```
jaypipes, line 20: Can we call this just `development` and make it the default instead of `unknown`?

pkg/testdata/models/metadata.yaml
```diff
@@ -0,0 +1,8 @@
+service:
+  full_name: Amazon Test Service
+  short_name: ATS
+  link: https://example.com/
+  documentation: https://docs.example.com/
+versions:
+- api_version: v1alpha1
+  status: development
```
jaypipes, line 8: It might be useful to have a helper script in the code-generator that will examine a new controller repository and construct a default metadata file given a service name...

jaypipes (changes requested):

pkg/metadata/api_info.go
A-Hilaly: I'm wondering if we really need a `development` status for API versions. I think that we should increase the API Version whenever the CRD schema changes.

A-Hilaly: IMO `development` should be a status for the service controller and not a specific API version.

pkg/metadata/service_metadata.go
```diff
@@ -0,0 +1,118 @@
...
+	"io/ioutil"
+
+	"github.com/ghodss/yaml"
+)
+
+// ServiceMetadata consists of information about the service and relative links as well
+// as a list of supported/deprecated versions
+type ServiceMetadata struct {
+	Service ServiceDetails `json:"service"`
+	// CRDs to ignore. ACK generator would skip these resources.
```
A-Hilaly, line 27: Comment unrelated to `Versions`

pkg/metadata/service_metadata.go
```diff
@@ -0,0 +1,118 @@
...
+	Versions []ServiceVersion `json:"versions"`
```
A-Hilaly, line 28: Let's rename this variable `APIVersions` ? just to avoid confusion with controller image versions (`v0.1.2`... etc)

pkg/metadata/service_metadata.go
```diff
@@ -0,0 +1,118 @@
...
+	Link string `json:"link"`
+	// The URL of the service's main documentation/user guide
+	Documentation string `json:"documentation"`
+}
+
+// ServiceVersion describes the status of all existing version of the controller
+type ServiceVersion struct {
+	APIVersion string    `json:"api_version"`
+	Status     APIStatus `json:"status"`
+}
```
A-Hilaly, line 49: We could also add a boolean field called `IsHub`, to be able to override the hub version. In most of the cases the latest `Available` version is the hub  (prefered storage version). However, according to [k8s deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-parts-of-the-api) sometimes the hub can be different from the latest version.

pkg/model/model.go
```diff
@@ -40,6 +41,8 @@ type Model struct {
 	typeDefs     []*TypeDef
 	typeImports  map[string]string
 	typeRenames  map[string]string
+	// Metadata for the service
+	meta *ackmetadata.ServiceMetadata
```
A-Hilaly, line 45: Do we need `*ackmetadata.ServiceMetadata` in `ackmode.Model` and `multiversion.APIVersionManager`? what are the use cases when generating a one api version (`ack-generate apis`)?

pkg/model/multiversion/manager.go
```diff
@@ -87,25 +102,27 @@ func NewAPIVersionManager(
...
 	sort.Strings(spokeVersions)
 	model := &APIVersionManager{
 		gitRepo:       gitRepo,
+		metadata:      metadata,
 		hubVersion:    hubVersion,
 		spokeVersions: spokeVersions,
 		apiInfos:      apisInfo,
 		models:        models,
 	}
-	// TODO(hilalymh): Audit deprecated and removed versions
```
A-Hilaly, line 108: I'm still going to send a PR for API versions examination. Mainly to verify that service controller maintainers didn't violate one the deprecation/removal policies.

pkg/model/multiversion/manager.go
```diff
@@ -35,31 +36,40 @@ var (
...
-	spokeVersions := make([]string, 0, len(apisInfo)-1)
+	metadata, err := ackmetadata.NewServiceMetadata(metadataPath)
+	if err != nil {
+		return nil, err
+	}
+
+	hubVersion, err := metadata.GetLatestAPIVersion()
+	if err != nil {
+		return nil, err
+	}
```
A-Hilaly, line 69: The hub version is not always the latest version. I would like to keep the `APIVersionManager` flexible in term of choosing a hub version. However we can still enforce "choosing the latest version as hub" in the build script or `cmd` / `pkg/generate/ack`

pkg/model/multiversion/manager.go
```diff
@@ -68,11 +78,16 @@ func NewAPIVersionManager(
...
+	models := map[string]*ackmodel.Model{}
+	for _, version := range metadata.Versions {
+		if version.APIVersion == hubVersion || version.Status == ackmetadata.APIStatusDeprecated || version.Status == ackmetadata.APIStatusRemoved {
+			continue
+		}
+
+		spokeVersions = append(spokeVersions, version.APIVersion)
+		apiInfo, ok := apisInfo[version.APIVersion]
+		if !ok {
+			return nil, fmt.Errorf("could not find API info for API version %s", version.APIVersion)
```
A-Hilaly, line 90: We should verify that we have an API info for the hub version as well

pkg/metadata/service_metadata.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+	return versions
+}
+
+// NewServiceMetadata returns a new Metadata object given a supplied
+// path to a metadata file
+func NewServiceMetadata(
+	metadataPath string,
+) (ServiceMetadata, error) {
```
A-Hilaly, line 105: nit: you can return a `*ServiceMetadata` to avoid allocating a new `ServiceMetadata{}` when you return an error

pkg/metadata/service_metadata.go
RedbackThomson: I am quite averse to the idea of the hub being anything other than the latest generated version - we rely on it being the latest for the common runtime elements to match.

pkg/model/model.go
RedbackThomson: True. I had this in before your last PR, I'll remove it again.

pkg/model/multiversion/manager.go
RedbackThomson: Sounds good. Will wait for your changes to `cmd` - we can load it in there.

pkg/metadata/api_info.go
RedbackThomson: That seems fair. The API version in development will always be the latest available option. "In development" vs. "released" is controlled by our controller release already.

pkg/metadata/service_metadata.go
```diff
@@ -0,0 +1,109 @@
...
+	return m.getVersionsByStatus(APIStatusDeprecated)
+}
+
+func (m *ServiceMetadata) GetRemovedAPIVersions() []string {
+	return m.getVersionsByStatus(APIStatusRemoved)
+}
+
+func (m *ServiceMetadata) GetAvailableAPIVersions() []string {
+	return m.getVersionsByStatus(APIStatusAvailable)
+}
```
A-Hilaly, line 76: half nit:please add godoc comment, at least for the exported functions

A-Hilaly (commented):
Looks good overall! :+1:

A-Hilaly (approved):

scripts/construct-metadata.sh
```diff
@@ -0,0 +1,68 @@
...
+    echo "${USAGE}"
+    exit 1
+fi
+
+METADATA_TEMPLATE_PATH="$ROOT_DIR/templates/metadata.yaml"
+
+DEFAULT_METADATA_OUTPUT_PATH="$SERVICE_CONTROLLER_SOURCE_PATH/metadata.yaml"
+METADATA_OUTPUT_PATH="${METADATA_OUTPUT_PATH:-$DEFAULT_METADATA_OUTPUT_PATH}"
+
+echo "🧙 Welcome to the service metadata setup wizard"
```
jaypipes, line 51: sweet emojis.

jaypipes (approved):
Nice work on this @RedbackThomson :)

PR #127, opened 2021-07-15
Recursively build a clean GoType for nested maps and arrays

A-Hilaly:
Issue N/A

Description of changes:
This patch fixes a bug where `ack-generate` used to  prepend a `[]`
prefix to any map of structs (example `Route_SDK.RequestParameters`
in apigatewayv2)

Also in the same PR, moving the GoType cleanup logic to a separate 
function `model.getShapeCleanGoType` to simplify `model.GetTypeDefs`
code.


jaypipes (approved):
Very nice. And good spot on that bug in maps.

PR #130, opened 2021-07-20
Add script and Github workflow for ensuring version in sync

RedbackThomson:
Description of changes:
Adds a script that compares the `aws-controllers-k8s/runtime` version listed in the `go.mod` is identical to the `VERSION` variable in the `Makefile`. This way we can ensure the generation metadata is kept in line with the version of the runtime it was built against.

/assign @A-Hilaly 

### DIFF
--- a/scripts/check-versions.sh
+++ b/scripts/check-versions.sh
@@ -31,7 +31,7 @@ if [ $# -ne 1 ]; then
 fi
 
 RUNTIME_DEPENDENCY_VERSION="aws-controllers-k8s/runtime"
-GO_MOD_VERSION=$(cat go.mod | grep $RUNTIME_DEPENDENCY_VERSION | cut -d ' ' -f2)
+GO_MOD_VERSION=$(go list -m -f '{{ .Version }}' github.com/aws-controllers-k8s/runtime)
 
 if [[ "$1" != "$GO_MOD_VERSION" ]]; then
     echo "Code-generator version in Makefile $1 does not match $RUNTIME_DEPENDENCY_VERSION version $GO_MOD_VERSION"
--- a/scripts/check-versions.sh
+++ b/scripts/check-versions.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+
+# A script that ensures the code-generator binary version matches its associated
+# runtime version
+
+set -eo pipefail
+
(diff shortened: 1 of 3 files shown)


scripts/check-versions.sh
```diff
@@ -0,0 +1,41 @@
...
+"
+
+if [ $# -ne 1 ]; then
+    echo "ERROR: $(basename "$0") only accepts a single parameter" 1>&2
+    echo "$USAGE"
+    exit 1
+fi
+
+RUNTIME_DEPENDENCY_VERSION="aws-controllers-k8s/runtime"
+GO_MOD_VERSION=$(cat go.mod | grep $RUNTIME_DEPENDENCY_VERSION | cut -d ' ' -f2)
```
A-Hilaly, line 34: nit: you can also use `go list` to get the version `ack/runtime` module. Some thing similar to:
```bash
go list -m -f '{{ .Version }}' github.com/aws-controllers-k8s/runtime
```

A-Hilaly (approved):
Nicee! With this CI job we'll never forget to update the version in the Makefile :) ! Left one small suggestion

scripts/check-versions.sh
RedbackThomson: Oh super cool! Thanks!

RedbackThomson:
I was hoping we could include this into something more akin to a Git hook - so that changes would not even make it into a branch. However, Git hooks are not included in the remote repository and I didn't want to require developers to manually enable them (no-one would, probably). This was my next best option.

scripts/check-versions.sh
```diff
@@ -0,0 +1,41 @@
...
+"
+
+if [ $# -ne 1 ]; then
+    echo "ERROR: $(basename "$0") only accepts a single parameter" 1>&2
+    echo "$USAGE"
+    exit 1
+fi
+
+RUNTIME_DEPENDENCY_VERSION="aws-controllers-k8s/runtime"
+GO_MOD_VERSION=$(go list -m -f '{{ .Version }}' github.com/aws-controllers-k8s/runtime)
```
A-Hilaly, line 34: `GO_MOD_PATH` is not used, you can add another flag to specify the `go.mod` file: `-modfile $GO_MOD_PATH`

PR #131, opened 2021-07-21
Secret errors to be terminal code

nmvk:
For now secret errors would be terminal code to
be consistent with ACK behavior.

```
  Conditions:
    Message:  kubernetes secret not found
    Status:   True
    Type:     ACK.Terminal
```

Runtime PR - https://github.com/aws-controllers-k8s/runtime/pull/33


A-Hilaly (approved):

jaypipes (changes requested):
We will want to cut a new release of ACK runtime and then you should update the go.mod file and test in pkg/generate/ack/runtime_test.go to pin to that latest runtime release.

PR #135, opened 2021-07-22
Update required controller-gen to `v0.6.1`

RedbackThomson:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/873

Description of changes:
- Removes duplication of the `CONTROLLER_TOOLS_VERSION` variable
- Deletes the defunct `k8s.sh` script (not being used anywhere in this repo)

I have tested the new `controller-gen` against the `dynamodb` controller and saw no changes in output CRDs (except for the version number in an annotation). None of the breaking changes called out in the release between `0.4.0` and `0.6.1` should have any effect on our generator.


scripts/lib/common.sh
```diff
@@ -88,3 +90,22 @@ debug_msg() {
...
     echo "$__debug_prefix$__indent$__msg"
 }
+
+# controller_gen_version_equals accepts a string version and returns 0 if the
+# installed version of controller-gen matches the supplied version, otherwise
+# returns 1
+#
+# Usage:
+#
+#   if controller_gen_version_equals "v0.4.0"; then
```
vijtrip2, line 100: nit: /s/controller_gen_version_equals/k8s_controller_gen_version_equals

scripts/install-controller-gen.sh
```diff
@@ -19,12 +19,10 @@ set -eo pipefail
 
 SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 ROOT_DIR="$SCRIPTS_DIR/.."
-CONTROLLER_TOOLS_VERSION="v0.4.0"
 
 source "$SCRIPTS_DIR/lib/common.sh"
-source "$SCRIPTS_DIR/lib/k8s.sh"
 
-if ! is_installed controller-gen; then
+if ! is_installed controller-gen || ! k8s_controller_gen_version_equals "$CONTROLLER_TOOLS_VERSION"; then
```
vijtrip2, line 25: Could there be possible collision between which controller-gen binary will be picked up if there is an already installed old version?
IOW, should other version be uninstalled? or better, should the script always install controller-gen in a specific path(which it does) and use complete path to invoke binary to avoid name collision in $PATH ?

vijtrip2 (approved):

scripts/install-controller-gen.sh
A-Hilaly: Good point @vijtrip2 , older versions should be uninstalled.

RedbackThomson: Many of our ACK contributors also use `controller-gen` (of different versions) for other projects. I don't want to dictate what version they have on their systems, if they've previously installed something else. From what I've seen, this script overrides the default version of `controller-gen` by default - but I do think it's a good idea to place it in a logical spot (say in `~/.cache`) and re-use that path, rather than putting it into `$PATH`

RedbackThomson: I just looked into this and I don't really want to expand the scope of this particular PR into something like that. I think there are lots of places for errors to creep in. This PR is meant to just increment the version number and clean up some scripts.

PR #138, opened 2021-07-26
Initial implementation of AWSResourceManager.LateInitialize() method.

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/812

Description of changes:

* Initial implementation of `AWSResourceManager.LateInitialize()` method code generator
* Adds two new hooks `late_initialize_pre_read_one` & `late_initialize_post_read_one`
* Introduces new LateInitializeConfig with DelaySeconds as single parameter
* Corresponding runtime PR : https://github.com/aws-controllers-k8s/runtime/pull/37

------------------------------------

Testing
* Unit Tests Successful
```
➜  code-generator git:(li-mlp) ✗ make test
go test -tags codegen ./...
?       github.com/aws-controllers-k8s/code-generator/cmd/ack-generate  [no test files]
?       github.com/aws-controllers-k8s/code-generator/cmd/ack-generate/command  [no test files]
ok      github.com/aws-controllers-k8s/code-generator/pkg/generate/ack  2.346s
ok      github.com/aws-controllers-k8s/code-generator/pkg/generate/code 17.798s
?       github.com/aws-controllers-k8s/code-generator/pkg/generate/config       [no test files]
?       github.com/aws-controllers-k8s/code-generator/pkg/generate/crossplane   [no test files]
?       github.com/aws-controllers-k8s/code-generator/pkg/generate/olm  [no test files]
?       github.com/aws-controllers-k8s/code-generator/pkg/generate/templateset  [no test files]
?       github.com/aws-controllers-k8s/code-generator/pkg/metadata      [no test files]
ok      github.com/aws-controllers-k8s/code-generator/pkg/model 14.671s
ok      github.com/aws-controllers-k8s/code-generator/pkg/model/multiversion    5.152s
ok      github.com/aws-controllers-k8s/code-generator/pkg/names (cached)
?       github.com/aws-controllers-k8s/code-generator/pkg/testutil      [no test files]
?       github.com/aws-controllers-k8s/code-generator/pkg/util  [no test files]
?       github.com/aws-controllers-k8s/code-generator/pkg/version       [no test files]

```

* Controller generation successful
```
➜  code-generator git:(li-mlp) ✗ make build-controller                
building ack-generate ... ok.
installing controller-gen v0.6.1 ... ok.
Copying common custom resource definitions into ecr
Building Kubernetes API objects for ecr
Generating deepcopy code for ecr
Generating custom resource definitions for ecr
Building service controller for ecr
Generating RBAC manifests for ecr
Running gofmt against generated code for ecr

```

* Sample generated code for ecr repository
```go
// LateInitialize returns an acktypes.AWSResource after setting the late initialized
// fields from the readOne call. This method will initialize the optional fields
// which were not provided by the k8s user but were defaulted by the AWS service.
// If there are no such fields to be initialized, the returned object is similar to
// object passed in the parameter.
func (rm *resourceManager) LateInitialize(
	ctx context.Context,
	res acktypes.AWSResource,
) (acktypes.AWSResource, error) {
	rlog := ackrtlog.FromContext(ctx)
	// If there are no fields to late initialize, do nothing
	if len(lateInitializeFieldNames) == 0 {
		rlog.Debug("no late initialization required.")
		return res, nil
	}
	r := rm.concreteResource(res)
	if r.ko == nil {
		// Should never happen... if it does, it's buggy code.
		panic("resource manager's LateInitialize() method received resource with nil CR object")
	}
	koWithDefaults := r.ko.DeepCopy()
	latestWithDefaults := &resource{koWithDefaults}
	lateInitConditionReason := ""
	lateInitConditionMessage := ""

	observed, err := rm.sdkFind(ctx, latestWithDefaults)
	if err != nil {
		lateInitConditionMessage = "Unable to complete Read operation required for late initialization"
		lateInitConditionReason = "Late Initialization Failure"
		ackcondition.SetLateInitialized(latestWithDefaults, corev1.ConditionFalse, &lateInitConditionMessage, &lateInitConditionReason)
		return latestWithDefaults, ackrequeue.NeededAfter(err, time.Duration(0)*time.Second)
	}
	observedKo := observed.ko
	if observedKo.Spec.Name != nil && koWithDefaults.Spec.Name == nil {
		koWithDefaults.Spec.Name = observedKo.Spec.Name
	}

	incompleteInitialization := rm.incompleteLateInitialization(latestWithDefaults)
	if incompleteInitialization {
		// Add the condition with LateInitialized=False
		lateInitConditionMessage = "Late initialization did not complete, requeuing with delay of 5 seconds"
		lateInitConditionReason = "Delayed Late Initialization"
		ackcondition.SetLateInitialized(latestWithDefaults, corev1.ConditionFalse, &lateInitConditionMessage, &lateInitConditionReason)
		return latestWithDefaults, ackrequeue.NeededAfter(nil, time.Duration(5)*time.Second)
	}
	// Set LateIntialized condition to True
	lateInitConditionMessage = "Late initialization successful"
	lateInitConditionReason = "Late initialization successful"
	ackcondition.SetLateInitialized(latestWithDefaults, corev1.ConditionTrue, &lateInitConditionMessage, &lateInitConditionReason)
	return latestWithDefaults, nil
}

// incompleteLateInitialization return true if there are fields which were supposed to be
// late initialized but are not. If all the fields are late initialized, false is returned
func (rm *resourceManager) incompleteLateInitialization(
	latestWithDefaults *resource,
) bool {
	ko := latestWithDefaults.ko
	if ko.Spec.Name == nil {
		return true
	}
	return false

}

```

* local-kind-e2e-test logs
```
                                                                                                    │
│ 2021-07-26T19:38:08.701Z    DEBUG    ackrt    > r.Sync    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "kind": "Repository" │
│ 2021-07-26T19:38:08.701Z    DEBUG    ackrt    >> rm.ReadOne    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_adopted": f │
│ 2021-07-26T19:38:08.701Z    DEBUG    ackrt    >>> rm.sdkFind    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_adopted":  │
│ 2021-07-26T19:38:09.641Z    DEBUG    ackrt    <<< rm.sdkFind    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_adopted":  │
│ 2021-07-26T19:38:09.641Z    DEBUG    ackrt    << rm.ReadOne    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_adopted": f │
│ 2021-07-26T19:38:09.641Z    DEBUG    ackrt    >> r.setResourceManaged    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_a │
│ 2021-07-26T19:38:09.641Z    DEBUG    ackrt    >>> kc.Patch (metadata + spec)    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2" │
│ 2021-07-26T19:38:09.694Z    DEBUG    ackrt    <<< kc.Patch (metadata + spec)    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2" │
│ 2021-07-26T19:38:09.694Z    DEBUG    ackrt    marked resource as managed    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "i │
│ 2021-07-26T19:38:09.694Z    DEBUG    ackrt    << r.setResourceManaged    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_a │
│ 2021-07-26T19:38:09.694Z    DEBUG    ackrt    >> rm.Create    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_adopted": fa │
│ 2021-07-26T19:38:09.694Z    DEBUG    ackrt    >>> rm.sdkCreate    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_adopted" │
│ 2021-07-26T19:38:09.725Z    DEBUG    ackrt    <<< rm.sdkCreate    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_adopted" │
│ 2021-07-26T19:38:09.725Z    DEBUG    ackrt    << rm.Create    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_adopted": fa │
│ 2021-07-26T19:38:09.725Z    INFO    ackrt    created new resource    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_adopt │
│ 2021-07-26T19:38:09.725Z    DEBUG    ackrt    >> rm.LateInitialize    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_adop │
│ 2021-07-26T19:38:09.725Z    INFO    ackrt    calculated late initialization delay is 0 seconds    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "re │
│ 2021-07-26T19:38:09.725Z    DEBUG    ackrt    >>> rm.sdkFind    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_adopted":  │
│ 2021-07-26T19:38:09.748Z    DEBUG    ackrt    <<< rm.sdkFind    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_adopted":  │
│ 2021-07-26T19:38:09.748Z    DEBUG    ackrt    << rm.LateInitialize    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_adop │
│ 2021-07-26T19:38:09.748Z    DEBUG    ackrt    >> r.patchResourceMetadataAndSpec    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west │
│ 2021-07-26T19:38:09.748Z    DEBUG    ackrt    >>> kc.Patch (metadata + spec)    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2" │
│ 2021-07-26T19:38:09.758Z    DEBUG    ackrt    <<< kc.Patch (metadata + spec)    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2" │
│ 2021-07-26T19:38:09.759Z    DEBUG    ackrt    patched resource metadata and spec    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-wes │
│ 2021-07-26T19:38:09.759Z    DEBUG    ackrt    << r.patchResourceMetadataAndSpec    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west │
│ 2021-07-26T19:38:09.759Z    DEBUG    ackrt    >> r.patchResourceStatus    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_ │
│ 2021-07-26T19:38:09.759Z    DEBUG    ackrt    >>> kc.Patch (status)    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_ado │
│ 2021-07-26T19:38:09.794Z    DEBUG    ackrt    <<< kc.Patch (status)    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_ado │
│ 2021-07-26T19:38:09.795Z    DEBUG    ackrt    patched resource status    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_a │
│ 2021-07-26T19:38:09.795Z    DEBUG    ackrt    << r.patchResourceStatus    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_ │
│ 2021-07-26T19:38:09.795Z    DEBUG    ackrt    < r.Sync    {"kind": "Repository", "namespace": "default", "name": "ecr-repository-wy6n6wjm8", "generation": 1, "account": "309117047740", "role": "", "region": "us-west-2", "is_adopted": false, │
│ 2021-07-26T19:38:09.795Z    DEBUG    controller-runtime.controller    Successfully Reconciled    {"controller": "repository", "request": "default/ecr-repository-wy6n6wjm8"}
```

### DIFF
--- a/pkg/generate/code/late_initialize.go
+++ b/pkg/generate/code/late_initialize.go
@@ -27,7 +27,7 @@ import (
 func FindLateInitializedFieldsWithDelay(
 	cfg *ackgenconfig.Config,
 	r *model.CRD,
-// Number of levels of indentation to use
+	// Number of levels of indentation to use
 	indentLevel int,
 ) string {
 	//Sample output
@@ -37,18 +37,18 @@ func FindLateInitializedFieldsWithDelay(
 	fieldNameToConfig := cfg.ResourceFields(r.Names.Original)
 	if len(fieldNameToConfig) > 0 {
 		fieldNameToDelaySeconds := make(map[string]int)
-		sortedFieldNames := make([]string, 0)
+		sortedLateInitFieldNames := make([]string, 0)
 		for fName, fConfig := range fieldNameToConfig {
 			if fConfig != nil && fConfig.LateInitialize != nil {
 				fieldNameToDelaySeconds[fName] = fConfig.LateInitialize.DelaySeconds
-				sortedFieldNames = append(sortedFieldNames, fName)
+				sortedLateInitFieldNames = append(sortedLateInitFieldNames, fName)
 			}
 		}
-		sort.Strings(sortedFieldNames)
+		sort.Strings(sortedLateInitFieldNames)
 		lateInitFieldToDelayValues := ""
-		if len(sortedFieldNames) > 0 {
-			for _, fName := range sortedFieldNames {
-				lateInitFieldToDelayValues += fmt.Sprintf("\"%s\":%d,",fName, fieldNameToDelaySeconds[fName])
+		if len(sortedLateInitFieldNames) > 0 {
+			for _, fName := range sortedLateInitFieldNames {
+				lateInitFieldToDelayValues += fmt.Sprintf("%q:%d,", fName, fieldNameToDelaySeconds[fName])
 			}
 			out += fmt.Sprintf("%svar lateInitializeFieldToDelaySeconds = map[string]int{%s}\n", indent, lateInitFieldToDelayValues)
 		}
@@ -56,8 +56,8 @@ func FindLateInitializedFieldsWithDelay(
 	return out
 }
 
-// lateInitializedFieldNames returns the field names which have LateInitialization configuration inside generator config
-func lateInitializedFieldNames(
+// findLateInitializedFieldNames returns the field names which have LateInitialization configuration inside generator config
+func findLateInitializedFieldNames(
 	cfg *ackgenconfig.Config,
 	r *model.CRD,
 ) []string {
@@ -77,63 +77,65 @@ func lateInitializedFieldNames(
 // Field path separated by '.' indicates members in a struct
 // Field path separated by '..' indicates member/key in a map
 // Note: Unlike Map, updating individual element of a list is not supported. LateInitializing complete list is supported.
-func LateInitializeFromReadOne (
+//
+// Sample generator config:
+// fields:
+//      Name:
+//        late_initialize: {}
+//      ImageScanningConfiguration.ScanOnPush:
+//        late_initialize:
+//          delay_seconds: 5
+//      map..subfield.x:
+//        late_initialize:
+//          delay_seconds: 5
+//      another.map..lastfield:
+//        late_initialize:
+//          delay_seconds: 5
+//      some.list:
+//        late_initialize:
+//          delay_seconds: 10
+//
+// Sample output:
+//if observed.Spec.ImageScanningConfiguration != nil && koWithDefaults.Spec.ImageScanningConfiguration != nil {
+//	if observed.Spec.ImageScanningConfiguration.ScanOnPush != nil && koWithDefaults.Spec.ImageScanningConfiguration.ScanOnPush == nil {
+//		koWithDefaults.Spec.ImageScanningConfiguration.ScanOnPush = observed.Spec.ImageScanningConfiguration.ScanOnPush
+//	}
+//}
+//if observed.Spec.Name != nil && koWithDefaults.Spec.Name == nil {
+//	koWithDefaults.Spec.Name = observed.Spec.Name
+//}
+//if observed.Spec.another != nil && koWithDefaults.Spec.another != nil {
+//	if observed.Spec.another.map != nil && koWithDefaults.Spec.another.map != nil {
+//		if observed.Spec.another.map[lastfield] != nil && koWithDefaults.Spec.another.map[lastfield] == nil {
+//		koWithDefaults.Spec.another.map[lastfield] = observed.Spec.another.map[lastfield]
+//	}
+//	}
+//}
+//if observed.Spec.map != nil && koWithDefaults.Spec.map != nil {
+//	if observed.Spec.map[subfield] != nil && koWithDefaults.Spec.map[subfield] != nil {
+//	if observed.Spec.map[subfield].x != nil && koWithDefaults.Spec.map[subfield].x == nil {
+//	koWithDefaults.Spec.map[subfield].x = observed.Spec.map[subfield].x
+//}
+//}
+//}
+//if observed.Spec.some != nil && koWithDefaults.Spec.some != nil {
+//	if observed.Spec.some.list != nil && koWithDefaults.Spec.some.list == nil {
+//		koWithDefaults.Spec.some.list = observed.Spec.some.list
+//	}
+//}
+func LateInitializeFromReadOne(
 	cfg *ackgenconfig.Config,
 	r *model.CRD,
 	sourceKoVarName string,
 	targetKoVarName string,
-// Number of levels of indentation to use
+	// Number of levels of indentation to use
 	indentLevel int,
 ) string {
-	// Sample generator config:
-	// fields:
-	//      Name:
-	//        late_initialize: {}
-	//      ImageScanningConfiguration.ScanOnPush:
-	//        late_initialize:
-	//          delay_seconds: 5
-	//      map..subfield.x:
-	//        late_initialize:
-	//          delay_seconds: 5
-	//      another.map..lastfield:
-	//        late_initialize:
-	//          delay_seconds: 5
-	//      some.list:
-	//        late_initialize:
-	//          delay_seconds: 10
-	//
-	// Sample output:
-	//if observed.Spec.ImageScanningConfiguration != nil && koWithDefaults.Spec.ImageScanningConfiguration != nil {
-	//	if observed.Spec.ImageScanningConfiguration.ScanOnPush != nil && koWithDefaults.Spec.ImageScanningConfiguration.ScanOnPush == nil {
-	//		koWithDefaults.Spec.ImageScanningConfiguration.ScanOnPush = observed.Spec.ImageScanningConfiguration.ScanOnPush
-	//	}
-	//}
-	//if observed.Spec.Name != nil && koWithDefaults.Spec.Name == nil {
-	//	koWithDefaults.Spec.Name = observed.Spec.Name
-	//}
-	//if observed.Spec.another != nil && koWithDefaults.Spec.another != nil {
-	//	if observed.Spec.another.map != nil && koWithDefaults.Spec.another.map != nil {
-	//		if observed.Spec.another.map[lastfield] != nil && koWithDefaults.Spec.another.map[lastfield] == nil {
-	//		koWithDefaults.Spec.another.map[lastfield] = observed.Spec.another.map[lastfield]
-	//	}
-	//	}
-	//}
-	//if observed.Spec.map != nil && koWithDefaults.Spec.map != nil {
-	//	if observed.Spec.map[subfield] != nil && koWithDefaults.Spec.map[subfield] != nil {
-	//	if observed.Spec.map[subfield].x != nil && koWithDefaults.Spec.map[subfield].x == nil {
-	//	koWithDefaults.Spec.map[subfield].x = observed.Spec.map[subfield].x
-	//}
-	//}
-	//}
-	//if observed.Spec.some != nil && koWithDefaults.Spec.some != nil {
-	//	if observed.Spec.some.list != nil && koWithDefaults.Spec.some.list == nil {
-	//		koWithDefaults.Spec.some.list = observed.Spec.some.list
-	//	}
-	//}
 	out := ""
-	lateInitializedFieldNames := lateInitializedFieldNames(cfg, r)
+	lateInitializedFieldNames := findLateInitializedFieldNames(cfg, r)
 	// sorting helps produce consistent output for unit test reliability
 	sort.Strings(lateInitializedFieldNames)
+	// TODO(vijat@): Add validation for correct field path in lateInitializedFieldNames
 	for _, fName := range lateInitializedFieldNames {
 		// split the field name by period
 		// each substring represents a field. No support for '..' currently
@@ -146,7 +148,7 @@ func LateInitializeFromReadOne (
 		mapShapedParent := false
 		// for every part except last, perform the nil check
 		// entries in both source and target koVarName should not be nil
-		for i,fNamePart := range fNameParts {
+		for i, fNamePart := range fNameParts {
 			if fNamePart == "" {
 				mapShapedParent = true
 				continue
--- a/pkg/generate/code/late_initialize.go
+++ b/pkg/generate/code/late_initialize.go
@@ -74,7 +74,9 @@ func lateInitializedFieldNames(
 }
 
 // LateInitializeFromReadOne returns the gocode to set LateInitialization fields from the ReadOne output
-// TODO(vijat@): add support for Map and list. Currently only structs are supported.
+// Field path separated by '.' indicates members in a struct
+// Field path separated by '..' indicates member/key in a map
+// Note: Unlike Map, updating individual element of a list is not supported. LateInitializing complete list is supported.
 func LateInitializeFromReadOne (
 	cfg *ackgenconfig.Config,
 	r *model.CRD,
@@ -83,7 +85,24 @@ func LateInitializeFromReadOne (
 // Number of levels of indentation to use
 	indentLevel int,
 ) string {
-	//Sample output
+	// Sample generator config:
+	// fields:
+	//      Name:
+	//        late_initialize: {}
+	//      ImageScanningConfiguration.ScanOnPush:
+	//        late_initialize:
+	//          delay_seconds: 5
+	//      map..subfield.x:
+	//        late_initialize:
+	//          delay_seconds: 5
+	//      another.map..lastfield:
+	//        late_initialize:
+	//          delay_seconds: 5
+	//      some.list:
+	//        late_initialize:
+	//          delay_seconds: 10
+	//
+	// Sample output:
 	//if observed.Spec.ImageScanningConfiguration != nil && koWithDefaults.Spec.ImageScanningConfiguration != nil {
 	//	if observed.Spec.ImageScanningConfiguration.ScanOnPush != nil && koWithDefaults.Spec.ImageScanningConfiguration.ScanOnPush == nil {
 	//		koWithDefaults.Spec.ImageScanningConfiguration.ScanOnPush = observed.Spec.ImageScanningConfiguration.ScanOnPush
@@ -91,6 +110,25 @@ func LateInitializeFromReadOne (
 	//}
 	//if observed.Spec.Name != nil && koWithDefaults.Spec.Name == nil {
 	//	koWithDefaults.Spec.Name = observed.Spec.Name
+	//}
+	//if observed.Spec.another != nil && koWithDefaults.Spec.another != nil {
+	//	if observed.Spec.another.map != nil && koWithDefaults.Spec.another.map != nil {
+	//		if observed.Spec.another.map[lastfield] != nil && koWithDefaults.Spec.another.map[lastfield] == nil {
+	//		koWithDefaults.Spec.another.map[lastfield] = observed.Spec.another.map[lastfield]
+	//	}
+	//	}
+	//}
+	//if observed.Spec.map != nil && koWithDefaults.Spec.map != nil {
+	//	if observed.Spec.map[subfield] != nil && koWithDefaults.Spec.map[subfield] != nil {
+	//	if observed.Spec.map[subfield].x != nil && koWithDefaults.Spec.map[subfield].x == nil {
+	//	koWithDefaults.Spec.map[subfield].x = observed.Spec.map[subfield].x
+	//}
+	//}
+	//}
+	//if observed.Spec.some != nil && koWithDefaults.Spec.some != nil {
+	//	if observed.Spec.some.list != nil && koWithDefaults.Spec.some.list == nil {
+	//		koWithDefaults.Spec.some.list = observed.Spec.some.list
+	//	}
 	//}
 	out := ""
 	lateInitializedFieldNames := lateInitializedFieldNames(cfg, r)
@@ -105,26 +143,40 @@ func LateInitializeFromReadOne (
 		fNameIndentLevel := indentLevel
 		// fParentPath keeps track of parent path for any fNamePart
 		fParentPath := ""
+		mapShapedParent := false
 		// for every part except last, perform the nil check
 		// entries in both source and target koVarName should not be nil
 		for i,fNamePart := range fNameParts {
+			if fNamePart == "" {
+				mapShapedParent = true
+				continue
+			}
 			indent := strings.Repeat("\t", fNameIndentLevel)
-			// ignore last part
+			fNamePartAccesor := fmt.Sprintf("Spec%s.%s", fParentPath, fNamePart)
+			if mapShapedParent {
+				fNamePartAccesor = fmt.Sprintf("Spec%s[%s]", fParentPath, fNamePart)
+			}
+			// Handling for all parts except last one
 			if i != len(fNameParts)-1 {
-				out += fmt.Sprintf("%sif %s.Spec%s.%s != nil && %s.Spec%s.%s != nil {\n", indent, sourceKoVarName, fParentPath, fNamePart, targetKoVarName, fParentPath, fNamePart)
+				out += fmt.Sprintf("%sif %s.%s != nil && %s.%s != nil {\n", indent, sourceKoVarName, fNamePartAccesor, targetKoVarName, fNamePartAccesor)
 				// update fParentPath and fNameIndentLevel for next iteration
-				fParentPath = fmt.Sprintf("%s.%s", fParentPath, fNamePart)
+				if mapShapedParent {
+					fParentPath = fmt.Sprintf("%s[%s]", fParentPath, fNamePart)
+					mapShapedParent = false
+				} else {
+					fParentPath = fmt.Sprintf("%s.%s", fParentPath, fNamePart)
+				}
 				fNameIndentLevel = fNameIndentLevel + 1
+			} else {
+				// handle last part here
+				// for last part, set the lateInitialized field if user did not specify field value and readOne has server side defaulted value.
+				// i.e. field is not nil in sourceKoVarName but is nil in targetkoVarName
+				out += fmt.Sprintf("%sif %s.%s != nil && %s.%s == nil {\n", indent, sourceKoVarName, fNamePartAccesor, targetKoVarName, fNamePartAccesor)
+				fNameIndentLevel = fNameIndentLevel + 1
+				indent = strings.Repeat("\t", fNameIndentLevel)
+				out += fmt.Sprintf("%s%s.%s = %s.%s\n", indent, targetKoVarName, fNamePartAccesor, sourceKoVarName, fNamePartAccesor)
 			}
 		}
-		// for last part, set the lateInitialized field if user did not specify field value and readOne has server side defaulted value.
-		// i.e. field is not nil in sourceKoVarName but is nil in targetkoVarName
-		indent := strings.Repeat("\t", fNameIndentLevel)
-		lastfNamePart := fNameParts[len(fNameParts)-1]
-		out += fmt.Sprintf("%sif %s.Spec%s.%s != nil && %s.Spec%s.%s == nil {\n", indent, sourceKoVarName, fParentPath, lastfNamePart,targetKoVarName, fParentPath, lastfNamePart)
-		fNameIndentLevel = fNameIndentLevel + 1
-		indent = strings.Repeat("\t", fNameIndentLevel)
-		out += fmt.Sprintf("%s%s.Spec%s.%s = %s.Spec%s.%s\n", indent, targetKoVarName, fParentPath,lastfNamePart, sourceKoVarName, fParentPath, lastfNamePart)
 		// Close all if blocks with proper indentation
 		fNameIndentLevel = fNameIndentLevel - 1
 		for fNameIndentLevel >= indentLevel {
--- a/pkg/generate/code/late_initialize.go
+++ b/pkg/generate/code/late_initialize.go
@@ -85,7 +85,7 @@ func LateInitializeFromReadOne (
 ) string {
 	//Sample output
 	//if observed.Spec.ImageScanningConfiguration != nil && koWithDefaults.Spec.ImageScanningConfiguration != nil {
-	//	if observed.Spec.ScanOnPush != nil && koWithDefaults.Spec.ScanOnPush == nil {
+	//	if observed.Spec.ImageScanningConfiguration.ScanOnPush != nil && koWithDefaults.Spec.ImageScanningConfiguration.ScanOnPush == nil {
 	//		koWithDefaults.Spec.ImageScanningConfiguration.ScanOnPush = observed.Spec.ImageScanningConfiguration.ScanOnPush
 	//	}
 	//}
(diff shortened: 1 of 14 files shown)


vijtrip2:
IMO, this PR should be good to review, merge now. 
The only TODO left is implementing retry-backoff strategy for late initialization which I will start working on as part of https://github.com/aws-controllers-k8s/community/issues/879 and send in a separate PR.

pkg/generate/code/late_initialize.go
```diff
@@ -0,0 +1,188 @@
...
+	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/generate/config"
+	"github.com/aws-controllers-k8s/code-generator/pkg/model"
+)
+
+// FindLateInitializedFieldsWithDelay outputs the code to create a map of fieldName to
+// late intialization delay in seconds.
+func FindLateInitializedFieldsWithDelay(
+	cfg *ackgenconfig.Config,
+	r *model.CRD,
+// Number of levels of indentation to use
```
A-Hilaly, line 30:
```suggestion
    // Number of levels of indentation to use
```

pkg/generate/code/late_initialize.go
```diff
@@ -0,0 +1,188 @@
...
+// Note: Unlike Map, updating individual element of a list is not supported. LateInitializing complete list is supported.
+func LateInitializeFromReadOne (
+	cfg *ackgenconfig.Config,
+	r *model.CRD,
+	sourceKoVarName string,
+	targetKoVarName string,
+// Number of levels of indentation to use
+	indentLevel int,
+) string {
+	// Sample generator config:
```
A-Hilaly, line 88: I think it would be better to have this comment at the function docstring comment instead

pkg/generate/code/late_initialize.go
```diff
@@ -0,0 +1,188 @@
...
+	//}
+	//}
+	//}
+	//if observed.Spec.some != nil && koWithDefaults.Spec.some != nil {
+	//	if observed.Spec.some.list != nil && koWithDefaults.Spec.some.list == nil {
+	//		koWithDefaults.Spec.some.list = observed.Spec.some.list
+	//	}
+	//}
+	out := ""
+	lateInitializedFieldNames := lateInitializedFieldNames(cfg, r)
```
A-Hilaly, line 134: not a big deal but this block is overriding the function name `lateInitializedFieldNames` in the function scope. Which makes it impossible to re-call.

pkg/generate/code/late_initialize.go
```diff
@@ -0,0 +1,188 @@
...
+				out += fmt.Sprintf("%s%s.%s = %s.%s\n", indent, targetKoVarName, fNamePartAccesor, sourceKoVarName, fNamePartAccesor)
+			}
+		}
+		// Close all if blocks with proper indentation
+		fNameIndentLevel = fNameIndentLevel - 1
+		for fNameIndentLevel >= indentLevel {
+			out += fmt.Sprintf("%s}\n", strings.Repeat("\t", fNameIndentLevel))
+			fNameIndentLevel = fNameIndentLevel - 1
+		}
+	}
```
A-Hilaly, line 186: Feowww nice!

pkg/generate/code/late_initialize.go
```diff
@@ -0,0 +1,188 @@
...
+		fNameParts := strings.Split(fName, ".")
+		// fNameIndentLevel tracks the indentation level for every new line added
+		// This variable is incremented when building nested if blocks and decremented when closing those if blocks.
+		fNameIndentLevel := indentLevel
+		// fParentPath keeps track of parent path for any fNamePart
+		fParentPath := ""
+		mapShapedParent := false
+		// for every part except last, perform the nil check
+		// entries in both source and target koVarName should not be nil
+		for i,fNamePart := range fNameParts {
```
A-Hilaly, line 149: `gofmt` will slightly change this line

pkg/generate/code/late_intialize_test.go
```diff
@@ -0,0 +1,136 @@
...
+		koWithDefaults.Spec.ImageTagMutability = observed.Spec.ImageTagMutability
+	}
+	if observed.Spec.Name != nil && koWithDefaults.Spec.Name == nil {
+		koWithDefaults.Spec.Name = observed.Spec.Name
+	}
+`
+	assert.Equal(expected, code.LateInitializeFromReadOne(crd.Config(), crd, "observed", "koWithDefaults", 1))
+}
+
+func Test_LateInitializeFromReadOne_NestedPath(t *testing.T) {
```
A-Hilaly, line 93: Thinking that maybe we can add more unit tests for a "nested struct in maps in structs"

pkg/generate/code/late_initialize.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+	}
+	return fieldNames
+}
+
+// LateInitializeFromReadOne returns the gocode to set LateInitialization fields from the ReadOne output
+// Field path separated by '.' indicates members in a struct
+// Field path separated by '..' indicates member/key in a map
+// Note: Unlike Map, updating individual element of a list is not supported. LateInitializing complete list is supported.
+func LateInitializeFromReadOne (
```
A-Hilaly, line 80: Thinking we can use this function https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/model/crd.go#L444-L477 or something similar to verify that we are not passing a wrong field path or a to slice/non-struct type (should probably panic)

pkg/generate/code/late_intialize_test.go
```diff
@@ -0,0 +1,136 @@
+package code_test
```
A-Hilaly, line 1: nit: missing license header

A-Hilaly (commented):
Niice! Great start on this @vijtrip2!
I left few comments below

pkg/generate/code/late_initialize.go
```diff
@@ -0,0 +1,188 @@
...
+			if fConfig != nil && fConfig.LateInitialize != nil {
+				fieldNameToDelaySeconds[fName] = fConfig.LateInitialize.DelaySeconds
+				sortedFieldNames = append(sortedFieldNames, fName)
+			}
+		}
+		sort.Strings(sortedFieldNames)
+		lateInitFieldToDelayValues := ""
+		if len(sortedFieldNames) > 0 {
+			for _, fName := range sortedFieldNames {
+				lateInitFieldToDelayValues += fmt.Sprintf("\"%s\":%d,",fName, fieldNameToDelaySeconds[fName])
```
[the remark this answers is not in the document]
vijtrip2: TIL.

pkg/generate/code/late_initialize.go
```diff
@@ -0,0 +1,188 @@
...
+		sortedFieldNames := make([]string, 0)
+		for fName, fConfig := range fieldNameToConfig {
+			if fConfig != nil && fConfig.LateInitialize != nil {
+				fieldNameToDelaySeconds[fName] = fConfig.LateInitialize.DelaySeconds
+				sortedFieldNames = append(sortedFieldNames, fName)
+			}
+		}
+		sort.Strings(sortedFieldNames)
+		lateInitFieldToDelayValues := ""
+		if len(sortedFieldNames) > 0 {
```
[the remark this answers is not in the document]
vijtrip2: We need that check. Renamed it to `sortedLateInitFieldNames`

pkg/generate/code/late_initialize.go
vijtrip2: good point. renamed the function.

pkg/generate/code/late_initialize.go
vijtrip2: it did. : )

pkg/generate/code/late_initialize.go
vijtrip2: But we do support Map here, so that function will need some modification. I agree to adding more static validation, adding a TODO here. which I will address along with retry-backoff task. 

Also currently the generate code will fail to compile for service team if their path is not correct, so that's one safety mechanism currently.

pkg/generate/code/late_initialize.go
```diff
@@ -188,3 +182,146 @@ func LateInitializeFromReadOne(
...
+				fNameIndentLevel = fNameIndentLevel + 1
+			} else {
+				// handle last part here
+				// for last part, if the late initialized field is still nil, calculate the retry backoff using
+				// acktypes.LateInitializationRetryConfig abstraction and set the incompleteInitialization flag to true
+				out += fmt.Sprintf("%sif %s.%s == nil {\n", indent, koVarName, fNamePartAccesor)
+				fNameIndentLevel = fNameIndentLevel + 1
+				indent = strings.Repeat("\t", fNameIndentLevel)
+				minBackoffSeconds := fieldNameToLateInitConfig[fName].MinBackoffSeconds
+				maxBackoffSeconds := fieldNameToLateInitConfig[fName].MaxBackoffSeconds
```
RedbackThomson, line 313: These are optional fields? Want to use a getter that provides a sane default?

templates/pkg/resource/manager.go.tpl
```diff
@@ -190,36 +191,36 @@ func (rm *resourceManager) LateInitialize(
...
+		lateInitConditionMessage = fmt.Sprintf("Late initialization did not complete, requeuing with delay of %d seconds", requeueDelay)
+		lateInitConditionReason = "Delayed Late Initialization"
+		ackcondition.SetLateInitialized(latestWithDefaults, corev1.ConditionFalse, &lateInitConditionMessage, &lateInitConditionReason)
+		// increment 'services.k8s.aws/late-initialization-attempt' annotation
+		ackannotation.IncrementNumLateInitializationAttempt(koWithDefaults)
+		return latestWithDefaults, ackrequeue.NeededAfter(nil, time.Duration(requeueDelay)*time.Second)
+	}
 	// Set LateIntialized condition to True
 	lateInitConditionMessage = "Late initialition successful"
     lateInitConditionReason = "Late initialition successful"
```
RedbackThomson: Oops?

pkg/generate/code/late_initialize.go
vijtrip2: Yes, these are optional. I think 0 which is current default is the sane value already.
 if MinBackoffSeconds is 0, Exponential backoff will be 0, 1, 2, 4, 6, 8, 10 seconds..... which is fine.
And if MaxBackoffSeconds is 0, it is treated as user does not want a MaxBackoff limit. 

I can think of having a some sane upper bound for MaxBackoff but I do not think it is necessary. LMK your thoughts.

RedbackThomson: Actually, I'm confused as to how we are supposed to support MaxBackoff? At the moment we are taking the max of each of the backoffs to determine the next delay, but that guarantees the backoff will be greater than or equal to every other field's backoff.
For instance, if we have a maxbackoff of 3, 4 and 5 - after some iterations we will reach the backoff of 5. This will exceed the maxbackoff of the 3 and 4. 

I would be fine with continuing the exponential backoff indefinitely, or hardcoding a limit (say 30s) that applies for every late initialized field.

vijtrip2: [Here](https://github.com/aws-controllers-k8s/runtime/pull/37/files/54699a88f8e14c2f7daace83b98490a9741904be#diff-f411262cb2fadbaec50818511b49921468a6fc23eb6a2bd9f3801294149b2102) is the implementation for calculating the backoff. 

Currently, I won't add a hardcoded maximum but if the need arises for it, We can introduce it. It is not a one way door.

pkg/generate/code/late_initialize.go
```diff
@@ -0,0 +1,327 @@
...
+//          delay_seconds: 5
+//      map..subfield.x:
+//        late_initialize:
+//          delay_seconds: 5
+//      another.map..lastfield:
+//        late_initialize:
+//          delay_seconds: 5
+//      some.list:
+//        late_initialize:
+//          delay_seconds: 10
```
RedbackThomson, line 92: Needs an update :)

templates/pkg/resource/manager.go.tpl
```diff
@@ -156,6 +164,69 @@ func (rm *resourceManager) ARNFromName(name string) string {
...
+		lateInitConditionMessage = "Unable to complete Read operation required for late initialization"
+		lateInitConditionReason = "Late Initialization Failure"
+		ackcondition.SetLateInitialized(latestWithDefaults, corev1.ConditionFalse, &lateInitConditionMessage, &lateInitConditionReason)
+		return latestWithDefaults, ackrequeue.NeededAfter(err, time.Duration(0)*time.Second)
+	}
+	observedKo := observed.ko
+{{ GoCodeLateInitializeFromReadOne .CRD "observedKo" "koWithDefaults" 1 }}
+	requeueDelay := 0
+	incompleteInitialization := false
+{{ GoCodeCalculateRequeueDelay .CRD "koWithDefaults" "numLateInitializationAttempt" "requeueDelay" "incompleteInitialization" 1 }}
```
RedbackThomson, line 207: Would you mind making this into a method? Possibly returning the delay and a boolean. It's a bit hard to find this logic when it's nested deep inside this method.

RedbackThomson (approved):

pkg/generate/code/late_initialize.go
```diff
@@ -0,0 +1,321 @@
...
+	cfg *ackgenconfig.Config,
+	r *model.CRD,
+	resVarName string,
+	// Number of levels of indentation to use
+	indentLevel int,
+) string {
+	out := ""
+	indent := strings.Repeat("\t", indentLevel)
+	sortedFieldNames, _ := getSortedLateInitFieldsAndConfig(cfg, r)
+	if len(sortedFieldNames) > 0 {
```
A-Hilaly, line 41: QQ: would it make sense if we output a `var %s = []string{}` if there are no field names?

templates/pkg/resource/manager.go.tpl
```diff
@@ -156,6 +162,69 @@ func (rm *resourceManager) ARNFromName(name string) string {
...
+	if incompleteInitialization {
+		// Add the condition with LateInitialized=False
+		lateInitConditionMessage = "Late initialization did not complete, requeuing with delay of 5 seconds"
+		lateInitConditionReason = "Delayed Late Initialization"
+		ackcondition.SetLateInitialized(latestWithDefaults, corev1.ConditionFalse, &lateInitConditionMessage, &lateInitConditionReason)
+		return latestWithDefaults, ackrequeue.NeededAfter(nil, time.Duration(5)*time.Second)
+	}
+	// Set LateIntialized condition to True
+	lateInitConditionMessage = "Late initialization successful"
+    lateInitConditionReason = "Late initialization successful"
```
A-Hilaly, line 212: nit: s/spaces/tabs/

A-Hilaly (approved):
Nice, LGTM! :+1:

pkg/generate/code/late_initialize.go
vijtrip2: Yes. Done.

templates/pkg/resource/manager.go.tpl
```diff
@@ -156,6 +162,69 @@ func (rm *resourceManager) ARNFromName(name string) string {
...
+	if r.ko == nil {
+		// Should never happen... if it does, it's buggy code.
+		panic("resource manager's LateInitialize() method received resource with nil CR object")
+	}
+	koWithDefaults := r.ko.DeepCopy()
+	latestWithDefaults := &resource{koWithDefaults}
+	lateInitConditionReason := ""
+	lateInitConditionMessage := ""
+
+	observed, err := rm.sdkFind(ctx, latestWithDefaults)
```
jaypipes, line 193: Instead of calling `sdkFind` with a `*resource`, you can just call `ReadOne` and pass `res`, like so:

```go
var latest acktypes.AWSResource
if latest, err := rm.ReadOne(ctx, res); err != nil {
    	lateInitConditionMessage = "Unable to complete Read operation required for late initialization"
	lateInitConditionReason = "Late Initialization Failure"
	ackcondition.SetLateInitialized(
		latest,
		corev1.ConditionFalse,
		&lateInitConditionMessage,
		&lateInitConditionReason,
	)
	return latest, ackrequeue.NeededAfter(err, time.Duration(0)*time.Second)
}
```

templates/pkg/resource/manager.go.tpl
```diff
@@ -156,6 +162,69 @@ func (rm *resourceManager) ARNFromName(name string) string {
...
+	observedKo := observed.ko
+{{ GoCodeLateInitializeFromReadOne .CRD "observedKo" "koWithDefaults" 1 }}
+	incompleteInitialization := rm.incompleteLateInitialization(latestWithDefaults)
+	if incompleteInitialization {
+		// Add the condition with LateInitialized=False
+		lateInitConditionMessage = "Late initialization did not complete, requeuing with delay of 5 seconds"
+		lateInitConditionReason = "Delayed Late Initialization"
+		ackcondition.SetLateInitialized(latestWithDefaults, corev1.ConditionFalse, &lateInitConditionMessage, &lateInitConditionReason)
+		return latestWithDefaults, ackrequeue.NeededAfter(nil, time.Duration(5)*time.Second)
+	}
```
jaypipes: I'm struggling to understand why we need to construct another resource in `GoCodeLateInitializeFromReadOne` at all, since the `latest` returned from `rm.ReadOne` would have all the Spec and Status fields defaulted properly already.

We could then simply compare the original (`res`) with the `latest` resource using `newDelta()` and that would give us the fields that had been server-side defaulted.

Something like this:

```go
{{- if $hookCode := Hook .CRD "late_initialize_post_read_one" }}
{{ $hookCode }}
{{- end }}
	delta := newResourceDelta(res, latest)
	if delta.DifferentAt("Spec") {
	    msg := fmt.Sprintf(
	        "Fields were late-initialized (server-side defaults): %s",
	        delta.Differences,
	    )
	    ackcondition.SetLateInitialized(latest, corev1.ConditionTrue, &msg, &lateInitConditionReason)
	}
	return latest, nil
}
```

jaypipes (changes requested):
I think this implementation could be significantly simpler... with virtually no code-generating changes needed at all.

templates/pkg/resource/manager.go.tpl
vijtrip2: Thanks. I will make the change.

templates/pkg/resource/manager.go.tpl
vijtrip2: > I'm struggling to understand why we need to construct another resource in GoCodeLateInitializeFromReadOne at all, since the latest returned from rm.ReadOne would have all the Spec and Status fields defaulted properly already.

It is because the latest from ReadOne will not only have fields that are late initialized, It will have fields that were overriden by the server as well. Hence in the method `LateInitializeFromReadOne` , we only update the fields which were not specified by the user and added by the service.

> We could then simply compare the original (res) with the latest resource using newDelta() and that would give us the fields that had been server-side defaulted.

Excellent suggestion but I see some shortcomings in current delta code. It does not have complete support for nested Maps("..") i.e. when value of Map is struct. It may be a bigger effort to add support in delta but it is very easy to just do comparison check inside `rm.incompleteLateInitialization` even for nested fields.
I can add a TODO here to reuse Delta function when feasible in future.
Also, this simplified code does not handle case of partial late initialization where for example only one of two late initialized fields is initialized and other will be present in future readOne output.

Once we have agreement over this, I will make the code changes incorporation both of your comments.

jaypipes:
@vijtrip2 OK, I just cut the v0.10.0 release of ACK runtime with the related LateInitialize changes. Please update this PR to bring in a new ACK runtime go.mod dependency, update the `pkg/generate/ack/runtime_test.go` as needed and update the Makefile version.

vijtrip2:
> @vijtrip2 OK, I just cut the v0.10.0 release of ACK runtime with the related LateInitialize changes. Please update this PR to bring in a new ACK runtime go.mod dependency, update the `pkg/generate/ack/runtime_test.go` as needed and update the Makefile version.

@jaypipes , Done. 
Also ran `go mod tidy`
`ack/runtime_test.go` was already updated with `rm.LateInitialize()` method, so no change needed there. `make test` is successful.

templates/pkg/resource/manager.go.tpl
```diff
@@ -156,6 +162,69 @@ func (rm *resourceManager) ARNFromName(name string) string {
...
+	lateInitConditionMessage := ""
+{{- if $hookCode := Hook .CRD "late_initialize_pre_read_one" }}
+{{ $hookCode }}
+{{- end }}
+	observed, err := rm.ReadOne(ctx, latest)
+	if err != nil {
+		lateInitConditionMessage = "Unable to complete Read operation required for late initialization"
+		lateInitConditionReason = "Late Initialization Failure"
+		ackcondition.SetLateInitialized(latest, corev1.ConditionFalse, &lateInitConditionMessage, &lateInitConditionReason)
+		return latest, ackrequeue.NeededAfter(err, time.Duration(0)*time.Second)
```
jaypipes, line 190: This should probably be `return latest, err`. The `handleReconcileError` in the common runtime will handle requeues if necessary and by triggering an immediate requeue here if the ReadOne call errored is just going to cause an immediate call to ReadOne again (in the requeue'd reconcile event) which would likely just return the exact same error.

templates/pkg/resource/manager.go.tpl
```diff
@@ -156,6 +162,69 @@ func (rm *resourceManager) ARNFromName(name string) string {
...
+{{ $hookCode }}
+{{- end }}
+	latest = rm.lateInitializeFromReadOneOutput(observed, latest)
+	incompleteInitialization := rm.incompleteLateInitialization(latest)
+	if incompleteInitialization {
+		// Add the condition with LateInitialized=False
+		lateInitConditionMessage = "Late initialization did not complete, requeuing with delay of 5 seconds"
+		lateInitConditionReason = "Delayed Late Initialization"
+		ackcondition.SetLateInitialized(latest, corev1.ConditionFalse, &lateInitConditionMessage, &lateInitConditionReason)
+		return latest, ackrequeue.NeededAfter(nil, time.Duration(5)*time.Second)
```
jaypipes, line 202: In a future PR, can we place the hard-coded number `5` (here and in the condition message above) as a variable in the ACK runtime `pkg/conditions` package please?

pkg/testdata/models/apis/ecr/0000-00-00/generator-with-late-initialize.yaml
```diff
@@ -0,0 +1,15 @@
+resources:
+  Repository:
+    fields:
+      Name:
+        late_initialize: {}
+      ImageTagMutability:
+        late_initialize:
+          min_backoff_seconds: 5
```
jaypipes, line 8: I'm still concerned about the maintenance costs we're imposing on service teams that own the controller repos here. They will now have to manually specify pretty much every single optional field as at least `late_initialize: {}`. I understand the concerns about defaulting behavior of server-side override values, though. Maybe there's some compromise we can make in a future enhancement to the code-generator where we have a "default late_initialize value" that can be specified in a generator config file for a particular resource... more thoughts on this would be good (in a future PR, not saying we need to change this PR)

jaypipes (changes requested):
@vijtrip2 this is very close. I have one requested change to make in this PR (see inline, it's about returning just `latest, err` instead of a requeue needed when we error in the ReadOne call...)

templates/pkg/resource/manager.go.tpl
vijtrip2: In the future PR, this static delay will be replaced by what service teams configure in their late initialization configuration. Currently this static delay was introduced because of splitting the PR.

pkg/testdata/models/apis/ecr/0000-00-00/generator-with-late-initialize.yaml
vijtrip2: Fair enough.

vijtrip2:
> @vijtrip2 this is very close. I have one requested change to make in this PR (see inline, it's about returning just `latest, err` instead of a requeue needed when we error in the ReadOne call...)

Done.

```
➜  code-generator git:(li-mlp) ✗ make test
go test -tags codegen ./...
?       github.com/aws-controllers-k8s/code-generator/cmd/ack-generate  [no test files]
?       github.com/aws-controllers-k8s/code-generator/cmd/ack-generate/command  [no test files]
ok      github.com/aws-controllers-k8s/code-generator/pkg/generate/ack  (cached)
ok      github.com/aws-controllers-k8s/code-generator/pkg/generate/code (cached)
?       github.com/aws-controllers-k8s/code-generator/pkg/generate/config       [no test files]
?       github.com/aws-controllers-k8s/code-generator/pkg/generate/crossplane   [no test files]
?       github.com/aws-controllers-k8s/code-generator/pkg/generate/olm  [no test files]
?       github.com/aws-controllers-k8s/code-generator/pkg/generate/templateset  [no test files]
?       github.com/aws-controllers-k8s/code-generator/pkg/metadata      [no test files]
ok      github.com/aws-controllers-k8s/code-generator/pkg/model (cached)
ok      github.com/aws-controllers-k8s/code-generator/pkg/model/multiversion    (cached)
ok      github.com/aws-controllers-k8s/code-generator/pkg/names (cached)
?       github.com/aws-controllers-k8s/code-generator/pkg/testutil      [no test files]
?       github.com/aws-controllers-k8s/code-generator/pkg/util  [no test files]
?       github.com/aws-controllers-k8s/code-generator/pkg/version       [no test files]
➜  code-generator git:(li-mlp) ✗ export SERVICE=ecr
➜  code-generator git:(li-mlp) ✗ make build-controller
building ack-generate ... ok.
Copying common custom resource definitions into ecr
Building Kubernetes API objects for ecr
Generating deepcopy code for ecr
Generating custom resource definitions for ecr
Building service controller for ecr
Generating RBAC manifests for ecr
Running gofmt against generated code for ecr
➜  code-generator git:(li-mlp) ✗ cd ../ecr-controller 
➜  ecr-controller git:(ack-main) ✗ make test
go test -v ./...
?       github.com/aws-controllers-k8s/ecr-controller/apis/v1alpha1     [no test files]
?       github.com/aws-controllers-k8s/ecr-controller/cmd/controller    [no test files]
?       github.com/aws-controllers-k8s/ecr-controller/pkg/resource      [no test files]
?       github.com/aws-controllers-k8s/ecr-controller/pkg/resource/repository   [no test files]
?       github.com/aws-controllers-k8s/ecr-controller/pkg/version       [no test files]
➜  ecr-controller git:(ack-main) ✗ 

```

jaypipes (approved):
OK, I'm OK with this functionally, thank you @vijtrip2 :)

In a future PR, please wrap code and code comments at 80 chars to follow the existing style as much as possible (80 chars makes demo and screen-sharing of the code easier since it allows the code to fit in a single screen without horizontal scrolling, FWIW).

PR #139, opened 2021-07-27
Allow member fields from other operations in spec

RedbackThomson:
Description of changes:
Allow the use of member fields contained within other API operations within a resource's spec. For example, `LoggingStatus` for S3 is only in the `PutBucketLogging` operation, but should be settable from spec.

### DIFF
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -143,6 +143,11 @@ func (r *CRD) HasShapeAsMember(toFind string) bool {
 			}
 		}
 	}
+	for _, field := range r.SpecFields {
+		if shapeHasMember(field.ShapeRef.Shape, toFind) {
+			return true
+		}
+	}
 	return false
 }
 
@@ -231,6 +236,17 @@ func (r *CRD) SpecFieldNames() []string {
 	res := make([]string, 0, len(r.SpecFields))
 	for fieldName := range r.SpecFields {
 		res = append(res, fieldName)
+	}
+	sort.Strings(res)
+	return res
+}
+
+// SpecFieldShapeNames returns a sorted slice of shape names for each of the
+// Spec fields
+func (r *CRD) SpecFieldShapeNames() []string {
+	res := make([]string, 0, len(r.SpecFields))
+	for _, field := range r.SpecFields {
+		res = append(res, field.ShapeRef.ShapeName)
 	}
 	sort.Strings(res)
 	return res
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -241,17 +241,6 @@ func (r *CRD) SpecFieldNames() []string {
 	return res
 }
 
-// SpecFieldShapeNames returns a sorted slice of shape names for each of the
-// Spec fields
-func (r *CRD) SpecFieldShapeNames() []string {
-	res := make([]string, 0, len(r.SpecFields))
-	for _, field := range r.SpecFields {
-		res = append(res, field.ShapeRef.ShapeName)
-	}
-	sort.Strings(res)
-	return res
-}
-
 // UnpacksAttributesMap returns true if the underlying API has
 // Get{Resource}Attributes/Set{Resource}Attributes API calls that map real,
 // schema'd fields to a raw `map[string]*string` for this resource (see SNS and
--- a/pkg/model/model_s3_test.go
+++ b/pkg/model/model_s3_test.go
@@ -75,6 +75,7 @@ func TestS3_Bucket(t *testing.T) {
 		"GrantReadACP",
 		"GrantWrite",
 		"GrantWriteACP",
+		"Logging",
 		// NOTE(jaypipes): Original field name in CreateBucket input is
 		// "Bucket" but should be renamed to "Name" from the generator.yaml (in
 		// order to match with the name of the field in the Output shape for a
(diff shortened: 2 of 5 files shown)


vijtrip2:
Is this a PR in progress? 
From the code changes, I could not relate how these change will allow mentioning spec fields from another operation.  Can you please elaborate for me? 
Also No unit tests added?

RedbackThomson:
> Is this a PR in progress?
Nope. This is just a small PR.

> From the code changes, I could not relate how these change will allow mentioning spec fields from another operation. Can you please elaborate for me?
I want to add `LoggingStatus` as a `spec` field for S3. To do this, I would add a new field in the configuration and set the operation to `PutBucketLogging` with path `LoggingStatus`. When I go to generate this code, the generator will add the `LoggingStatus` field to the CRD, but the underlying go type for that field (and any of its members) are not added to the type defs. This is because [this part of the generator](https://github.com/aws-controllers-k8s/code-generator/blob/5a40e37032506f12a07b7de12a6005b525d13ad9/pkg/model/model.go#L272-L280) ensures that all type defs are in the input or output of some API - but in this case `PutBucketLogging` is neither.

> Also No unit tests added?
I don't see any unit tests for the type defs methods. Any pointers?

vijtrip2:
> > Is this a PR in progress?
> 
> Nope. This is just a small PR.
> 
> > From the code changes, I could not relate how these change will allow mentioning spec fields from another operation. Can you please elaborate for me?
> 
> I want to add `LoggingStatus` as a `spec` field for S3. To do this, I would add a new field in the configuration and set the operation to `PutBucketLogging` with path `LoggingStatus`. When I go to generate this code, the generator will add the `LoggingStatus` field to the CRD, but the underlying go type for that field (and any of its members) are not added to the type defs. This is because [this part of the generator](https://github.com/aws-controllers-k8s/code-generator/blob/5a40e37032506f12a07b7de12a6005b525d13ad9/pkg/model/model.go#L272-L280) ensures that all type defs are in the input or output of some API - but in this case `PutBucketLogging` is neither.
> 
> > Also No unit tests added?
> 
> I don't see any unit tests for the type defs methods. Any pointers?

Thanks for explaining. We should ideally make an effort to increase test coverage of our codebase (maybe a collective effort as a team in coming sprints) but it's definitely out of scope for this PR.

LGTM. I'll let one more team member review this before merging.

vijtrip2 (approved):

pkg/model/crd.go
```diff
@@ -236,6 +241,17 @@ func (r *CRD) SpecFieldNames() []string {
...
+// Spec fields
+func (r *CRD) SpecFieldShapeNames() []string {
+	res := make([]string, 0, len(r.SpecFields))
+	for _, field := range r.SpecFields {
+		res = append(res, field.ShapeRef.ShapeName)
+	}
+	sort.Strings(res)
+	return res
+}
+
```
jaypipes, line 254: This isn't needed, AFAICT, nor is the `CRD.IsShapeInSpec` method below since the block of code you added above in `CRD.HasShapeAsMember` effectively will return true when the CRD has a shape as a member (or nested member). I believe you only added this as a caching mechanism in the GetTypeDefs() function, is that right?

I'd prefer if we kept the code as simple as possible and so think you can remove this and the `IsShapeInSpec` methods.

jaypipes (changes requested):
@RedbackThomson please see my inline comment.

Also, to unit test this functionality, simply add the generator config changes you are testing out for the S3 controller's update code paths to the testdata/ generator yaml for S3:

https://github.com/aws-controllers-k8s/code-generator/blob/5a40e37032506f12a07b7de12a6005b525d13ad9/pkg/testdata/models/apis/s3/0000-00-00/generator.yaml

and add a test for the now-expected new Spec field for Logging in this unit test file:

https://github.com/aws-controllers-k8s/code-generator/blob/5a40e37032506f12a07b7de12a6005b525d13ad9/pkg/model/model_s3_test.go#L70

and a test for the now-expected LoggingStatus type defs in the same model_s3_test.go file. You can use the `testutil.GetTypeDefByName` utility function from here:

https://github.com/aws-controllers-k8s/code-generator/blob/5a40e37032506f12a07b7de12a6005b525d13ad9/pkg/testutil/get.go#L45

pkg/model/crd.go
RedbackThomson: Oh wow. You are absolutely correct. My change to `CRD.HasShapeAsMember` came after I added these other methods, and that is the only thing that was necessary. Thank you!

RedbackThomson:
Cleaned up unnecessary methods and added unit tests for the S3 `Logging` type

A-Hilaly (approved):
👍

pkg/model/model_s3_test.go
```diff
@@ -88,4 +89,13 @@ func TestS3_Bucket(t *testing.T) {
...
 	assert.Equal(expStatusFieldCamel, attrCamelNames(statusFields))
+
+	expTypeDefCamel := []string{
+		"BucketLoggingStatus",
+		"LoggingEnabled",
+		"TargetGrant",
+	}
+	for _, typeDef := range expTypeDefCamel {
+		assert.NotNil(testutil.GetTypeDefByName(t, g, typeDef))
+	}
```
jaypipes, line 100: :+1:

jaypipes (approved):
Bingo. ++

PR #140, opened 2021-07-29
Expose `SetSDKForStruct` and `SetResourceForStruct` in templates

RedbackThomson:
Description of changes:
Makes `setSDKForStruct` and `setResourceForStruct` public, so that they are accessible for other parts of the generator. Exposes `SetSDKForStruct` as `GoCodeSetSDKForStruct` and `SetResourceForStruct` as `GoCodeSetResourceForStruct` in the generator, and includes `SDKAPI` as a template variable to allow querying of API shapes to use with parameters in the new template methods.

### DIFF
--- a/pkg/generate/ack/apis.go
+++ b/pkg/generate/ack/apis.go
@@ -84,6 +84,7 @@ func APIs(
 		crdFileName := strcase.ToSnake(crd.Kind) + ".go"
 		crdVars := &templateCRDVars{
 			metaVars,
+			m.SDKAPI,
 			crd,
 		}
 		if err = ts.Add(crdFileName, "apis/crd.go.tpl", crdVars); err != nil {
@@ -105,5 +106,6 @@ type templateAPIVars struct {
 // code for a single top-level resource's API definition
 type templateCRDVars struct {
 	templateset.MetaVars
-	CRD *ackmodel.CRD
+	SDKAPI *ackmodel.SDKAPI
+	CRD    *ackmodel.CRD
 }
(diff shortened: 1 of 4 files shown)


pkg/generate/ack/controller.go
```diff
@@ -93,6 +94,12 @@ var (
 		"GoCodeSetDeleteInput": func(r *ackmodel.CRD, sourceVarName string, targetVarName string, indentLevel int) string {
 			return code.SetSDK(r.Config(), r, ackmodel.OpTypeDelete, sourceVarName, targetVarName, indentLevel)
 		},
+		"GoCodeSetOperationStruct": func(r *ackmodel.CRD, targetFieldName string, targetVarName string, targetShapeRef *awssdkmodel.ShapeRef, sourceFieldPath string, sourceVarName string, indentLevel int) string {
```
jaypipes, line 97: Can you call this `GoCodeSetSDKFromStruct` in order to match the name of the associated `pkg/generate/code` function? (Yes, I know there are other functions that do not match the name of the associated function, but I plan to clean those up in a separate PR...)

pkg/generate/ack/apis.go
```diff
@@ -105,5 +106,6 @@ type templateAPIVars struct {
 // code for a single top-level resource's API definition
 type templateCRDVars struct {
 	templateset.MetaVars
-	CRD *ackmodel.CRD
+	SDKAPI *ackmodel.SDKAPI
```
jaypipes, line 109: You might consider adding this to the "root" `templateset.MetaVars` at some point in the future.

jaypipes (changes requested):
Teeny renaming request inline, otherwise looks great :)

jaypipes (approved):
:+1:

PR #145, opened 2021-08-03
Remove loops from GetWrapperOutputShape

chlunde:
I believe two loops were left over from refactoring:

* One loop over fieldPathParts have been replaced with recursion. The
  current loop will only execute the first element, so we can remove the
  loop.

* Another loop over map keys, but we can simply use a map lookup
  instead for a bit better performance

There is already a unit test for this code.


A-Hilaly (approved):
Nice! Thanks @chlunde :)

jaypipes (approved):
Nice cleanup @chlunde thank you! :)

PR #149, opened 2021-08-05
Install flag for helm and additional kustomize file for raw yaml for pure Namespace scoping

ryansteakley:
Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/770

Description of changes:

- Adds `namespacedInstallation` flag for helm` values.yaml` to specify that `Role `and` RoleBinding` should be used instead of `ClusterRole `and `ClusterRoleBinding`
- Specifies that `watchNamespace `in helm` values.yaml` must be set if using `namespacedInstallation` flag.
- Contains a sed to replace the original output of the `role-controller.yaml` file with the template code that allows for the programmatic setting of `namespacedInstallation` in the helm` values.yaml`
- Adds a `overlays/namespaced` directory for the raw yamls in config. This folder contains the json patch files, `role.yaml` and `role-binding.yaml` that will modify the `ClusterRole` and `ClusterRoleBinding` files to be `Role `and `RoleBinding`. This allows the user to apply these changes using raw yaml and not helm if these wish.

Tested locally 

### DIFF
--- a/scripts/build-controller-release.sh
+++ b/scripts/build-controller-release.sh
@@ -201,8 +201,12 @@ controller-gen rbac:roleName=$K8S_RBAC_ROLE_NAME paths=./... output:rbac:artifac
 # controller-gen rbac outputs a ClusterRole definition in a
 # $config_output_dir/rbac/role.yaml file. We have some other standard Role
 # files for a reader and writer role, so here we rename the `role.yaml` file to
-# `cluster-role-controller.yaml` to better reflect what is in that file.
-mv $helm_output_dir/templates/role.yaml $helm_output_dir/templates/cluster-role-controller.yaml
+# `cluster-role-controller.yaml` to better reflect what is in that file. We additionally add the ability
+# for the user to specify if they want the role to be ClusterRole or Role by specifying installation scope
+# in the helm values.yaml
+cp -r $ROOT_DIR/templates/helm/templates/_namespaced-install.tpl $helm_output_dir/templates/
+sed -e '1r '"$helm_output_dir/templates/_namespaced-install.tpl"''  -e '1, 7d'  $helm_output_dir/templates/role.yaml > $helm_output_dir/templates/cluster-role-controller.yaml
+rm $helm_output_dir/templates/role.yaml
 
 popd 1>/dev/null
 
--- a/templates/config/overlays/namespaced/kustomization.yaml.tpl
+++ b/templates/config/overlays/namespaced/kustomization.yaml.tpl
@@ -1,0 +1,15 @@
+resources:
+- ../../default
+patches:
+- path: role.json
+  target:
+    group: rbac.authorization.k8s.io
+    version: v1
+    kind: ClusterRole
+    name: ack-sagemaker-controller
+- path: role-binding.json
+  target:
+    group: rbac.authorization.k8s.io
+    version: v1
+    kind: ClusterRoleBinding
+    name: ack-sagemaker-controller-rolebinding
--- a/templates/config/overlays/namespaced/role-binding.json
+++ b/templates/config/overlays/namespaced/role-binding.json
@@ -1,0 +1,14 @@
+[{"op": "replace",
+    "path": "/kind",
+    "value": "RoleBinding"},
+ 
+ 
+   {"op": "add",
+    "path": "/metadata/namespace",
+    "value":  "ack-system"},
+
+
+    {"op": "replace",
+        "path": "/roleRef/kind",
+        "value": "Role"}
+   ]
--- a/templates/config/overlays/namespaced/role.json
+++ b/templates/config/overlays/namespaced/role.json
@@ -1,0 +1,9 @@
+  [{"op": "replace",
+   "path": "/kind",
+   "value": "Role"},
+
+   
+  {"op": "add",
+   "path": "/metadata/namespace",
+   "value": "ack-system"}
+  ]
--- a/templates/helm/templates/_namespaced-install.tpl
+++ b/templates/helm/templates/_namespaced-install.tpl
@@ -1,0 +1,15 @@
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+{{ if not .Values.namespacedInstallation }}
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  name: ack-sagemaker-controller
+{{ else }}
+kind: Role
+metadata:
+  creationTimestamp: null
+  name: ack-sagemaker-controller
+  namespace: {{ .Release.Namespace }}
+{{ end }}
--- a/templates/helm/templates/cluster-role-binding.yaml.tpl
+++ b/templates/helm/templates/cluster-role-binding.yaml.tpl
@@ -1,4 +1,5 @@
 apiVersion: rbac.authorization.k8s.io/v1
+{{ "{{ if not .Values.namespacedInstallation }}" }}
 kind: ClusterRoleBinding
 metadata:
   name: {{ "{{ include \"app.fullname\" . }}" }}
@@ -6,6 +7,16 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: ack-{{ .ServiceIDClean }}-controller
+{{ "{{ else }}" }}
+kind: RoleBinding
+metadata:
+  name: {{ "{{ include \"app.fullname\" . }}" }}
+  namespace: {{ "{{ .Release.Namespace }}" }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: ack-{{ .ServiceIDClean }}-controller
+{{ "{{ end }}" }}
 subjects:
 - kind: ServiceAccount
   name: {{ "{{ include \"service-account.name\" . }}" }}
--- a/templates/helm/values.yaml.tpl
+++ b/templates/helm/values.yaml.tpl
@@ -47,6 +47,10 @@ log:
 # If specified, the service controller will watch for object creation only in the provided namespace
 watchNamespace: ""
 
+# If specified, the service controller will use Role/Rolebinding instead of ClusterRole/ClusterRoleBindings
+# If set to true, user must set watchNamespace as well.
+namespacedInstallation: false
+
 resourceTags:
   # Configures the ACK service controller to always set key/value pairs tags on resources that it manages.
   - services.k8s.aws/managed=true
--- a/scripts/build-controller.sh
+++ b/scripts/build-controller.sh
@@ -207,8 +207,13 @@ controller-gen rbac:roleName=$K8S_RBAC_ROLE_NAME paths=./... output:rbac:artifac
 # controller-gen rbac outputs a ClusterRole definition in a
 # $config_output_dir/rbac/role.yaml file. We have some other standard Role
 # files for a reader and writer role, so here we rename the `role.yaml` file to
-# `cluster-role-controller.yaml` to better reflect what is in that file.
-mv $config_output_dir/rbac/role.yaml $config_output_dir/rbac/cluster-role-controller.yaml
+# `role-controller.yaml` to better reflect what is in that file.
+mv $config_output_dir/rbac/role.yaml $config_output_dir/rbac/role-controller.yaml
+# Copy definitions for json patches which allow the user to patch the controller
+# with Role/Rolebinding and be purely namespaced scoped instead of using Cluster/ClusterRoleBinding
+# using kustomize
+mkdir -p $config_output_dir/overlays/namespaced
+cp -r $ROOT_DIR/templates/config/overlays/namespaced/* $config_output_dir/overlays/namespaced
 
 popd 1>/dev/null
 


templates/config/overlays/namespaced/kustomization.yaml.tpl
```diff
@@ -0,0 +1,15 @@
+resources:
```
RedbackThomson, line 1: I don't believe this needs to be a `.yaml.tpl` file since it contains no Go template injections. Renaming it to `.yaml` and adding it to the list of files that is copied when running `ack-generate release` ([adding it here](https://github.com/aws-controllers-k8s/code-generator/blob/e2a9338988f0f7c78c6073c959ff4994ef8394e9/pkg/generate/ack/release.go#L33)) should suffice.

templates/helm/values.yaml.tpl
```diff
@@ -47,6 +47,10 @@ log:
 # If specified, the service controller will watch for object creation only in the provided namespace
 watchNamespace: ""
 
+# If specified, the service controller will use Role/Rolebinding instead of ClusterRole/ClusterRoleBindings
```
RedbackThomson, line 50: nit: Make the description a little higher level. Suggestion like "If specified, the service controller will only be able to operate within a single namespace"

templates/config/overlays/namespaced/role-binding.json
```diff
@@ -0,0 +1,14 @@
+[{"op": "replace",
+    "path": "/kind",
+    "value": "RoleBinding"},
+
```
RedbackThomson, line 4: Please remove excess whitespace

templates/helm/templates/_namespaced-install.tpl
```diff
@@ -0,0 +1,15 @@
+
+---
```
RedbackThomson, line 2: Please remove these two lines of whitespace

scripts/build-controller-release.sh
```diff
@@ -201,8 +201,12 @@ controller-gen rbac:roleName=$K8S_RBAC_ROLE_NAME paths=./... output:rbac:artifac
...
 # $config_output_dir/rbac/role.yaml file. We have some other standard Role
 # files for a reader and writer role, so here we rename the `role.yaml` file to
-# `cluster-role-controller.yaml` to better reflect what is in that file.
-mv $helm_output_dir/templates/role.yaml $helm_output_dir/templates/cluster-role-controller.yaml
+# `cluster-role-controller.yaml` to better reflect what is in that file. We additionally add the ability
+# for the user to specify if they want the role to be ClusterRole or Role by specifying installation scope
+# in the helm values.yaml
+cp -r $ROOT_DIR/templates/helm/templates/_namespaced-install.tpl $helm_output_dir/templates/
+sed -e '1r '"$helm_output_dir/templates/_namespaced-install.tpl"''  -e '1, 7d'  $helm_output_dir/templates/role.yaml > $helm_output_dir/templates/cluster-role-controller.yaml
+rm $helm_output_dir/templates/role.yaml
```
RedbackThomson, line 209: This could be simplified into:
```bash
sed -e '1r '"$ROOT_DIR/templates/helm/templates/_namespaced-install.tpl"''  -e '1, 7d'  $helm_output_dir/templates/role.yaml > $helm_output_dir/templates/controller-role.yaml
rm $helm_output_dir/templates/role.yaml
```

templates/helm/templates/_namespaced-install.tpl
```diff
@@ -0,0 +1,15 @@
+
```
RedbackThomson, line 1: Please rename this file to `.yaml`. `.tpl` is reserved for Go template files.

Also perhaps since this file is used to patch, we could instead call it `controller-role-kind-patch.yaml`?

templates/helm/values.yaml.tpl
```diff
@@ -47,6 +47,10 @@ log:
...
+# If set to true, user must set watchNamespace as well.
```
RedbackThomson, line 51: You can actually force this by adding a `{{ required ... }}` block inside one of your conditionals.
See https://v2.helm.sh/docs/developing_charts/#using-the-required-function for information about required.

I could imagine you have, in :
```
{{ if not .Values.namespacedInstallation }}
...
{{ else }}
{{ required "watchNamespace must be set for namespaced installation" .Values.watchNamespace }}
...
```

RedbackThomson (commented):
+1 on using JSON patches with Kustomize
+1 on working around the limitation with the autogenerated role YAML file

I have some inline suggestions about file extensions and names.

templates/helm/templates/_namespaced-install.tpl
ryansteakley: Sure

templates/helm/values.yaml.tpl
ryansteakley: I'll into that and iterate

templates/config/overlays/namespaced/kustomization.yaml.tpl
ryansteakley: Sure doesn't need to be a .tpl file. Should it be a part of that ack-generate release however since all those other files pertain to helm/templates

templates/config/overlays/namespaced/kustomization.yaml
```diff
@@ -0,0 +1,15 @@
+resources:
+- ../../default
+patches:
+- path: role.json
+  target:
+    group: rbac.authorization.k8s.io
+    version: v1
+    kind: ClusterRole
+    name: ack-sagemaker-controller
```
vijtrip2, line 9: Does "sagemaker" need to be hardcoded here?

templates/helm/templates/_controller-role-kind-patch.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+{{ if not .Values.namespacedInstallation }}
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  name: ack-sagemaker-controller
```
vijtrip2, line 6: hardcoded name

templates/helm/templates/_controller-role-kind-patch.yaml
```diff
@@ -0,0 +1,13 @@
...
+{{ else if and .Values.namespacedInstallation (required "watchNamespace must be set for namespaced installation" .Values.watchNamespace) }}
+kind: Role
+metadata:
+  creationTimestamp: null
+  name: ack-sagemaker-controller
```
vijtrip2, line 11: hardcoded name

templates/config/overlays/namespaced/kustomization.yaml
ryansteakley: Thought I removed this must have pushed it again by accident :(

RedbackThomson: Fair point. These might need to be made into Go templates and inject `{{ .ServiceIDClean }}` just like the `cluster-role-binding.yaml.tpl`

ryansteakley: Yep had to make them into go templates and put them in release.go and controller.go respectively

jaypipes (commented):
@ryansteakley you marked a number of conversations between you and @RedbackThomson as resolved, but I don't actually see changes made... did you forget to `git push` your changes?

templates/helm/templates/_controller-role-kind-patch.yaml.tpl
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
RedbackThomson, line 1: We only use this file once - to patch the `controller-role` resource. I'm worried that leaving the `_controller-role-kind-patch.yaml` file in the `templates/helm/templates` directory, Helm will try to do a `kubectl apply -f _controller-role-kind-patch.yaml` when we install the template. I think we have 2 choices here:
1. (Probably preferred) Delete the file after we are done patching
2. Add the file name to [`.helmignore`](https://helm.sh/docs/chart_template_guide/helm_ignore_file/) in the directory.

templates/config/overlays/namespaced/role-binding.json
RedbackThomson: Still some excess whitespace here. Probably can go to 1 line per object:
```json
[{"op": "replace", "path": "/kind", "value": "RoleBinding"},
{"op": "add", "path": "/metadata/namespace", "value":  "ack-system"},
{"op": "replace","path": "/roleRef/kind","value": "Role"}] 
```

RedbackThomson (changes requested):
Just have one more concern about our helm patch file - see inline comments

ryansteakley:
> @ryansteakley you marked a number of conversations between you and @RedbackThomson as resolved, but I don't actually see changes made... did you forget to `git push` your changes?

Going to look into it, it is very possible some changes got git stash and never applied again

templates/config/overlays/namespaced/role-binding.json
```diff
@@ -0,0 +1,3 @@
+[{"op": "replace", "path": "/kind", "value": "RoleBinding"},
+{"op": "add", "path": "/metadata/namespace", "value":  "ack-system"},
```
jaypipes, line 2: This is replacing the namespace that should be supplied by the user in `.Values.watchNamespace` with the hard-coded string "ack-system", which is defeating the original purpose of a namespace-scoped installation, no?

templates/config/overlays/namespaced/role.json
```diff
@@ -0,0 +1,2 @@
+[{"op": "replace", "path": "/kind", "value": "Role"},
+{"op": "add", "path": "/metadata/namespace", "value": "ack-system"}]
```
jaypipes, line 2: same comment as above about hard-coded "ack-system".

jaypipes (changes requested):
this needs a bit more thought put into it. I need to do some more review here.

templates/config/overlays/namespaced/role-binding.json
RedbackThomson: This file is for the Kustomize overlay - no Helm values to replace, here. All of these Kustomize files use `ack-system` as the default namespace. The user will need to override the values for all resources anyway, so may as well make this one match.

templates/helm/values.yaml.tpl
```diff
@@ -47,6 +47,10 @@ log:
 # If specified, the service controller will watch for object creation only in the provided namespace
 watchNamespace: ""
 
+# If specified, the service controller will only be able to operate in a single namespace. Set by default to K8S_NAMESPACE
+# If set to true, user must set watchNamespace as well.
+namespacedInstallation: false
```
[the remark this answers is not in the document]
RedbackThomson: A customer may want to set the `watchNamespace` but still have their resource be cluster-scoped. This allows for flexibility in changing the `watchNamespace` value later, and not having to redeploy anything except the deployment. It also means that a user could install the controller into the `ack-system` namespace and watch another namespace. Installing as `namespacedInstallation = true` requires the Helm chart be installed into the `watchNamespace`

templates/helm/templates/_controller-role-kind-patch.yaml.tpl
```diff
@@ -0,0 +1,13 @@
...
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  name: ack-{{ .ServiceIDClean }}-controller
+{{ "{{ else if and .Values.namespacedInstallation (required \"watchNamespace must be set for namespaced installation\"  .Values.watchNamespace) }}" }}
+kind: Role
+metadata:
+  creationTimestamp: null
+  name: ack-{{ .ServiceIDClean }}-controller
+  namespace: {{ "{{ .Release.Namespace }}" }}
```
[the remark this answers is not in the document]
ryansteakley: I agree here

templates/helm/templates/cluster-role-binding.yaml.tpl
```diff
@@ -1,11 +1,22 @@
...
   name: {{ "{{ include \"app.fullname\" . }}" }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: ack-{{ .ServiceIDClean }}-controller
+{{ "{{ else }}" }}
+kind: RoleBinding
+metadata:
+  name: {{ "{{ include \"app.fullname\" . }}" }}
+  namespace: {{ "{{ .Release.Namespace }}" }}
```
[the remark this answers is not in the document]
ryansteakley: agreed

scripts/build-controller-release.sh
```diff
@@ -201,8 +201,12 @@ controller-gen rbac:roleName=$K8S_RBAC_ROLE_NAME paths=./... output:rbac:artifac
 # controller-gen rbac outputs a ClusterRole definition in a
 # $config_output_dir/rbac/role.yaml file. We have some other standard Role
 # files for a reader and writer role, so here we rename the `role.yaml` file to
-# `cluster-role-controller.yaml` to better reflect what is in that file.
-mv $helm_output_dir/templates/role.yaml $helm_output_dir/templates/cluster-role-controller.yaml
+# `cluster-role-controller.yaml` to better reflect what is in that file. We additionally add the ability
+# for the user to specify if they want the role to be ClusterRole or Role by specifying installation scope
+# in the helm values.yaml
+sed -e '1r '"$helm_output_dir/templates/_controller-role-kind-patch.yaml"''  -e '1, 7d'  $helm_output_dir/templates/role.yaml > $helm_output_dir/templates/cluster-role-controller.yaml
```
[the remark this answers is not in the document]
ryansteakley: Issue is that the role.yaml file is auto-generated by controller-gen. Going to have to be some strange way of overriding the file but yes, maybe there is a clearer way of doing it

templates/helm/templates/cluster-role-binding.yaml.tpl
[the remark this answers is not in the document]
RedbackThomson: The Role needs to be installed into the same namespace as the SA it will be attached to. That SA, and the controller deployment, are deployed into `{{ .Release.Namespace }}`. `.Values.watchNamespace` is an independent variable passed to the controller.

templates/config/overlays/namespaced/kustomization.yaml.tpl
jaypipes: This does need to be a .tpl file because it contains template variables..

templates/helm/templates/cluster-role-binding.yaml.tpl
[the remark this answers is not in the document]
jaypipes: > The Role needs to be installed into the same namespace as the SA it will be attached to. That SA, and the controller deployment, are deployed into `{{ .Release.Namespace }}`. `.Values.watchNamespace` is an independent variable passed to the controller.

In the case of a namespaced installation, though, shouldn't they be the same thing? It's not going to work if the namespaces are different...

templates/helm/values.yaml.tpl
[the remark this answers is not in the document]
jaypipes: > A customer may want to set the `watchNamespace` but still have their resource be cluster-scoped. 

Why? Isn't the whole point of namespaced installation to limit the CRDs, the RBACs that can read those CRs and the namespace that those CRs can be created in to just a single namespace?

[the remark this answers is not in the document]
RedbackThomson: I am trying to not conflate the two ideas:
1. Watch a single namespace
2. Be installed into the scope of the single namespace

Users may want 1 but not 2. But if they want 2, it is required that they also do 1. 

If you want to bundle them up into the same thing and ship them together, then ok I'm on board. I wanted to offer the choice, at least.

templates/helm/templates/cluster-role-binding.yaml.tpl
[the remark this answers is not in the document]
RedbackThomson: If `{{ .Values.namespacedInstallation }}`, then yes they will need to be the same thing. But we cannot impose a limit on `{{ .Release.Namespace }}`, as that's defined at install-time.

RedbackThomson:
As described on Slack with @jaypipes and @ryansteakley:
With the current implementation of `.Values.watchNamespace`, there are two places where the namespace will need to be specified when installing with namespace-scope. Those two places are the `{{ .Release.Namespace }}` (`--namespace` when calling `helm install`) as well as the `--set watchNamespace` value. When installing in namespace-scope, these values NEED to be identical, otherwise the `Role` will not have the privilege to access the resources.

I suggest that we remove the `watchNamespace` value altogether, and default the `--watch-namespace` argument to `{{ .Release.Namespace}}`. We should also have a value to track whether the controller should be installed into cluster or namespace scope - e.g. `clusterScope: cluster|namespace`. If the `clusterScope == namespace`, then we change all `ClusterRole` to `Role`, and set the `--watch-namespace` argument. 

This would reduce the chance of a user accidentally applying the Helm chart into a namespace where the deployment does not have sufficient privileges. It also reduces the number of independent variables that need to align for a correct configuration.

templates/helm/templates/_helpers.tpl
```diff
@@ -30,3 +30,9 @@ If release name contains chart name it will be used as a full name.
 {{- define "service-account.name" -}}
     {{ default "default" .Values.serviceAccount.name }}
 {{- end -}}
+
+{{- define "watch-namespace" -}}
+{{- if eq .Values.installScope "namespace" -}}
+{{- .Release.Namespace -}}
+{{- end -}}
+{{- end -}}
```
RedbackThomson, line 38: Have we confirmed that this defaults to an empty string when install scope is cluster?

templates/helm/templates/cluster-role-binding.yaml.tpl
```diff
@@ -1,11 +1,22 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: ack-{{ .ServiceIDClean }}-controller
```
RedbackThomson, line 18: nit: Remove duplication by moving `apiGroup` and `name` to after `{{ end }}` (and removing from both conditional blocks)

templates/helm/values.yaml.tpl
```diff
@@ -44,8 +44,9 @@ log:
   enable_development_logging: false
   level: info
 
-# If specified, the service controller will watch for object creation only in the provided namespace
-watchNamespace: ""
+# Set to namespace to install the controller in a namespaced scope, will only watch for object creation
```
RedbackThomson, line 47: nit: For clarity,
```suggestion
# Set to "namespace" to install the controller in a namespaced scope, will only watch for object creation
```

templates/helm/templates/_helpers.tpl
ryansteakley:
```
Environment:
      K8S_NAMESPACE:                   ack-system (v1:metadata.namespace)
      AWS_ACCOUNT_ID:                  123456789012
      AWS_REGION:                      us-west-1
      AWS_ENDPOINT_URL:                
      ACK_WATCH_NAMESPACE:             
      ACK_ENABLE_DEVELOPMENT_LOGGING:  true
      ACK_LOG_LEVEL:                   debug
      ACK_RESOURCE_TAGS:               services.k8s.aws/managed=true,services.k8s.aws/created=%UTCNOW%,services.k8s.aws/namespace=%KUBERNETES_NAMESPACE%
      AWS_ROLE_ARN:                    arn:aws:iam::123456789012:role/ack-controller-role-namespaced
      AWS_WEB_IDENTITY_TOKEN_FILE:     /var/run/secrets/eks.amazonaws.com/serviceaccount/token
```

Defaults to an empty string when using cluster.
When using namespace, value is correctly passed.

```
  Environment:
      K8S_NAMESPACE:                   ack-system (v1:metadata.namespace)
      AWS_ACCOUNT_ID:                  123456789012
      AWS_REGION:                      us-west-1
      AWS_ENDPOINT_URL:                
      ACK_WATCH_NAMESPACE:             ack-system
      ACK_ENABLE_DEVELOPMENT_LOGGING:  true
      ACK_LOG_LEVEL:                   debug
      ACK_RESOURCE_TAGS:               services.k8s.aws/managed=true,services.k8s.aws/created=%UTCNOW%,services.k8s.aws/namespace=%KUBERNETES_NAMESPACE%
      AWS_ROLE_ARN:                    arn:aws:iam::123456789012:role/ack-controller-role-namespaced
      AWS_WEB_IDENTITY_TOKEN_FILE:     /var/run/secrets/eks.amazonaws.com/serviceaccount/token

```

jaypipes (approved):
Yeah, I like it.

jaypipes:
@RedbackThomson feel free to lgtm if you're also good with this.

A-Hilaly (approved):
:+1:

RedbackThomson (approved):
Ship it!
/lgtm

PR #150, opened 2021-08-09
Minor Refactor in `pkg/generate/code`

brycahta:
Issue #, if available: [#890](https://github.com/aws-controllers-k8s/community/issues/890)
* This is preliminary work to aid in resolving this issue. A PR will follow to close out this issue completely.

Description of changes:
* Adds `getSanitizedMemberPath` helper to **check.go**
* Adds new file `common.go` containing helpers that can be used throughout `code` pkg. These 2 helpers make it easier to extract identifier(s) from a given Shape or CRD

Testing:
* `make test` ✅

### DIFF
--- a/pkg/generate/code/check.go
+++ b/pkg/generate/code/check.go
@@ -132,32 +132,17 @@ func checkRequiredFieldsMissingFromShape(
 			missing = append(missing, primaryARNCondition)
 			continue
 		}
-		cleanMemberNames := names.New(memberName)
-		cleanMemberName := cleanMemberNames.Camel
 
-		resVarPath := koVarName
-		// Check that the field has potentially been renamed
-		renamedName, wasRenamed := r.InputFieldRename(
-			op.Name, memberName,
-		)
-		_, found := r.SpecFields[renamedName]
-		if found && !wasRenamed {
-			resVarPath = resVarPath + r.Config().PrefixConfig.SpecField + "." + cleanMemberName
-		} else if found {
-			resVarPath = resVarPath + r.Config().PrefixConfig.SpecField + "." + renamedName
-		} else {
-			_, found = r.StatusFields[memberName]
-			if !found {
-				// If it isn't in our spec/status fields, we have a problem!
-				msg := fmt.Sprintf(
-					"GENERATION FAILURE! there's a required field %s in "+
-						"Shape %s that isn't in either the CR's Spec or "+
-						"Status structs!",
-					memberName, shape.ShapeName,
-				)
-				panic(msg)
-			}
-			resVarPath = resVarPath + r.Config().PrefixConfig.StatusField + "." + cleanMemberName
+		resVarPath, err := getSanitizedMemberPath(memberName, r, op, koVarName)
+		if err != nil {
+			// If it isn't in our spec/status fields, we have a problem!
+			msg := fmt.Sprintf(
+				"GENERATION FAILURE! there's a required field %s in "+
+					"Shape %s that isn't in either the CR's Spec or "+
+					"Status structs!",
+				memberName, shape.ShapeName,
+			)
+			panic(msg)
 		}
 		missing = append(missing, fmt.Sprintf("%s == nil", resVarPath))
 	}
@@ -165,4 +150,35 @@ func checkRequiredFieldsMissingFromShape(
 	// is not created yet
 	missingCondition := strings.Join(missing, " || ")
 	return fmt.Sprintf("%sreturn %s\n", indent, missingCondition)
+}
+
+// getSanitizedMemberPath takes a shape member field, checks for renames, checks
+// for existence in Spec and Status, then constructs and returns the var path.
+// Returns error if memberName is not present in either Spec or Status.
+func getSanitizedMemberPath(
+	memberName string,
+	r *model.CRD,
+	op *awssdkmodel.Operation,
+	koVarName string) (string, error) {
+	resVarPath := koVarName
+	cleanMemberNames := names.New(memberName)
+	cleanMemberName := cleanMemberNames.Camel
+	// Check that the field has potentially been renamed
+	renamedName, wasRenamed := r.InputFieldRename(
+		op.Name, memberName,
+	)
+	_, found := r.SpecFields[renamedName]
+	if found && !wasRenamed {
+		resVarPath = resVarPath + r.Config().PrefixConfig.SpecField + "." + cleanMemberName
+	} else if found {
+		resVarPath = resVarPath + r.Config().PrefixConfig.SpecField + "." + renamedName
+	} else {
+		_, found = r.StatusFields[memberName]
+		if !found {
+			return "", fmt.Errorf(
+				"the required field %s is NOT present in CR's Spec or Status", memberName)
+		}
+		resVarPath = resVarPath + r.Config().PrefixConfig.StatusField + "." + cleanMemberName
+	}
+	return resVarPath, nil
 }
--- a/pkg/generate/code/common.go
+++ b/pkg/generate/code/common.go
@@ -14,22 +14,21 @@
 package code
 
 import (
-	"errors"
-	"fmt"
+	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
 
 	"github.com/aws-controllers-k8s/code-generator/pkg/model"
 	"github.com/aws-controllers-k8s/code-generator/pkg/util"
-
-	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
 )
 
-// FindIdentifierInShape returns the identifier field of a given shape which
+// FindIdentifiersInShape returns the identifier fields of a given shape which
 // can be singular or plural. Errors iff multiple identifier fields detected
 // in the shape.
-func FindIdentifierInShape(
+func FindIdentifiersInShape(
 	r *model.CRD,
-	shape *awssdkmodel.Shape) (string, error) {
+	shape *awssdkmodel.Shape) []string {
 	identifierLookup := []string{
+		"Id",
+		"Ids",
 		"Name",
 		"Names",
 		r.Names.Original + "Name",
@@ -37,29 +36,25 @@ func FindIdentifierInShape(
 		r.Names.Original + "Id",
 		r.Names.Original + "Ids",
 	}
-	identifier := ""
+	var identifiers []string
 
 	for _, memberName := range shape.MemberNames() {
 		if util.InStrings(memberName, identifierLookup) {
-			if identifier == "" {
-				identifier = memberName
-			} else {
-				return "", errors.New(fmt.Sprintf(
-					"Found multiple possible identifiers for %s: %s, %s ",
-					r.Names.Original, identifier, memberName))
-			}
+			identifiers = append(identifiers, memberName)
 		}
 	}
 
-	return identifier, nil
+	return identifiers
 }
 
-// FindIdentifierInCRD returns the identifier field of a given CRD which
+// FindIdentifiersInCRD returns the identifier field of a given CRD which
 // can be singular or plural. Errors iff multiple identifier fields detected
 // in the CRD.
-func FindIdentifierInCRD(
-	r *model.CRD) (string, error) {
+func FindIdentifiersInCRD(
+	r *model.CRD) []string {
 	identifierLookup := []string{
+		"Id",
+		"Ids",
 		"Name",
 		"Names",
 		r.Names.Original + "Name",
@@ -67,7 +62,7 @@ func FindIdentifierInCRD(
 		r.Names.Original + "Id",
 		r.Names.Original + "Ids",
 	}
-	identifier := ""
+	var identifiers []string
 
 	for _, id := range identifierLookup {
 		_, found := r.SpecFields[id]
@@ -75,15 +70,9 @@ func FindIdentifierInCRD(
 			_, found = r.StatusFields[id]
 		}
 		if found {
-			if identifier == "" {
-				identifier = id
-			} else {
-				return "", errors.New(fmt.Sprintf(
-					"Found multiple possible identifiers for %s: %s, %s ",
-					r.Names.Original, identifier, id))
-			}
+			identifiers = append(identifiers, id)
 		}
 	}
 
-	return identifier, nil
+	return identifiers
 }
--- a/pkg/generate/code/common.go
+++ b/pkg/generate/code/common.go
@@ -1,0 +1,89 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package code
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/aws-controllers-k8s/code-generator/pkg/model"
+	"github.com/aws-controllers-k8s/code-generator/pkg/util"
+
+	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
+)
+
+// FindIdentifierInShape returns the identifier field of a given shape which
+// can be singular or plural. Errors iff multiple identifier fields detected
+// in the shape.
+func FindIdentifierInShape(
+	r *model.CRD,
+	shape *awssdkmodel.Shape) (string, error) {
+	identifierLookup := []string{
+		"Name",
+		"Names",
+		r.Names.Original + "Name",
+		r.Names.Original + "Names",
+		r.Names.Original + "Id",
+		r.Names.Original + "Ids",
+	}
+	identifier := ""
+
+	for _, memberName := range shape.MemberNames() {
+		if util.InStrings(memberName, identifierLookup) {
+			if identifier == "" {
+				identifier = memberName
+			} else {
+				return "", errors.New(fmt.Sprintf(
+					"Found multiple possible identifiers for %s: %s, %s ",
+					r.Names.Original, identifier, memberName))
+			}
+		}
+	}
+
+	return identifier, nil
+}
+
+// FindIdentifierInCRD returns the identifier field of a given CRD which
+// can be singular or plural. Errors iff multiple identifier fields detected
+// in the CRD.
+func FindIdentifierInCRD(
+	r *model.CRD) (string, error) {
+	identifierLookup := []string{
+		"Name",
+		"Names",
+		r.Names.Original + "Name",
+		r.Names.Original + "Names",
+		r.Names.Original + "Id",
+		r.Names.Original + "Ids",
+	}
+	identifier := ""
+
+	for _, id := range identifierLookup {
+		_, found := r.SpecFields[id]
+		if !found {
+			_, found = r.StatusFields[id]
+		}
+		if found {
+			if identifier == "" {
+				identifier = id
+			} else {
+				return "", errors.New(fmt.Sprintf(
+					"Found multiple possible identifiers for %s: %s, %s ",
+					r.Names.Original, identifier, id))
+			}
+		}
+	}
+
+	return identifier, nil
+}
--- a/pkg/generate/code/common_test.go
+++ b/pkg/generate/code/common_test.go
@@ -1,0 +1,94 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	 http://aws.amazon.com/apache2.0/
+//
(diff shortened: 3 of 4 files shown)


pkg/generate/code/check.go
```diff
@@ -14,7 +14,9 @@
 package code
 
 import (
+	"errors"
 	"fmt"
+	"github.com/gertd/go-pluralize"
```
brycahta, line 19: cleanup imports

pkg/generate/code/check.go
```diff
@@ -100,16 +104,18 @@ func checkRequiredFieldsMissingFromShape(
...
+	missing := []string{}
 
 	// Loop over the required member fields in the shape and identify whether
 	// the field exists in either the Status or the Spec of the resource and
 	// generate an if condition checking for all required fields having non-nil
 	// corresponding resource Spec/Status values
-	missing := []string{}
+	notRequiredFields := shape.MemberRefs
 	for _, memberName := range shape.Required {
+		delete(notRequiredFields, memberName)
```
RedbackThomson, line 118: `delete` scares me - I think it's modifying `shape.MemberRefs`. Could we copy the map beforehand? https://stackoverflow.com/a/23058707/1278784

pkg/generate/code/check.go
```diff
@@ -132,37 +138,76 @@ func checkRequiredFieldsMissingFromShape(
...
+	// corresponding resource Spec/Status values (required for ReadMany)
+	for field := range notRequiredFields {
+		crKind := strings.ToLower(r.Kind)
+		if strings.Contains(strings.ToLower(field), crKind) {
+			pluralize := pluralize.NewClient()
+			// CR shouldn't have plural of its kind as a field
+			singular := pluralize.Singular(field)
+
+			resVarPath, err := getSanitizedMemberPath(singular, r, op, koVarName)
+			if err != nil {
```
RedbackThomson, line 168: If there is a non-required field, and we don't have a value to plug it, then I don't think that's an error? It would be nice to have an answer for every input, but I wouldn't make that assumption.

pkg/generate/code/check.go
```diff
@@ -132,37 +138,76 @@ func checkRequiredFieldsMissingFromShape(
...
 						"Status structs!",
-					memberName, shape.ShapeName,
+					singular, shape.ShapeName,
 				)
 				panic(msg)
 			}
-			resVarPath = resVarPath + r.Config().PrefixConfig.StatusField + "." + cleanMemberName
+			missing = append(missing, fmt.Sprintf("%s == nil", resVarPath))
+			// Not expecting multiple ID fields
+			break
```
RedbackThomson, line 180: We can't be sure that the one we caught was the ID field, though? Some ReadMany calls may have multiple fields with the resource name (yep.) These are treated more like filters than selectors. I don't have any concrete examples off the top of my head, but my hypothetical would be `InstanceType` and `InstanceName`. We would want to catch `InstanceName` only. Filtering by `InstanceType` would narrow down results but not be useful for our `ReadOne` method.

I believe this is why in the `SetResourceIdentifiers`, we went with a list of expected field keys. `<Resource>Name`, `<Resource>ID`, `ID`, `Name`. We could recycle that list, just pluralising it?

RedbackThomson (commented):
Great first PR. Have some inline comments.

pkg/generate/code/check.go
brycahta: So I'm on the same page, are you describing the scenario where there's a non-required field containing the `crKind`, but not necessarily an ID/Name field within crKind.Spec or crKind.Status?

RedbackThomson: Exactly

pkg/generate/code/check.go
brycahta: I see your point; I'll go with the list in next revision.

Related: did you see my bit about SecurityGroups in the PR description (at the bottom)? It seems like both GroupName and GroupId would be valid-- any thoughts there? I know when Setting Resource Identifiers we panic if there are multiple possibilities

pkg/generate/code/check.go
A-Hilaly: Since go 1.11 it is totally safe to loop and delete a maps elements: https://golang.org/doc/go1.11#performance-compiler

jaypipes: > Since go 1.11 it is totally safe to loop and delete a maps elements: https://golang.org/doc/go1.11#performance-compiler

I think @RedbackThomson was just concerned with mutating the aws-sdk-go Shape struct instead of a copy of the Shape's MemberRefs map.

templates/pkg/resource/sdk_find_read_many.go.tpl
```diff
@@ -51,6 +58,19 @@ func (rm *resourceManager) sdkFind(
...
 }
 
+// requiredFieldsMissingFromReadManyInput returns true if there are any fields
+// for the ReadMany Input shape that are required but not present in the
+// resource's Spec or Status
+func (rm *resourceManager) requiredFieldsMissingFromReadManyInput(
+	r *resource,
+) bool {
+{{- if $customCheckMethod := .CRD.GetCustomCheckRequiredFieldsMissingMethod .CRD.Ops.ReadMany }}
+return rm.{{ $customCheckMethod }}(r)
```
jaypipes, line 68: Needs one TAB indent above.

pkg/generate/code/check.go
```diff
@@ -132,37 +138,76 @@ func checkRequiredFieldsMissingFromShape(
...
+		missing = append(missing, fmt.Sprintf("%s == nil", resVarPath))
+	}
+
+	// Loop over the not-required member fields in the shape and determine
+	// if an ID field can be populated by an ID in either the Status or the Spec of the resource
+	// and generate an if condition checking for all required fields having non-nil
+	// corresponding resource Spec/Status values (required for ReadMany)
+	for field := range notRequiredFields {
+		crKind := strings.ToLower(r.Kind)
+		if strings.Contains(strings.ToLower(field), crKind) {
```
jaypipes, line 162: You probably want to check `strings.HasPrefix` here instead of `strings.Contains`. Consider a `crKind` of `User`. Then consider that there might be a field called `MasterUserName` or `DefaultUsers`, neither of which may refer to a User Kind but rather some other user-ish concept.

pkg/generate/code/check.go
```diff
@@ -132,37 +138,76 @@ func checkRequiredFieldsMissingFromShape(
...
+			pluralize := pluralize.NewClient()
```
jaypipes, line 163: You can place this outside the for loop.

pkg/generate/code/check.go
```diff
@@ -132,37 +138,76 @@ func checkRequiredFieldsMissingFromShape(
...
+			panic(msg)
+		}
+		missing = append(missing, fmt.Sprintf("%s == nil", resVarPath))
+	}
+
+	// Loop over the not-required member fields in the shape and determine
+	// if an ID field can be populated by an ID in either the Status or the Spec of the resource
+	// and generate an if condition checking for all required fields having non-nil
+	// corresponding resource Spec/Status values (required for ReadMany)
+	for field := range notRequiredFields {
```
jaypipes, line 160: please wrap at 80 chars.

pkg/generate/code/check_test.go
```diff
@@ -74,6 +74,27 @@ func TestCheckRequiredFields_Attributes_StatusField(t *testing.T) {
 	)
 }
 
+func TestCheckRequiredFields_Attributes_StatusField_ReadMany(t *testing.T) {
```
jaypipes, line 77: Please remove `_Attributes` from the test name (that refers to APIs that follow the Get Attributes pattern (i.e. SNS and SQS)

pkg/generate/code/check.go
jaypipes: Agree with @RedbackThomson here. Instead of looping through the "non-required fields" like you have here, it may make more sense to simply search for `ID`, `Name`, `<Resource>ID` and `<Resource>Name` fields in the Spec or Status that have a matching field (or matching plural-named field) in the ReadMany Input shape.

pkg/generate/code/check.go
```diff
@@ -132,37 +138,76 @@ func checkRequiredFieldsMissingFromShape(
...
+
 	// Use '||' because if any of the required fields are missing the object
 	// is not created yet
 	missingCondition := strings.Join(missing, " || ")
 	return fmt.Sprintf("%sreturn %s\n", indent, missingCondition)
 }
+
+// getSanitizedMemberPath takes a shape member field, checks for renames, checks for existence in Spec and Status, then
+// constructs and returns the var path. Returns error if memberName is not present in either Spec or Status.
+func getSanitizedMemberPath(memberName string, r *model.CRD, op *awssdkmodel.Operation, koVarName string) (string, error) {
```
jaypipes, line 192: Nice little cleanup function, thank you. Please remember to wrap comments at 80 chars and also for long (>80 char) function signatures like this, I like to break the function parameters by newline:

```go
func getSanitizedMemberPath(
	memberName string,
	r *model.CRD,
	op *awssdkmodel.Operation,
	koVarName string,
) (string, error) {
...
}
```

jaypipes (changes requested):
Thanks for working on this @brycahta! Great start! A few notes for you inline, but overall really nice. :)

templates/pkg/resource/sdk_find_read_many.go.tpl
brycahta: To confirm: tab `return rm.{{ $customCheckMethod }}(r)` on line 68?

jaypipes: yuppers.

pkg/model/crd.go
```diff
@@ -657,6 +657,45 @@ func (r *CRD) GetAllRenames(op OpType) (map[string]string, error) {
 	return renames, nil
 }
 
+// FindIdentifierInShape returns true and the resource's identifier iff shape
+// contains a matching field or corresponding plural meaning it can be populated
+// using the resource identifier.
+func (r *CRD) FindIdentifierInShape(
```
brycahta, line 663: refactoring this into a common util file instead of crd.go

pkg/generate/code/check.go
```diff
@@ -14,14 +14,16 @@
...
 
-	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
-
-	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/generate/config"
 	"github.com/aws-controllers-k8s/code-generator/pkg/model"
 	"github.com/aws-controllers-k8s/code-generator/pkg/names"
+	"github.com/aws-controllers-k8s/code-generator/pkg/util"
+
+	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/generate/config"
+	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
```
A-Hilaly, line 26: nit: The standard library packages are always in the first group, external modules are in the second one and internal imports are in the third group.

pkg/model/crd.go
```diff
@@ -19,12 +19,12 @@ import (
...
 
-	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
+	"github.com/aws-controllers-k8s/code-generator/pkg/names"
+	"github.com/aws-controllers-k8s/code-generator/pkg/util"
 	"github.com/gertd/go-pluralize"
 
 	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/generate/config"
-	"github.com/aws-controllers-k8s/code-generator/pkg/names"
-	"github.com/aws-controllers-k8s/code-generator/pkg/util"
+	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
```
A-Hilaly, line 27: Ditto

pkg/model/crd.go
A-Hilaly: Nice! Maybe we could also reuse it in here https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/code/set_resource.go#L855 :)

pkg/generate/code/check.go
```diff
@@ -132,37 +136,74 @@ func checkRequiredFieldsMissingFromShape(
...
+		resVarPath = resVarPath + r.Config().PrefixConfig.SpecField + "." + cleanMemberName
+	} else if found {
+		resVarPath = resVarPath + r.Config().PrefixConfig.SpecField + "." + renamedName
+	} else {
+		_, found = r.StatusFields[memberName]
+		if !found {
+			return "", errors.New(fmt.Sprintf("the required field %s is NOT present in CR's Spec or Status", memberName))
+		}
+		resVarPath = resVarPath + r.Config().PrefixConfig.StatusField + "." + cleanMemberName
+	}
```
A-Hilaly, line 207: nit: you could also use switch/case statement in here

pkg/model/crd.go
```diff
@@ -657,6 +657,45 @@ func (r *CRD) GetAllRenames(op OpType) (map[string]string, error) {
...
+	for _, id := range resourceIdentifiers {
+		_, found := r.SpecFields[id]
+		if !found {
+			_, found = r.StatusFields[id]
+		}
+		if found {
+			resIdentifier = id
+			break
+		}
+	}
```
A-Hilaly, line 686: Not a big deal but we could also exit early if !found in `SpecFields` and !found in `StatusFields`

A-Hilaly (approved):
LGTM Overall :+1:
I left few nit suggestions blow

pkg/model/crd.go
```diff
@@ -657,6 +657,45 @@ func (r *CRD) GetAllRenames(op OpType) (map[string]string, error) {
 	return renames, nil
 }
 
+// FindIdentifierInShape returns true and the resource's identifier iff shape
```
A-Hilaly, line 660: Also can we add some unit tests for this function?

pkg/generate/code/check.go
```diff
@@ -132,37 +136,74 @@ func checkRequiredFieldsMissingFromShape(
...
+	)
+	_, found := r.SpecFields[renamedName]
+	if found && !wasRenamed {
+		resVarPath = resVarPath + r.Config().PrefixConfig.SpecField + "." + cleanMemberName
+	} else if found {
+		resVarPath = resVarPath + r.Config().PrefixConfig.SpecField + "." + renamedName
+	} else {
+		_, found = r.StatusFields[memberName]
+		if !found {
+			return "", errors.New(fmt.Sprintf("the required field %s is NOT present in CR's Spec or Status", memberName))
```
A-Hilaly, line 204: s/`errors.New(fmt.Sprintf())`/`fmt.Errorf()`/

pkg/generate/code/check.go
```diff
@@ -132,37 +136,82 @@ func checkRequiredFieldsMissingFromShape(
...
+				}
+				missing = append(missing, fmt.Sprintf("%s == nil", resVarPath))
+			}
+		}
+	}
+
 	// Use '||' because if any of the required fields are missing the object
 	// is not created yet
 	missingCondition := strings.Join(missing, " || ")
 	return fmt.Sprintf("%sreturn %s\n", indent, missingCondition)
```
brycahta, line 185: I think I'll need a check on len(missing) here for `return fmt.Sprintf("%sreturn false", indent)` in the event that neither of these scenarios (ShapeRequired or ReadMany Ids) apply. 🥲

pkg/generate/code/common.go
```diff
@@ -0,0 +1,89 @@
...
+		"Names",
+		r.Names.Original + "Name",
+		r.Names.Original + "Names",
+		r.Names.Original + "Id",
+		r.Names.Original + "Ids",
+	}
+	identifier := ""
+
+	for _, memberName := range shape.MemberNames() {
+		if util.InStrings(memberName, identifierLookup) {
```
brycahta, line 43: GH is rendering extra tabs here for some reason. In my IDE, it's formatted as expected. Will take a look

pkg/generate/code/check.go
```diff
@@ -153,31 +151,24 @@ func checkRequiredFieldsMissingFromShape(
...
-		if strings.Contains(strings.ToLower(field), crKind) {
-			pluralize := pluralize.NewClient()
-			// CR shouldn't have plural of its kind as a field
-			singular := pluralize.Singular(field)
-
-			resVarPath, err := getSanitizedMemberPath(singular, r, op, koVarName)
+	// If ReadMany input shape can take a resource identifier or collection of
+	// identifiers, then generate an if condition checking for resource
+	// identifier field having non-nil corresponding resource Spec/Status values
+	if identifier, found := r.FindIdentifierInShape(shape); found {
```
RedbackThomson: OMG this is so clean! ++

pkg/model/crd.go
```diff
@@ -657,6 +657,45 @@ func (r *CRD) GetAllRenames(op OpType) (map[string]string, error) {
...
+// contains a matching field or corresponding plural meaning it can be populated
+// using the resource identifier.
+func (r *CRD) FindIdentifierInShape(
+	shape *awssdkmodel.Shape) (string, bool) {
+	resIdentifier := ""
+	found := false
+	resourceIdentifiers := []string{
+		"Name",
+		"Names",
```
RedbackThomson: also `Id` and `Ids`

pkg/model/crd.go
```diff
@@ -657,6 +657,45 @@ func (r *CRD) GetAllRenames(op OpType) (map[string]string, error) {
...
 }
 
+// FindIdentifierInShape returns true and the resource's identifier iff shape
+// contains a matching field or corresponding plural meaning it can be populated
+// using the resource identifier.
+func (r *CRD) FindIdentifierInShape(
+	shape *awssdkmodel.Shape) (string, bool) {
+	resIdentifier := ""
+	found := false
+	resourceIdentifiers := []string{
```
RedbackThomson: Should these have capitalised initialisms? i.e. `r.Names.Original + "ID"`

I think our spec and status field keys will have proper casing.

pkg/model/crd.go
```diff
@@ -657,6 +657,45 @@ func (r *CRD) GetAllRenames(op OpType) (map[string]string, error) {
...
+		"Name",
+		"Names",
+		r.Names.Original + "Name",
+		r.Names.Original + "Names",
+		r.Names.Original + "Id",
+		r.Names.Original + "Ids",
+	}
+
+	// Locate an identifier within resource
```
RedbackThomson: Some resources may have multiple identifiers :D 

Example: An [APIGateway API](https://docs.aws.amazon.com/sdk-for-go/api/service/apigatewayv2/#GetApiOutput) has `ApiId` and `Name`. Therefore we will have `Spec.Name` and `Status.APIID`. 

We won't know ahead of time which identifier will be needed in the `ReadMany` filter. Lets create an array of identifiers and then loop over them in L690?

RedbackThomson (commented):
Oh no I found an edge case. Please see inline comments :)

pkg/generate/code/check.go
```diff
@@ -132,37 +136,82 @@ func checkRequiredFieldsMissingFromShape(
...
+		if !util.InStrings(shapeIdentifier, shape.Required) {
+			// both have identifiers, but do they refer to the same thing
+			pluralize := pluralize.NewClient()
+			if strings.EqualFold(pluralize.Singular(crIdentifier),
+				pluralize.Singular(shapeIdentifier)) {
+				resVarPath, err := getSanitizedMemberPath(crIdentifier, r, op, koVarName)
+				if err != nil {
+					// If it isn't in our spec/status fields, we have a problem!
+					msg := fmt.Sprintf(
+						"GENERATION FAILURE! there's a required field %s in "+
```
vijtrip2, line 170: This error message is not correct. In this loop we are processing fields which are not required in sdk shape.

pkg/generate/code/check.go
```diff
@@ -132,37 +136,82 @@ func checkRequiredFieldsMissingFromShape(
...
+					"Shape %s that isn't in either the CR's Spec or "+
+					"Status structs!",
+				memberName, shape.ShapeName,
+			)
+			panic(msg)
 		}
 		missing = append(missing, fmt.Sprintf("%s == nil", resVarPath))
 	}
+
+	// If ReadMany input shape can take a resource identifier or collection of
```
vijtrip2, line 154: Should this code block under condition like if op == r.Ops.ReadMany ??

jaypipes (changes requested):

pkg/generate/code/check.go
brycahta: Yeah, was torn between reducing branched logic (constant work) or explicitly separating the logic out per your suggestions. The latter makes it clearer and is consistent with existing code so will move forward with that.

Will break out the PR 👍

pkg/model/crd.go
brycahta: I don't think capitalized initialisms are needed. Using the existing `CheckRequiredFieldsMissingFromShape` as reference, when the Spec/Status field path is being constructed the lookup in Spec/Status is done using memberName (VpcId) not cleanMemberName (i.e. VPCID). I'll dive deeper to verify though

pkg/model/crd.go
brycahta: Good callout. I'm refactoring this logic into 2 separate helper funcs (will be in latest revision), and I think you're right that I'll need to return a slice of identifiers:

If the scenario happens where the Shape contains `Names` and `Ids` field, but the CRD only contains `Id` field, then returning just the Names and comparing Id would yield no required fields (when really we want to require `Id` because `Ids` is present in Shape).

jaypipes (approved):
Thanks @brycahta :)

PR #155, opened 2021-08-12
Implement ReadMany for CheckRequiredFieldsMissingFromShape

brycahta:
Issue #, if available: [#890](https://github.com/aws-controllers-k8s/community/issues/890)

Description of changes:
* Adds checkRequiredFieldsMissingFromShapeReadMany to `check.go` to handle ReadMany operations without a corresponding ReadOne. 
* The new method will **require resource identifier field** should there exist an *identifier* or *identifierS* field in the ReadMany operation. 
  * Eventually `newListRequestPayload` will use this identifier field to populate its request in order to guarantee the ReadMany operation returns the desired resource.
  * ex: DescribeVpcs has a VpcIds field. Therefore, Vpc shape should require VpcId, which is part of its status field.
* Updated tests and templates

Testing:
* `make test` ✅
* `make build-controller`  ✅

```
func (rm *resourceManager) sdkFind(
	ctx context.Context,
	r *resource,
) (latest *resource, err error) {
	rlog := ackrtlog.FromContext(ctx)
	exit := rlog.Trace("rm.sdkFind")
	defer exit(err)
	// If any required fields in the input shape are missing, AWS resource is
	// not created yet. Return NotFound here to indicate to callers that the
	// resource isn't yet created.
	if rm.requiredFieldsMissingFromReadManyInput(r) {
		return nil, ackerr.NotFound
	}

	input, err := rm.newListRequestPayload(r)

...


// requiredFieldsMissingFromReadManyInput returns true if there are any fields
// for the ReadMany Input shape that are required but not present in the
// resource's Spec or Status
func (rm *resourceManager) requiredFieldsMissingFromReadManyInput(
	r *resource,
) bool {
	return r.ko.Status.VPCID == nil

}
```

### DIFF
--- a/pkg/generate/code/check.go
+++ b/pkg/generate/code/check.go
@@ -196,6 +196,10 @@ func checkRequiredFieldsMissingFromShapeReadMany(
 		for _, ci := range crIdentifiers {
 			if strings.EqualFold(pluralize.Singular(si),
 				pluralize.Singular(ci)) {
+				// The CRD identifiers being used for comparison reflect the
+				// *original* field names in the API model shape.
+				// Field renames are handled below in the call to
+				// getSanitizedMemberPath.
 				if reqIdentifier == "" {
 					reqIdentifier = ci
 				} else {
--- a/pkg/generate/code/check.go
+++ b/pkg/generate/code/check.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 
 	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
+	"github.com/gertd/go-pluralize"
 
 	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/generate/config"
 	"github.com/aws-controllers-k8s/code-generator/pkg/model"
@@ -74,6 +75,10 @@ func CheckRequiredFieldsMissingFromShape(
 	switch opType {
 	case model.OpTypeGet:
 		op = r.Ops.ReadOne
+	case model.OpTypeList:
+		op = r.Ops.ReadMany
+		return checkRequiredFieldsMissingFromShapeReadMany(
+			r, koVarName, indentLevel, op, op.InputRef.Shape)
 	case model.OpTypeGetAttributes:
 		op = r.Ops.GetAttributes
 	case model.OpTypeSetAttributes:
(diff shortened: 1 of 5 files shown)


pkg/generate/code/check_test.go
```diff
@@ -115,3 +115,24 @@ func TestCheckRequiredFields_RenamedSpecField(t *testing.T) {
 		strings.TrimSpace(gotCode),
 	)
 }
+
```
brycahta, line 118: I wanted a test where the Shape/CR have multiple identifiers, but the `Id` gets prioritized. The only one I could think of was *SecurityGroups*. The DescribeInput can take either *Name* or *Id*, and the SecurityGroup CR has both Name in the Spec and Id in the Status. 

In this scenario `Id` should be a required field. However, when trying to test this I uncovered an edge case in the `FindIdentifiers` helper funcs. No identifiers are returned for SecurityGroup CR or the DescribeInput because the fields are not securityGroupName/securityGroupId, but groupName/groupId. 😒

I can address this edge case in a different PR, but if anyone else has a good API to test the above case I'd love to add and test it now!

templates/pkg/resource/sdk_find_read_many.go.tpl
```diff
@@ -51,6 +58,19 @@ func (rm *resourceManager) sdkFind(
...
 }
 
+// requiredFieldsMissingFromReadManyInput returns true if there are any fields
+// for the ReadMany Input shape that are required but not present in the
+// resource's Spec or Status
+func (rm *resourceManager) requiredFieldsMissingFromReadManyInput(
+	r *resource,
+) bool {
+{{- if $customCheckMethod := .CRD.GetCustomCheckRequiredFieldsMissingMethod .CRD.Ops.ReadMany }}
+    return rm.{{ $customCheckMethod }}(r)
```
brycahta, line 68: @jaypipes hope this is properly formatted. I tried your suggestion from last PR, but that adjusted the spacing between the comments within the template as well

pkg/generate/code/check.go
```diff
@@ -152,6 +157,70 @@ func checkRequiredFieldsMissingFromShape(
...
+//
+// As an example, DescribeVpcs EC2 API call doesn't have a ReadOne operation or
+// required fields. However, the input shape has a VpcIds field which can be
+// populated using a VpcId, a field in the VPC CR's Status. Therefore, require
+// the VpcId field to be present to ensure the returned array from the API call
+// consists only of the desired Vpc.
+//
+// Sample Output:
+//
+// return r.ko.Status.VPCID == nil
```
jaypipes, line 176: Great explanation. ++

pkg/generate/code/common.go
```diff
@@ -62,7 +67,6 @@ func FindIdentifiersInCRD(
 		r.Names.Original + "Id",
 		r.Names.Original + "Ids",
```
jaypipes, line 68: I just noticed this... we should move the `r.Names.Original + "Id[s]"` lines above the `"Name"` line.

pkg/generate/code/check.go
```diff
@@ -152,6 +157,70 @@ func checkRequiredFieldsMissingFromShape(
...
+	if len(shapeIdentifiers) == 0 || len(crIdentifiers) == 0 {
+		return result
+	}
+
+	pluralize := pluralize.NewClient()
+	reqIdentifier := ""
+	for _, si := range shapeIdentifiers {
+		for _, ci := range crIdentifiers {
+			if strings.EqualFold(pluralize.Singular(si),
+				pluralize.Singular(ci)) {
```
jaypipes, line 198: I can see someone reading this in the future and saying to themselves "hmmm, what if the field was renamed?" ...

You may want to make a note that this only works because the `SpecFields` and `StatusFields` maps in the CRD are keyed by the *original* member (field) name in the API model shape, and `FindIdentifiersInCRD` returns these map keys, and that we handle field renames below in the call to `getSanitizedMemberPath` :)

pkg/generate/code/common.go
```diff
@@ -47,11 +49,14 @@ func FindIdentifiersInShape(
 	return identifiers
 }
 
-// FindIdentifiersInCRD returns the identifier field of a given CRD which
-// can be singular or plural. Errors iff multiple identifier fields detected
-// in the CRD.
+// FindIdentifiersInCRD returns the identifier fields of a given CRD which
+// can be singular or plural.
```
jaypipes, line 53: I think it's worth noting in this docstring that this function returns the *original* member (field) name from the API model shape, not the renamed field name.

templates/pkg/resource/sdk_find_read_many.go.tpl
jaypipes: It's OK :) We run `go fmt` against the outputted code anyway, so even though no, it's still using spaces, it will get corrected ex post facto...

jaypipes (changes requested):
@brycahta this is almost there. I have a request for a couple clarifying code comments to help future readers of this code inline.

pkg/generate/code/check.go
brycahta: Instead of needing the comment, it might be cleaner to just use `getSanitizedMemberPath` here. Something like:

```
	for _, si := range shapeIdentifiers {
		if pluralize.IsPlural(si) {
			flipped = pluralize.Singular(si)
		} else {
			flipped = pluralize.Plural(si)
		}

		// CRD shouldn't contain both singular and plural
		resVarPath, err = getSanitizedMemberPath(si, r, op, koVarName)
		if err != nil {
			resVarPath, err = getSanitizedMemberPath(flipped, r, op, koVarName)
			if err != nil {
				continue
			}
		}

		// If there are multiple identifiers, then prioritize the
		// 'Id' identifier. Checking 'Id' to determine resource
		// creation should be safe as the field is usually
		// present in CR.Status.
		if strings.HasSuffix(resVarPath, "Id") ||
			strings.HasSuffix(resVarPath, "Ids") {
			break
		}
	}
```

what do you think?

jaypipes: No, I think the code you have here already is good. Just that a code comment would be useful :)

pkg/generate/code/check.go
```diff
@@ -152,6 +157,74 @@ func checkRequiredFieldsMissingFromShape(
...
+				// The CRD identifiers being used for comparison reflect the
+				// *original* field names in the API model shape.
+				// Field renames are handled below in the call to
+				// getSanitizedMemberPath.
```
jaypipes, line 202: :+1:

jaypipes (approved):
Great work on this @brycahta, thank you!

PR #156, opened 2021-08-16
Implement ReadMany operation for SetSDK generator function

brycahta:
Issue #, if available: [#891](https://github.com/aws-controllers-k8s/community/issues/891)

Description of changes:
* adds `setSDKReadMany` func to SetSDK
* adds unit test

Testing:
* `make test` ✅
* `make build-controller`  ✅

```
// newListRequestPayload returns SDK-specific struct for the HTTP request
// payload of the List API call for the resource
func (rm *resourceManager) newListRequestPayload(
	r *resource,
) (*svcsdk.DescribeVpcsInput, error) {
	res := &svcsdk.DescribeVpcsInput{}

	if r.ko.Spec.DryRun != nil {
		res.SetDryRun(*r.ko.Spec.DryRun)
	}
	if r.ko.Status.VPCID != nil {
		f4 := []*string{}
		f4 = append(f4, r.ko.Status.VPCID)
		res.SetVpcIds(f4)
	}

	return res, nil
}
```

### DIFF
--- a/pkg/generate/code/set_sdk.go
+++ b/pkg/generate/code/set_sdk.go
@@ -735,6 +735,14 @@ func SetSDKSetAttributes(
 // a VpcId, a field in the VPC CR's Status. Therefore, populate VpcIds field
 // with the *single* VpcId value to ensure the returned array from the API call
 // consists only of the desired Vpc.
+//
+// Sample Output:
+//
+// 	if r.ko.Status.VPCID != nil {
+//		f4 := []*string{}
+//		f4 = append(f4, r.ko.Status.VPCID)
+//		res.SetVpcIds(f4)
+//	}
 func setSDKReadMany(
 	cfg *ackgenconfig.Config,
 	r *model.CRD,
@@ -767,7 +775,8 @@ func setSDKReadMany(
 				case "string":
 					value = "\"" + value + "\""
 				default:
-					panic("Member type not handled")
+					panic(fmt.Sprintf("Unsupported shape type %s in "+
+						"generate.code.setSDKReadMany", memberShape.Type))
 				}
 
 				out += fmt.Sprintf("%s%s.Set%s(%s)\n", indent, targetVarName, memberName, value)
@@ -791,6 +800,11 @@ func setSDKReadMany(
 				if resVarPath == "" || (!strings.HasSuffix(resVarPath, "Id") ||
 					!strings.HasSuffix(resVarPath, "Ids")) {
 					resVarPath, err = getSanitizedMemberPath(flipped, r, op, sourceVarName)
+					if err != nil {
+						panic(fmt.Sprintf(
+							"Unable to locate identifier field %s in "+
+								"%s Spec/Status in generate.code.setSDKReadMany", flipped, r.Kind))
+					}
 				}
 			} else {
 				// TODO(jaypipes): check generator config for exceptions?
@@ -807,34 +821,32 @@ func setSDKReadMany(
 		switch memberShape.Type {
 		case "list":
 			// Expecting slice of identifiers
-			{
-				memberVarName := fmt.Sprintf("f%d", memberIndex)
-				// f0 := []*string{}
-				out += varEmptyConstructorSDKType(
-					cfg, r,
-					memberVarName,
-					memberShape,
-					indentLevel+1,
-				)
+			memberVarName := fmt.Sprintf("f%d", memberIndex)
+			// f0 := []*string{}
+			out += varEmptyConstructorSDKType(
+				cfg, r,
+				memberVarName,
+				memberShape,
+				indentLevel+1,
+			)
 
-				//  f0 = append(f0, sourceVarName)
-				out += fmt.Sprintf("%s\t%s = append(%s, %s)\n", indent,
-					memberVarName, memberVarName, resVarPath)
+			//  f0 = append(f0, sourceVarName)
+			out += fmt.Sprintf("%s\t%s = append(%s, %s)\n", indent,
+				memberVarName, memberVarName, resVarPath)
 
-				// res.SetIds(f0)
-				out += setSDKForScalar(
-					cfg, r,
-					memberName,
-					targetVarName,
-					inputShape.Type,
-					sourceVarName,
-					memberVarName,
-					memberShapeRef,
-					indentLevel+1,
-				)
-			}
+			// res.SetIds(f0)
+			out += setSDKForScalar(
+				cfg, r,
+				memberName,
+				targetVarName,
+				inputShape.Type,
+				sourceVarName,
+				memberVarName,
+				memberShapeRef,
+				indentLevel+1,
+			)
 		default:
-			// For rare ReadMany that have a singular identifier field.
+			// For ReadMany that have a singular identifier field.
 			// ex: DescribeReplicationGroups
 			out += setSDKForScalar(
 				cfg, r,
--- a/pkg/generate/code/set_sdk.go
+++ b/pkg/generate/code/set_sdk.go
@@ -1106,7 +1106,7 @@ func SetSDKForStruct(
 }
 
 // setSDKForSlice returns a string of Go code that sets a target variable value
-// to a source variable when the type of the source variable is a struct.
+// to a source variable when the type of the source variable is a slice.
 func setSDKForSlice(
 	cfg *ackgenconfig.Config,
 	r *model.CRD,
@@ -1171,7 +1171,7 @@ func setSDKForSlice(
 }
 
 // setSDKForMap returns a string of Go code that sets a target variable value
-// to a source variable when the type of the source variable is a struct.
+// to a source variable when the type of the source variable is a map.
 func setSDKForMap(
 	cfg *ackgenconfig.Config,
 	r *model.CRD,
--- a/pkg/generate/code/set_sdk.go
+++ b/pkg/generate/code/set_sdk.go
@@ -19,10 +19,12 @@ import (
 	"strings"
 
 	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
+	"github.com/gertd/go-pluralize"
 
 	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/generate/config"
 	"github.com/aws-controllers-k8s/code-generator/pkg/model"
 	"github.com/aws-controllers-k8s/code-generator/pkg/names"
+	"github.com/aws-controllers-k8s/code-generator/pkg/util"
 )
 
 // SetSDK returns the Go code that sets an SDK input shape's member fields from
@@ -98,6 +100,8 @@ func SetSDK(
 		op = r.Ops.ReadOne
 	case model.OpTypeList:
 		op = r.Ops.ReadMany
+		return setSDKReadMany(cfg, r, op,
+			sourceVarName, targetVarName, indentLevel)
 	case model.OpTypeUpdate:
 		op = r.Ops.Update
 	case model.OpTypeDelete:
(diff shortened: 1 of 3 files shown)


brycahta:
### Refactor Opportunity 
* There's duplicate logic between this ReadMany impl and the one done for CheckRequiredFields in this [pr](https://github.com/aws-controllers-k8s/code-generator/pull/155).
* I'd like to encapsulate this locating-and-matching-identifier logic as `/pkg/generate/code/identifier.go` ([common.go](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/code/common.go) will be deleted after moving helpers to this new file.)
* This set of helpers will not only help locate identifiers, but also match identifier fields between API Input Shape and a CRD model.
* **Are there other areas in the project that reconcile complex data (i.e. comparing names and plurals) between API Input Shape and CRD?**
 * If so, please post details and we can see if a higher level abstraction would be more appropriate

vijtrip2:
> * I'd like to encapsulate this locating-and-matching-identifier logic as `/pkg/generate/code/identifier.go` ([common.go](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/code/common.go) will be deleted after moving helpers to this new file.)

+1 . The less places we have the duplicated code the better because if a bug fix is needed in future, duplicated code may cause some places to be missed with the fix.

pkg/generate/code/set_sdk_test.go
```diff
@@ -1778,3 +1778,28 @@ func TestSetSDK_MQ_Broker_Create(t *testing.T) {
...
+	assert := assert.New(t)
+	require := require.New(t)
+
+	g := testutil.NewModelForService(t, "ec2")
+
+	crd := testutil.GetCRDByName(t, g, "Vpc")
+	require.NotNil(crd)
+
+	expected := `
+	if r.ko.Spec.DryRun != nil {
```
vijtrip2, line 1792: nit: wasn't dry-run supposed to be excluded from the VPC CR? It may just be unit tests but better to keep it consistent with actual controller. : )

vijtrip2 (commented):
Looks good to me. Just one very minor nit about unit test data.

pkg/generate/code/set_sdk_test.go
brycahta: That's correct. Does that mean service controller's _generator.yaml_ needs to stay in sync with https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/testdata/models/apis/ec2/0000-00-00/generator.yaml ?

vijtrip2: not necessarily that's why I mentioned nit.
I mentioned this comment just because why unit test something(dry-run here) which is never actually used in reality.

pkg/generate/code/set_sdk.go
```diff
@@ -718,6 +722,139 @@ func SetSDKSetAttributes(
...
+// fields-- specifically identifier(s). This method populates this identifier
+// field with the identifier shared between the shape and the CR. Note, in the
+// case of multiple matching identifiers, the identifier field containing 'Id'
+// will be the only field populated.
+//
+// As an example, DescribeVpcs EC2 API call doesn't have a ReadOne operation or
+// required fields. However, the input shape VpcIds field can be populated using
+// a VpcId, a field in the VPC CR's Status. Therefore, populate VpcIds field
+// with the *single* VpcId value to ensure the returned array from the API call
+// consists only of the desired Vpc.
```
jaypipes, line 737: If you could, please place example Go code that is output from this function into the docstring here. You can simply copy the output Go code from your unit test if you want.

pkg/generate/code/set_sdk.go
```diff
@@ -718,6 +722,139 @@ func SetSDKSetAttributes(
...
+					flipped = pluralize.Plural(memberName)
+				}
+				// If there are multiple identifiers, then prioritize the
+				// 'Id' identifier.
+				if resVarPath == "" || (!strings.HasSuffix(resVarPath, "Id") ||
+					!strings.HasSuffix(resVarPath, "Ids")) {
+					resVarPath, err = getSanitizedMemberPath(flipped, r, op, sourceVarName)
+				}
+			} else {
+				// TODO(jaypipes): check generator config for exceptions?
```
jaypipes, line 796: Leaving me TODOs, huh? ;) Sneaky.

pkg/generate/code/set_sdk.go
```diff
@@ -718,6 +722,139 @@ func SetSDKSetAttributes(
...
+				}
+
+				out += fmt.Sprintf("%s%s.Set%s(%s)\n", indent, targetVarName, memberName, value)
+				continue
+			}
+		}
+
+		// Field renames are handled in getSanitizedMemberPath
+		resVarPath, err = getSanitizedMemberPath(memberName, r, op, sourceVarName)
+		if err != nil {
```
jaypipes, line 780: OK, so `err` will be non-`nil` if `memberName` is not found in either Spec or Status of the CR, And so you check below in lines 782-794 whether there is a field in Spec or Status that is the plural (or singular) of the `memberName` (if the `memberName` is also an identifier field).

That is all good, however on line 793, you aren't handling a non-`nil` `err` returned from looking up the sanitized member path of the "flipped" memberName. Which means we're kind of flying blind in the case of odd Shapes in APIs.

I think it would be safest to `panic` after receiving a non-`nil` `err` on line 793. What do you think?

pkg/generate/code/set_sdk.go
```diff
@@ -718,6 +722,139 @@ func SetSDKSetAttributes(
...
+					inputShape.Type,
+					sourceVarName,
+					memberVarName,
+					memberShapeRef,
+					indentLevel+1,
+				)
+			}
+		default:
+			// For rare ReadMany that have a singular identifier field.
+			// ex: DescribeReplicationGroups
```
jaypipes, line 838: Not necessarily rare. :) Pretty much all of the RDS ReadMany APIs are like this.

Por ejemplo:

https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBParameterGroups.html
https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBParameters.html
https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBProxies.html
https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBProxyEndpoints.html

That last one is interesting because it has *two* scalar-typed filter fields :)

Also, there may be scalar-typed fields in a ReadMany input shape that act as filters on, say, boolean fields.

For example:

https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeOrderableDBInstanceOptions.html

Check out the `Vpc` member of the input shape. It's a boolean. Yep. This is why we can't have nice things.

jaypipes (changes requested):
I'm impressed with you digging into the code-generator like this, @brycahta. Well done. :)

I've left a couple comments inline for you to ponder.

pkg/generate/code/set_sdk.go
```diff
@@ -718,6 +722,139 @@ func SetSDKSetAttributes(
...
+			value, ok := opConfig[memberName]
+			memberShapeRef, _ := inputShape.MemberRefs[memberName]
+			memberShape := memberShapeRef.Shape
+			if ok {
+				switch memberShape.Type {
+				case "boolean", "integer":
+				case "string":
+					value = "\"" + value + "\""
+				default:
+					panic("Member type not handled")
```
A-Hilaly, line 770:
```suggestion
					panic(fmt.Sprintf("Unsupported shape type %s in generate.code.setSDKReadMany", memberShape.Type))
```

pkg/generate/code/set_sdk.go
brycahta: Good catch-- my only reservation with the `panic` is **if** the following scenario is possible:

* Shape has `Names` and `Identifiers` field
* CRD model only has `Identifier` field
* We hit `Names` Shape member in this loop, flip it, and try to set `resVarPath` since it's empty string, but then we panic and exit since `Name` isn't in CRD
* One more iteration we would've hit `Identifiers` and properly mapped the field.

Have you come across any APIs like this?

pkg/generate/code/set_sdk.go
```diff
@@ -718,6 +722,139 @@ func SetSDKSetAttributes(
...
+	out := "\n"
+	indent := strings.Repeat("\t", indentLevel)
+
+	resVarPath := ""
+	pluralize := pluralize.NewClient()
+	opConfig, override := cfg.OverrideValues(op.Name)
+	shapeIdentifiers := FindIdentifiersInShape(r, inputShape)
+	var err error
+	for memberIndex, memberName := range inputShape.MemberNames() {
+		if override {
```
A-Hilaly, line 760: nit:
```suggestion
      if override {
	      for memberIndex, memberName := range inputShape.MemberNames() {
```

pkg/generate/code/set_sdk.go
```diff
@@ -718,6 +722,139 @@ func SetSDKSetAttributes(
...
+				if pluralize.IsPlural(memberName) {
+					flipped = pluralize.Singular(memberName)
+				} else {
+					flipped = pluralize.Plural(memberName)
+				}
+				// If there are multiple identifiers, then prioritize the
+				// 'Id' identifier.
+				if resVarPath == "" || (!strings.HasSuffix(resVarPath, "Id") ||
+					!strings.HasSuffix(resVarPath, "Ids")) {
+					resVarPath, err = getSanitizedMemberPath(flipped, r, op, sourceVarName)
```
A-Hilaly, line 793: This error needs to be handled, probably a panic?

pkg/generate/code/set_sdk.go
```diff
@@ -718,6 +722,139 @@ func SetSDKSetAttributes(
...
+		memberShapeRef, _ := inputShape.MemberRefs[memberName]
+		memberShape := memberShapeRef.Shape
+		out += fmt.Sprintf(
+			"%sif %s != nil {\n", indent, resVarPath,
+		)
+
+		switch memberShape.Type {
+		case "list":
+			// Expecting slice of identifiers
+			{
```
A-Hilaly, line 810: nit: I believe that creating a new scope is not necessary here

pkg/generate/code/set_sdk.go
brycahta: Woww thanks for these; I'll update the comment. Just when you think you were starting to get the hang of the API...

pkg/generate/code/set_sdk.go
```diff
@@ -969,7 +1106,7 @@ func SetSDKForStruct(
 }
 
 // setSDKForSlice returns a string of Go code that sets a target variable value
-// to a source variable when the type of the source variable is a struct.
+// to a source variable when the type of the source variable is a slice.
```
A-Hilaly, line 1109: Yikes nice catch!

pkg/generate/code/set_sdk.go
brycahta: Haha figured both of these TODOs could be addressed at the same time :D

pkg/generate/code/set_sdk_test.go
```diff
@@ -1778,3 +1778,28 @@ func TestSetSDK_MQ_Broker_Create(t *testing.T) {
...
+		res.SetDryRun(*r.ko.Spec.DryRun)
+	}
+	if r.ko.Status.VPCID != nil {
+		f4 := []*string{}
+		f4 = append(f4, r.ko.Status.VPCID)
+		res.SetVpcIds(f4)
+	}
```
A-Hilaly, line 1799: This is 100% correct. Just pointing that If an array/map is made of Go built-in type we could just do:

```go
res.SetVpcIds(r.ko.Status.VPCID)
```

A-Hilaly (commented):
Awesome stuff! :+1:

I Ieft few comments below

pkg/generate/code/set_sdk_test.go
brycahta: Ah good point. If this is preferred by the team, then I could just call `setSDKForScalar` instead of the emptyVar call, then the append, then the `setSDKForScalar`. 

**Any opinions/preferences from the team?**

pkg/generate/code/set_sdk.go
jaypipes: > Good catch-- my only reservation with the `panic` is **if** the following scenario is possible:
> 
> * Shape has `Names` and `Identifiers` field
> * CRD model only has `Identifier` field
> * We hit `Names` Shape member in this loop, flip it, and try to set `resVarPath` since it's empty string, but then we panic and exit since `Name` isn't in CRD
> * One more iteration we would've hit `Identifiers` and properly mapped the field.
> 
> Have you come across any APIs like this?

No, but I am no longer surprised by any AWS API :) Just being safe here.

pkg/generate/code/set_sdk.go
jaypipes: No, we cannot move the `if override` outside of the for loop because we want to loop over all member names in the shape, even when there is no overridden input field values.

pkg/generate/code/set_sdk.go
jaypipes: I think Bryan is just copying code from other places in this file. :)

pkg/generate/code/set_sdk_test.go
jaypipes: I prefer to keep it the way it is, since it is the way it is due to the need to construct both scalar and complex Go types for nested fields. It's a little verbose, but it's consistent for all types, not just builtin types.

pkg/generate/code/set_sdk.go
brycahta: ^^ I stand on the shoulder's of giants

brycahta:
> > * I'd like to encapsulate this locating-and-matching-identifier logic as `/pkg/generate/code/identifier.go` ([common.go](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/code/common.go) will be deleted after moving helpers to this new file.)
> 
> +1 . The less places we have the duplicated code the better because if a bug fix is needed in future, duplicated code may cause some places to be missed with the fix.

@vijtrip2 agreed. I will submit a separate refactor PR following the merge of this one

vijtrip2 (approved):
Looks great!

jaypipes (approved):
Very nice work on this, @brycahta, thank you sir!

PR #157, opened 2021-08-16
use reflect to compare map/slice struct

jaypipes:
Although it's not what I wanted and isn't the most efficient thing we
could do, I'm implementing the slice of struct and map of struct
comparisons using `reflect.DeepEqual`. Long-term, I'd like to enable our
generator-config CompareConfig for nested struct fields, and to do that
we cannot use `reflect.DeepEqual`, however for the time-being, since
slice and map of struct is a hole in our delta logic, I'm pushing up
this workaround.

### DIFF
--- a/pkg/generate/code/compare.go
+++ b/pkg/generate/code/compare.go
@@ -36,6 +36,44 @@ import (
 // constructed from the Read operation, is compared to the Resource we got from
 // the Kubernetes API server's event bus. The code that is returned from this
 // function is the code that compares those two Resources.
+//
+// The Go code we return depends on the Go type of the various fields for the
+// resource being compared.
+//
+// For *scalar* Go types, the output Go code looks like this:
+//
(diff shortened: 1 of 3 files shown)


vijtrip2 (approved):

RedbackThomson:
Worth noting that `DeepEqual` will return false if the slices are ordered differently. This means our systems require determinism in all slices, and we would expect the same from the services.

I'm happy with this call for now. Functionality >> Performance with delta right now.

/approve

A-Hilaly (approved):
/lgtm

jaypipes:
> Worth noting that `DeepEqual` will return false if the slices are ordered differently. This means our systems require determinism in all slices, and we would expect the same from the services.

Yeah, exactly. And that was a bridge too far for me to implement at this juncture... thus, this hack :)

PR #162, opened 2021-08-19
Support default non-release version in build-controller-release.sh

vijtrip2:
Description of changes:
* With the latest addition of presubmit 'release-test' job, service teams are required to execute 'build-controller-release.sh ' script for a successful PR build. 'build-controller-release.sh' script takes VERSION as required parameter.
* Sagemaker team provided the feedback on how they do not want to worry about controller versions until they are doing an actual release. which is fair. 
* So this code change, introduces a default version 'v0.0.0-non-release-version' which allows for successful release-test for service teams without worrying about what VERSION to use until actual release.
* This code change also adds the validation for the `VERSION` parameter.

Why not use something like `git describe --tags --always --dirty` to generate version ?
* The git command works in most cases if the repository already has a legit semver tag. If that is not the case then we need to add additional handling to create an actual semver tag.
* Instead of dealing with all the edge cases with above command, having a static non-release version gets the job done. Because it is really just a placeholder value for the release-test execution.

### DIFF
--- a/scripts/build-controller-release.sh
+++ b/scripts/build-controller-release.sh
@@ -44,13 +44,13 @@ RUNTIME_API_VERSION=${RUNTIME_API_VERSION:-"v1alpha1"}
 
 USAGE="
 Usage:
-  $(basename "$0") <service> <release_version>
+  $(basename "$0") <service> [<release_version>]
 
 <service> should be an AWS service API aliases that you wish to build -- e.g.
 's3' 'sns' or 'sqs'
 
 <release_version> should be the SemVer version tag for the release -- e.g.
-'v0.1.3'
+'v0.1.3' . If no release_version is specified 'v0.0.0-non-release-version' is used.
 
 Environment variables:
   ACK_GENERATE_CACHE_DIR                Overrides the directory used for caching
@@ -91,9 +91,14 @@ Environment variables:
                                         custom resource definitions.
                                         Default: $K8S_RBAC_ROLE_NAME
 "
+if [ $# -lt 1 ]; then
+    echo "ERROR: $(basename "$0") needs one required parameter, the SERVICE" 1>&2
+    echo "$USAGE"
+    exit 1
+fi
 
-if [ $# -ne 2 ]; then
-    echo "ERROR: $(basename "$0") accepts exactly two parameters, the SERVICE and the RELEASE_VERSION" 1>&2
+if [ $# -gt 2 ]; then
+    echo "ERROR: $(basename "$0") accepts atmost two parameters, the SERVICE(required) and the RELEASE_VERSION(optional)" 1>&2
     echo "$USAGE"
     exit 1
 fi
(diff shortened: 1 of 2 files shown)


scripts/build-controller-release.sh
```diff
@@ -91,9 +91,14 @@ Environment variables:
...
+if [ $# -lt 1 ]; then
+    echo "ERROR: $(basename "$0") needs one required parameter, the SERVICE" 1>&2
+    echo "$USAGE"
+    exit 1
+fi
 
-if [ $# -ne 2 ]; then
-    echo "ERROR: $(basename "$0") accepts exactly two parameters, the SERVICE and the RELEASE_VERSION" 1>&2
+if [ $# -gt 2 ]; then
+    echo "ERROR: $(basename "$0") accepts atmost two parameters, the SERVICE(required) and the RELEASE_VERSION(optional)" 1>&2
```
jaypipes, line 101: Can we revert this change and continue using two required parameters please? We can pass `v0.0.0-non-release-version` string from wherever we call this script (in other words, the prow job...)

scripts/build-controller-release.sh
```diff
@@ -131,7 +136,28 @@ if [[ ! -d $SERVICE_CONTROLLER_SOURCE_PATH ]]; then
...
+  # executing presubmit 'release-test' job on those artifacts.
+  # ACK postsubmit release job makes sure this version does not get released to
+  # public ecr repository.
+  #
+  # Using a static non-release version works because this is only a placeholder
+  # value which gets replaced during presubmit 'release-test' job. Having a
+  # default non-release value also helps AWS service teams to develop the
+  # controller without worrying about the version until actual controller
+  # release.
+  RELEASE_VERSION="v0.0.0-non-release-version"
```
jaypipes, line 159: We can put the above into the prow job that executes `build-controller-release.sh`, no?

vijtrip2: None of the presubmit prowjobs execute this script.

Currently all presubmit tests run on artifacts already present in the PR. They are not generating it for service teams.

scripts/build-controller-release.sh
vijtrip2: > We can pass v0.0.0-non-release-version string from wherever we call this script (in other words, the prow job...)

This is currently invoked manually by service teams. Hence I was adding the default static value on their request.

jaypipes: Why is this being manually invoked by service teams?

Generally, for optional parameters to our Bash scripts, we use environment variables, not arguments...

vijtrip2: > "Why is this being manually invoked by service teams?"

Isn't this same as service teams running `make build-controller` manually. Maybe I am missing something.
@jaypipes , Can you please elaborate on why the prowjob should be executing the build-controller-release.sh script?
(IMO, prowjobs should be testing whatever will be merged into the main branch.)

@RedbackThomson suggested to run build-controller-release.sh script as part of `make build-controller`. I think that could work here. wdyt?


> "Generally, for optional parameters to our Bash scripts, we use environment variables, not arguments..."

Sounds good. I will update this.

RedbackThomson: I also agree that the `build-controller-release.sh` logic should remain the same, as it is required that we pass a version. When being called by developers, perhaps the `build-controller` script can pass this `v0.0.0-non-release-version`?

jaypipes (approved):
Sweet, thanks @vijtrip2!

scripts/build-controller-release.sh
vijtrip2: @RedbackThomson , made the change to execute both scripts from make build-controller target.

PR #163, opened 2021-08-20
Allow override resource identifier to ARN

RedbackThomson:
Closes https://github.com/aws-controllers-k8s/community/issues/909

Description of changes:
The SageMaker `ModelPackage` (in some cases) needs to be described using the resource ARN, even though the spec has a `ModelPackageName` and the `DescribeModelPackage` input shape has a `ModelPackageName` field. This pull requests allows the service teams to set the `primary_identifier_field_name` as `"ARN"`, to override the default generator logic in order to set the `ACKResourceMetadata.ARN` field to be `identifier.ARN`. 

Also refactored the identifier code to use the new `FindIdentifiersInShape` common method and to use the `setResourceForScalar` method.

### DIFF
--- a/pkg/generate/code/set_resource.go
+++ b/pkg/generate/code/set_resource.go
@@ -845,22 +845,14 @@ func SetResourceIdentifiers(
 	primaryKeyConditionalOut += fmt.Sprintf("%s\treturn ackerrors.MissingNameIdentifier\n", indent)
 	primaryKeyConditionalOut += fmt.Sprintf("%s}\n", indent)
 
-	arnOut := ""
+	arnOut := "\n"
 	// if r.ko.Status.ACKResourceMetadata == nil {
 	//  r.ko.Status.ACKResourceMetadata = &ackv1alpha1.ResourceMetadata{}
 	// }
 	// r.ko.Status.ACKResourceMetadata.ARN = identifier.ARN
+	arnOut += ackResourceMetadataGuardConstructor(fmt.Sprintf("%s.Status", targetVarName), indentLevel)
 	arnOut += fmt.Sprintf(
-		"\n%sif %s.Status.ACKResourceMetadata == nil {",
-		indent, targetVarName,
-	)
-	arnOut += fmt.Sprintf(
-		"\n\t%s%s.Status.ACKResourceMetadata = &ackv1alpha1.ResourceMetadata{}",
-		indent, targetVarName,
-	)
-	arnOut += fmt.Sprintf("\n%s}", indent)
-	arnOut += fmt.Sprintf(
-		"\n%s%s.Status.ACKResourceMetadata.ARN = %s.ARN\n",
+		"%s%s.Status.ACKResourceMetadata.ARN = %s.ARN\n",
 		indent, targetVarName, sourceVarName,
 	)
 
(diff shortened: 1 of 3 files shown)


pkg/generate/code/set_resource.go
```diff
@@ -955,18 +983,33 @@ func SetResourceIdentifiers(
...
 		}
 	}
 
 	// Only use at most one of ARN or nameOrID as primary identifier outputs
-	if arnOut != "" {
+	if primaryIdentifier == "ARN" || primaryKeyOut == "" {
 		return arnOut + additionalKeyOut
 	}
-	return primaryKeyOut + additionalKeyOut
+	return primaryKeyConditionalOut + primaryKeyOut + additionalKeyOut
```
jaypipes, line 1012: These code changes are perfectly fine. I do think you might want to break the function into a couple smaller helper functions in a future PR, though. The function is still pretty longish. :)

pkg/generate/code/set_resource_test.go
```diff
@@ -3092,3 +3092,29 @@ func TestSetResource_APIGWV2_ApiMapping_SetResourceIdentifiers(t *testing.T) {
...
+	f0, f0ok := identifier.AdditionalKeys["modelPackageName"]
+	if f0ok {
+		r.ko.Spec.ModelPackageName = &f0
+	}
+`
+	assert.Equal(
+		expected,
+		code.SetResourceIdentifiers(crd.Config(), crd, "identifier", "r.ko", 1),
+	)
+}
```
jaypipes, line 3120: :+1:

jaypipes (approved):
Nice work on this @RedbackThomson. Do we have a GH issue associated with this, btw? If not, feel free to create one and pop it into the current sprint.

pkg/generate/code/set_resource.go
```diff
@@ -908,40 +922,54 @@ func SetResourceIdentifiers(
...
 		case inStatus:
 			memberPath = cfg.PrefixConfig.StatusField
+			targetField = statusField
 		case isPrimaryIdentifier:
 			panic("Primary identifier field '" + memberName + "' in operation '" + op.Name + "' cannot be found in either spec or status.")
 		default:
 			continue
 		}
 
+		targetVarPath := fmt.Sprintf("%s%s", targetVarName, memberPath)
```
brycahta, line 954: Could use [getSanitizedMemberPath](https://github.com/aws-controllers-k8s/code-generator/blob/93f829bb5b5420a20cb1af70a0c562eb61402f2b/pkg/generate/code/check.go#L231) here

pkg/generate/code/set_resource.go
```diff
@@ -852,33 +879,20 @@ func SetResourceIdentifiers(
...
-						"`primary_identifier_field_name` for the " + op.Name +
-						" operation in the generator config.")
-				}
-			}
-		}
+		identifiers := FindIdentifiersInShape(r, inputShape)
 
-		// Still haven't determined the identifier? Panic
-		if primaryIdentifier == "" {
+		switch len(identifiers) {
```
brycahta, line 884: v clean 👍

pkg/generate/code/set_resource.go
RedbackThomson: Yeah this method has grown and grown over time. Will break it out when I implement https://github.com/aws-controllers-k8s/community/issues/874

pkg/generate/code/set_resource.go
RedbackThomson: Hmm, I think there would still be roughly the same number of LoC even if I use `getSanitizedMemberPath`, here. Mainly because I need a reference to the `*Field` - as its required for the `setResourceForScalar` parameters. Although I'm essentially doing the same logic, just with an extra return value. 

I might refactor `getSanitizedMemberPath` to return the field as well, but not in this PR.

pkg/generate/code/set_resource.go
RedbackThomson: So clean. I love this!

PR #165, opened 2021-08-23
add ami to initialisms

brycahta:
Issue #, if available: N/A

Description of changes:
* Add `ami` entry to initialisms with test

### DIFF
--- a/pkg/names/names.go
+++ b/pkg/names/names.go
@@ -68,6 +68,7 @@ var (
 		// Prevent "MultipartUpload" from becoming "MultIPartUpload"
 		{"Ip", "IP", "ip", regexp.MustCompile("Ip(?!art)", regexp.None)},
 		// Easy find-and-replacements...
+		{"Ami", "AMI", "ami", nil},
 		{"Acl", "ACL", "acl", nil},
 		{"Acp", "ACP", "acp", nil},
 		{"Api", "API", "api", nil},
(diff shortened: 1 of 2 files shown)


RedbackThomson (approved):
/lgtm

pkg/names/names.go
```diff
@@ -68,6 +68,7 @@ var (
 		// Prevent "MultipartUpload" from becoming "MultIPartUpload"
 		{"Ip", "IP", "ip", regexp.MustCompile("Ip(?!art)", regexp.None)},
 		// Easy find-and-replacements...
+		{"Ami", "AMI", "ami", nil},
```
jaypipes, line 71: @brycahta @RedbackThomson I think this is going to have issues with "Family"... which this is going to turn into "FAMIly"

A-Hilaly: We can use something similar to what we used for URI/SecURIty: `{"Uri", "URI", "uri", regexp.MustCompile("(?!sec)uri(?!ty)|(Uri)", regexp.None)},`

brycahta: Good points, I was surprised no tests were failing tbh. Let me submit a revision to handle "Family"

brycahta: https://github.com/aws-controllers-k8s/code-generator/pull/167

PR #166, opened 2021-08-24
Allow OperatorType to be a list such that one API can map to multiple operations

mbaijal:
Allow OperatorType to be a list such that one API can map to multiple operations. 
This is required in certain cases where the same API call maps to both Create and Update operations such as `RegisterScalableTarget` and `PutScalingPolicy`

[Link to ApplicationAutoscaling PR with the related change](https://github.com/aws-controllers-k8s/applicationautoscaling-controller/pull/27)
Note: This change will require all other services to update their generator configs as well. 

Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/867

### Description of changes:
The `getOpTypeAndResourceName` method handles each of the following cases same as before in addition to the new case 5- 
1.  the cfg itself is nil
2.  cfg is not nil but opId is not in the cfg
3.  cfg is not nil, the opID is in the cfg but the opType is not specified
  ```
  operations:
    DescribeScalableTargets:
    primary_identifier_field_name: ResourceID
  ```
4. One API -> one operations
```
operations:
  PutScalingPolicy:
    operation_type: 
    - Create
  ```
5. One API -> multiple operations
```
operations:
  PutScalingPolicy:
    operation_type: 
    - Create
    - Update
  ```
### Testing
- I tested by generating the controller code for both applicationAutoscaling and the SageMaker repos. 
- The unit tests pass as is. 

### DIFF
--- a/pkg/generate/config/operation.go
+++ b/pkg/generate/config/operation.go
@@ -14,10 +14,16 @@
 package config
 
 import (
+	"encoding/json"
+
 	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
 
-	"github.com/aws-controllers-k8s/code-generator/pkg/util"
+	"github.com/aws-controllers-k8s/code-generator/pkg/util"	
 )
+
+// StringArray is a type that can be represented in JSON as *either* a string
+// *or* an array of strings
+type StringArray []string
 
 // OperationConfig represents instructions to the ACK code generator to
 // specify the overriding values for API operation parameters and its custom implementation.
@@ -37,7 +43,8 @@ type OperationConfig struct {
 	ResourceName string `json:"resource_name"`
 	// Override for operation type in case of heuristic failure
 	// An example of this is `Put...` or `Register...` API operations not being correctly classified as `Create` op type
-	OperationType string `json:"operation_type"`
+	// OperationType []string `json:"operation_type"`
+	OperationType StringArray `json:"operation_type"`
 	// PrimaryIdentifierFieldName provides the name of the field that should be
 	// interpreted as the "primary" identifier field. This field will be used as
 	// the primary field for resource adoption.
@@ -74,4 +81,22 @@ func (c *Config) ListOpMatchFieldNames(
 		return res
 	}
 	return rConfig.ListOperation.MatchFields
+}
+
+// UnmarshalJSON parses input for a either a string or
+// or a list and returns a StringArray.
+func (a *StringArray) UnmarshalJSON(b []byte) error {
+	var multi []string
+	err := json.Unmarshal(b, &multi)
+	if err != nil {
+		var single string
+		err := json.Unmarshal(b, &single)
+		if err != nil {
+			return err
+		}
+		*a = []string{single}
+	} else {
+		*a = multi
+	}
+	return nil
 }
--- a/pkg/model/sdk_helper.go
+++ b/pkg/model/sdk_helper.go
@@ -14,7 +14,6 @@
 package model
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -387,7 +386,6 @@ func (a *SDKAPI) SDKAPIInterfaceTypeName() string {
 
 // Override the operation type and/or resource name if specified in config
 func getOpTypeAndResourceName(opID string, cfg *ackgenconfig.Config) ([]OpType, string) {
-	var unmarshaledOpTypes ackgenconfig.StringArray
 	opType, resName := GetOpTypeAndResourceNameFromOpID(opID)
 	opTypes := []OpType{opType}
 
@@ -399,12 +397,9 @@ func getOpTypeAndResourceName(opID string, cfg *ackgenconfig.Config) ([]OpType, 
 			resName = operationConfig.ResourceName
 		}
 
-		if b, err := json.Marshal(operationConfig.OperationType); err == nil {
-			json.Unmarshal(b, &unmarshaledOpTypes)
-			for _, operationType := range unmarshaledOpTypes {
-				opType = OpTypeFromString(operationType)
-				opTypes = append(opTypes, opType)
-			}
+		for _, operationType := range operationConfig.OperationType {
+			opType = OpTypeFromString(operationType)
+			opTypes = append(opTypes, opType)
 		}
 	}
 	return opTypes, resName
--- a/pkg/model/sdk_helper.go
+++ b/pkg/model/sdk_helper.go
@@ -14,6 +14,7 @@
 package model
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -27,6 +28,7 @@ import (
 	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/generate/config"
 	"github.com/aws-controllers-k8s/code-generator/pkg/names"
 	"github.com/aws-controllers-k8s/code-generator/pkg/util"
+
 	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
 )
 
@@ -206,11 +208,13 @@ func (a *SDKAPI) GetOperationMap(cfg *ackgenconfig.Config) *OperationMap {
 	// create an index of Operations by operation types and resource name
 	opMap := OperationMap{}
 	for opID, op := range a.API.Operations {
-		opType, resName := getOpTypeAndResourceName(opID, cfg)
-		if _, found := opMap[opType]; !found {
-			opMap[opType] = map[string]*awssdkmodel.Operation{}
+		opTypeArray, resName := getOpTypeAndResourceName(opID, cfg)
+		for _, opType := range opTypeArray {
+			if _, found := opMap[opType]; !found {
+				opMap[opType] = map[string]*awssdkmodel.Operation{}
+			}
+			opMap[opType][resName] = op
 		}
-		opMap[opType][resName] = op
 	}
 	a.opMap = &opMap
 	return &opMap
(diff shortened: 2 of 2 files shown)


pkg/model/sdk_helper.go
```diff
@@ -382,20 +384,27 @@ func (a *SDKAPI) SDKAPIInterfaceTypeName() string {
...
-			}
-
 			if operationConfig.ResourceName != "" {
 				resName = operationConfig.ResourceName
 			}
-		}
-	}
 
-	return opType, resName
+			if len(operationConfig.OperationType) > 0 {
```
mbaijal, line 397: clean up

pkg/model/sdk_helper.go
```diff
@@ -382,20 +384,27 @@ func (a *SDKAPI) SDKAPIInterfaceTypeName() string {
 }
 
 // Override the operation type and/or resource name if specified in config
-func getOpTypeAndResourceName(opID string, cfg *ackgenconfig.Config) (OpType, string) {
+func getOpTypeAndResourceName(opID string, cfg *ackgenconfig.Config) ([]OpType, string) {
 	opType, resName := GetOpTypeAndResourceNameFromOpID(opID)
-
+	opTypeArray := make([]OpType, 0)
+	// TODO: or use var opTypeArray []OpType
```
A-Hilaly, line 390: Please remove the `TODO` comment. I suggest using `opTypes := []OpType{}` here

pkg/model/sdk_helper.go
```diff
@@ -382,20 +384,27 @@ func (a *SDKAPI) SDKAPIInterfaceTypeName() string {
...
 	if cfg != nil {
 		if operationConfig, exists := cfg.Operations[opID]; exists {
-			if operationConfig.OperationType != "" {
-				opType = OpTypeFromString(operationConfig.OperationType)
-			}
-
 			if operationConfig.ResourceName != "" {
 				resName = operationConfig.ResourceName
```
A-Hilaly, line 394: This call now overrides `resName` every iteration. Maybe panic if the resource name changes?

pkg/model/sdk_helper.go
A-Hilaly: Also the code below will be simpler if you initialize the array using `opTypes := []OpType{opType}`

A-Hilaly: You can get something very similar to 
```go
func getOpTypeAndResourceName(opID string, cfg *ackgenconfig.Config) ([]OpType, string) {
	opType, resName := GetOpTypeAndResourceNameFromOpID(opID)
	opTypes := []OpType{opType}

	if cfg == nil {
		return opTypes, resName
	}
	if operationConfig, exists := cfg.Operations[opID]; exists {
		if operationConfig.ResourceName != "" {
			resName = operationConfig.ResourceName
		}
		for _, operationType := range operationConfig.OperationType {
			opType = OpTypeFromString(operationType)
			opTypes = append(opTypes, opType)
		}
	}
	return opTypes, resName
}

```

mbaijal: thanks, this was the intent of the mentioned cleanup :)

pkg/model/sdk_helper.go
mbaijal: good point, I assumed a valid config.

mbaijal: On second thought, each opID will always have only a single resName so there is no overriding happening here, so a panic is unnecessary. Am I missing some case ?

A-Hilaly: My point is that, right now we are assuming that all operations have the same `resName`,which is most likely correct. However just in case, we need to panic if we receive multiple operations for different `resName`.

pkg/generate/config/operation.go
```diff
@@ -15,10 +15,14 @@ package config
 
 import (
 	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
-
+	"encoding/json"
```
jaypipes, line 18: This is a standard library import, so please place above the `awssdkmodel` import line and separate with an empty newline.

pkg/model/sdk_helper.go
```diff
@@ -21,12 +21,14 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"encoding/json"
```
jaypipes, line 24: Please run `go fmt` on your code or add an auto-formatter to your IDE. It will automatically correct the above into an alphabetical import order.

jaypipes (changes requested):
Couple tiny nits inline, @mbaijal, otherwise this looks great.

pkg/generate/config/operation.go
mbaijal: done.

pkg/model/sdk_helper.go
mbaijal: > add an auto-formatter to your IDE

I should really, always forget to run the formatter. Ran it manually on the two files for now.

pkg/generate/config/operation.go
```diff
@@ -14,11 +14,15 @@
 package config
 
 import (
-	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
-
+	"encoding/json"
 	"github.com/aws-controllers-k8s/code-generator/pkg/util"
+	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
```
A-Hilaly, line 19: please change the import section to:
```go
import (
	"encoding/json"

        awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"

	"github.com/aws-controllers-k8s/code-generator/pkg/util"
)
```

pkg/model/sdk_helper.go
```diff
@@ -382,20 +386,26 @@ func (a *SDKAPI) SDKAPIInterfaceTypeName() string {
...
+	}
+	if operationConfig, exists := cfg.Operations[opID]; exists {
+		if operationConfig.ResourceName != "" {
+			resName = operationConfig.ResourceName
+		}
 
-			if operationConfig.ResourceName != "" {
-				resName = operationConfig.ResourceName
+		if b, err := json.Marshal(operationConfig.OperationType); err == nil {
+			unmarshaledOpTypes.UnmarshalJSON(b)
```
A-Hilaly, line 403: You don't have to call the method `UnmarshalJSON` here, bur rather call `json.Unmarshal(b, &unmarshaledOpTypes)`

A-Hilaly (changes requested):

pkg/generate/config/operation.go
mbaijal: Sure. `gofmt` does not fix this. Would you recommend using  `goimports` or such to avoid such issues in the future ?

A-Hilaly: Yes exactly :). Sometimes `goimports` doesn't work as expected so you might need to fix the imports order manually. Here are the guidelines for imports https://github.com/golang/go/wiki/CodeReviewComments#imports

pkg/model/sdk_helper.go
```diff
@@ -382,20 +386,26 @@ func (a *SDKAPI) SDKAPIInterfaceTypeName() string {
...
+	}
+	if operationConfig, exists := cfg.Operations[opID]; exists {
+		if operationConfig.ResourceName != "" {
+			resName = operationConfig.ResourceName
+		}
 
-			if operationConfig.ResourceName != "" {
-				resName = operationConfig.ResourceName
+		if b, err := json.Marshal(operationConfig.OperationType); err == nil {
+			json.Unmarshal(b, &unmarshaledOpTypes)
```
A-Hilaly, line 403: the `Unmarshal` error should be handled here

pkg/model/sdk_helper.go
```diff
@@ -382,20 +386,26 @@ func (a *SDKAPI) SDKAPIInterfaceTypeName() string {
...
+		return opTypes, resName
+	}
+	if operationConfig, exists := cfg.Operations[opID]; exists {
+		if operationConfig.ResourceName != "" {
+			resName = operationConfig.ResourceName
+		}
 
-			if operationConfig.ResourceName != "" {
-				resName = operationConfig.ResourceName
+		if b, err := json.Marshal(operationConfig.OperationType); err == nil {
```
A-Hilaly, line 402: Thinking a second time about this, i think that you don't have to call `json.Marshal`/`json.Unmarshall` here - because `StringArray.UnmarshalJSON` will be called when `model.Model` is getting initialized https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/model/model.go#L700

mbaijal: Yep you are right :)

pkg/model/sdk_helper.go
mbaijal: Not applicable now.

pkg/model/sdk_helper.go
mbaijal: In this case, is operation.go the right place for the UnmarshalJSON method ?

pkg/generate/config/operation.go
mbaijal: Thank you!

pkg/model/sdk_helper.go
A-Hilaly: Yes it makes sense to have the type and it methods on the same file so it's good!

pkg/model/sdk_helper.go
```diff
@@ -27,6 +27,7 @@ import (
 	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/generate/config"
 	"github.com/aws-controllers-k8s/code-generator/pkg/names"
 	"github.com/aws-controllers-k8s/code-generator/pkg/util"
+
 	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
```
A-Hilaly, line 31: Tiny nit: this import should be in the second block of imports

A-Hilaly (approved):
Looks good! thanks @mbaijal!

/lgtm

pkg/model/sdk_helper.go
mbaijal: let me set up the goimports and create a tiny PR for this one.

PR #173, opened 2021-08-30
Prepend 'https://proxy.golang.org' to GOPROXY in Dockerfile for building controller image.

vijtrip2:
Description of changes:

* Prepend 'https://proxy.golang.org' to GOPROXY in Dockerfile for building controller image.
* This GoPROXY helps fix the building of controller image during e2e tests. Tested by locally running `make kind-test`


RedbackThomson:
/approve

When we do eventually update past 1.14, I would highly recommend we change this to use `|` rather than `,`. It is supported from 1.15 onwards. As per the documentation:
> If a proxy URL is followed by a pipe character, the go command will try the next proxy in the list after any error

PR #174, opened 2021-08-30
Refactor identifier and path logic to `crd.go`

brycahta:
Issue #, if available: [#922](https://github.com/aws-controllers-k8s/community/issues/922)

Description of changes:
* Moving 2 helper funcs to become *methods* of CRD Struct for clearer encapsulation:
  * `GetSanitizedMemberPath`
  * `GetIdentifiers`

### DIFF
--- a/pkg/generate/code/set_sdk.go
+++ b/pkg/generate/code/set_sdk.go
@@ -784,8 +784,8 @@ func setSDKReadMany(
 			}
 		}
 
-		// Field renames are handled in getSanitizedMemberPath
-		resVarPath, err = getSanitizedMemberPath(memberName, r, op, sourceVarName)
+		// Field renames are handled in GetSanitizedMemberPath
+		resVarPath, err = r.GetSanitizedMemberPath(memberName, op, sourceVarName)
 		if err != nil {
 			// if it's an identifier field check for singular/plural
 			if util.InStrings(memberName, shapeIdentifiers) {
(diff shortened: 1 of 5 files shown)


pkg/generate/code/set_sdk.go
```diff
@@ -784,8 +784,8 @@ func setSDKReadMany(
 			}
 		}
 
-		// Field renames are handled in getSanitizedMemberPath
-		resVarPath, err = getSanitizedMemberPath(memberName, r, op, sourceVarName)
+		// Field renames are handled in GetSanitizedMemberPath
+		resVarPath, err = r.GetSanitizedMemberPath(memberName, op, sourceVarName)
 		if err != nil {
```
RedbackThomson, line 789: Just double checking: Why do we not just do `if err == nil { continue }` here, instead?

RedbackThomson, line 789: Just double checking: Why do we not just do `if err == nil { continue }` here, instead?

RedbackThomson:
Just have one comment. Otherwise lgtm

/approve

RedbackThomson:
Just have one comment. Otherwise lgtm
/approve

pkg/generate/code/set_sdk.go
brycahta: In this context, `err == nil` means there was a field in the ReadMany shape that matched a field in the CRD exactly. 

For example, [DescribeReplicationGroupsInput](https://docs.aws.amazon.com/sdk-for-go/api/service/elasticache/#DescribeReplicationGroupsInput) is a ReadMany call, but has a *ReplicationGroupId* (note singular) field.

A-Hilaly (approved):
LGTM 👍

PR #175, opened 2021-08-31
update conditions on error for delete operation

surajkota:
Description of changes:
- Update conditions on delete in case of error. Right now the errors from delete are not reflected to the user
  - I have added a check to treat all errors as recoverable if the resource is being deleted as I am not sure what is the user experience if the resource hits a terminal condition after its deleted. Please advise

Testing:
Manually introduced an error from sdkDelete and verified it gets updated in conditions

### DIFF
--- a/templates/pkg/resource/sdk.go.tpl
+++ b/templates/pkg/resource/sdk.go.tpl
@@ -212,7 +212,7 @@ func (rm *resourceManager) updateConditions (
 		}
 	}
 
-	if !r.IsBeingDeleted() && (rm.terminalAWSError(err) || err ==  ackerr.SecretTypeNotSupported || err == ackerr.SecretNotFound) {
+	if rm.terminalAWSError(err) || err ==  ackerr.SecretTypeNotSupported || err == ackerr.SecretNotFound {
 		if terminalCondition == nil {
 			terminalCondition = &ackv1alpha1.Condition{
 				Type:   ackv1alpha1.ConditionTypeTerminal,
--- a/templates/pkg/resource/sdk.go.tpl
+++ b/templates/pkg/resource/sdk.go.tpl
@@ -224,7 +224,7 @@ func (rm *resourceManager) updateConditions (
 			errorMessage = err.Error()
 		} else {
 			awsErr, _ := ackerr.AWSError(err)
-			errorMessage = awsErr.Code() + ": " + awsErr.Message()
+			errorMessage = awsErr.Message()
 		}
 		terminalCondition.Status = corev1.ConditionTrue
 		terminalCondition.Message = &errorMessage
@@ -247,7 +247,7 @@ func (rm *resourceManager) updateConditions (
 			awsErr, _ := ackerr.AWSError(err)
 			errorMessage := err.Error()
 			if awsErr != nil {
-				errorMessage = awsErr.Code() + ": " + awsErr.Message()
+				errorMessage = awsErr.Message()
 			}
 			recoverableCondition.Message = &errorMessage
 		} else if recoverableCondition != nil {
--- a/templates/pkg/resource/sdk.go.tpl
+++ b/templates/pkg/resource/sdk.go.tpl
@@ -212,7 +212,7 @@ func (rm *resourceManager) updateConditions (
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


templates/pkg/resource/sdk.go.tpl
```diff
@@ -212,7 +212,7 @@ func (rm *resourceManager) updateConditions (
 		}
 	}
 
-	if rm.terminalAWSError(err) || err ==  ackerr.SecretTypeNotSupported || err == ackerr.SecretNotFound {
+	if !r.IsBeingDeleted() && (rm.terminalAWSError(err) || err ==  ackerr.SecretTypeNotSupported || err == ackerr.SecretNotFound) {
```
jaypipes, line 215: I suppose this is fine. But, would it be safer to just let the Terminal Condition be set on the resource even if it's being deleted?

templates/pkg/resource/sdk.go.tpl
```diff
@@ -224,7 +224,7 @@ func (rm *resourceManager) updateConditions (
 			errorMessage = err.Error()
 		} else {
 			awsErr, _ := ackerr.AWSError(err)
-			errorMessage = awsErr.Message()
+			errorMessage = awsErr.Code() + ": " + awsErr.Message()
```
jaypipes, line 227: unrelated change...

templates/pkg/resource/sdk.go.tpl
```diff
@@ -247,7 +247,7 @@ func (rm *resourceManager) updateConditions (
 			awsErr, _ := ackerr.AWSError(err)
 			errorMessage := err.Error()
 			if awsErr != nil {
-				errorMessage = awsErr.Message()
+				errorMessage = awsErr.Code() + ": " + awsErr.Message()
```
jaypipes, line 250: unrelated change...

templates/pkg/resource/sdk.go.tpl
brycahta: I don't think it's safe to set Terminal Condition here because if deletion encounters an error, then wouldn't the controller stop reconciling without deleting the finalizer?

surajkota:
@jaypipes @brycahta thanks for review

I have moved the unrelated change to separate PR, please take a look

templates/pkg/resource/sdk.go.tpl
surajkota: will this as-is for now, until we have another strategy. If a customer hits this(should be a corner case), the worse case is controller retires with back off

A-Hilaly (approved):

templates/pkg/resource/sdk.go.tpl
jaypipes: > I don't think it's safe to set Terminal Condition here because if deletion encounters an error, then wouldn't the controller stop reconciling without deleting the finalizer?

Yes, and that is what the controller should do I think. If the delete operation failed on the backend AWS service API call, then the resource will not be deleted from the AWS service, so we should keep the resource around on the k8s side, but in show a terminal condition.

brycahta: > Yes, and that is what the controller should do I think. If the delete operation failed on the backend AWS service API call, then the resource will not be deleted from the AWS service, so we should keep the resource around on the k8s side, but in show a terminal condition.

Does this mean the only way to delete that resource (after resolving any errors) is for users to manually remove the finalizer via `kubectl edit` ? That seems weird cx or am I missing some details

jaypipes (approved):
:+1:

templates/pkg/resource/sdk.go.tpl
jaypipes: > > Yes, and that is what the controller should do I think. If the delete operation failed on the backend AWS service API call, then the resource will not be deleted from the AWS service, so we should keep the resource around on the k8s side, but in show a terminal condition.
> 
> Does this mean the only way to delete that resource (after resolving any errors) is for users to manually remove the finalizer via `kubectl edit` ? That seems weird cx or am I missing some details

The correct user experience for manually removing a resource that is in a stuck state: using `kubectl delete --force`

PR #182, opened 2021-09-07
return nil if observed is nil from a successful delete

surajkota:
Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/931

Description of changes:
Solution 2 in the issue linked above. If `rm.Delete` does not return an error, and observed is also nil, return `nil, nil` to the caller instead of the resource passed to the method.

Testing:
`make test`
Manually using sagemaker controller

### DIFF
--- a/templates/pkg/resource/manager.go.tpl
+++ b/templates/pkg/resource/manager.go.tpl
@@ -157,7 +157,7 @@ func (rm *resourceManager) Delete(
 	if observed != nil {
 		return rm.onSuccess(observed)
 	}
-	return rm.onSuccess(r)
+	return nil, nil
 }
 
 // ARNFromName returns an AWS Resource Name from a given string name. This
--- a/templates/pkg/resource/manager.go.tpl
+++ b/templates/pkg/resource/manager.go.tpl
@@ -154,10 +154,7 @@ func (rm *resourceManager) Delete(
 		return rm.onError(r, err)
 	}
 
-	if observed != nil {
-		return rm.onSuccess(observed)
-	}
-	return nil, nil
+	return rm.onSuccess(observed)
 }
 
 // ARNFromName returns an AWS Resource Name from a given string name. This
@@ -266,6 +263,9 @@ func (rm *resourceManager) onError(
 	r *resource,
 	err error,
 ) (acktypes.AWSResource, error) {
+	if ackcompare.IsNil(r) {
+		return nil, err
+	}
 	r1, updated := rm.updateConditions(r, false, err)
 	if !updated {
 		return r, err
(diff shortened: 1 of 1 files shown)


A-Hilaly (approved):

templates/pkg/resource/manager.go.tpl
```diff
@@ -157,7 +157,7 @@ func (rm *resourceManager) Delete(
 	if observed != nil {
 		return rm.onSuccess(observed)
 	}
-	return rm.onSuccess(r)
+	return nil, nil
```
vijtrip2, line 160: I think this is not needed anymore because your runtime PR.

`Delete` method should return the AWS resource returned from the AWS delete call.
And the issue that we were patching status on unmanaged resource can be handled in runtime code (alongside the context where we mark the resource unmanaged).

surajkota: Right, that is handled in line 158. This is for the case where resource and error both are returned nil from sdkDelete. With this change, we are returning the output from sdkDelete instead of the AWS resource that was passed to this method. Wdyt?

vijtrip2: Okay, makes sense & sounds good to me.
Sorry I overlooked.

One suggestion though, how about the code becomes like following. The nil check moves to `rm.onSuccess()` method
```go
...
    return rm.onSuccess(observed)
}

func (rm *resourceManager) onSuccess(
	r *resource,
) (acktypes.AWSResource, error) {
        if ackCompare.isNil(r) {
             return nil, nil
        }
	r1, updated := rm.updateConditions(r, true, nil)
	if !updated {
		return r, nil
	}
	return r1, nil
}
```

surajkota: Good point, this is cleaner and robust. Will make the change

surajkota: Done!

templates/pkg/resource/manager.go.tpl
```diff
@@ -266,6 +263,9 @@ func (rm *resourceManager) onError(
 	r *resource,
 	err error,
 ) (acktypes.AWSResource, error) {
+	if ackcompare.IsNil(r) {
+		return nil, err
+	}
```
A-Hilaly, line 268: Shouldn't this work just with a simple nil comparison? `if r != nil { ... }`
/cc @vijtrip2 @surajkota

jaypipes: Yes, @A-Hilaly is correct. We do not need the `ackcompare.IsNil` here. `r` is a `*resource`, which is a pointer type, not an interface.

surajkota: isNil method [handles this case](https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/compare/nil.go#L36-L38) as well

jaypipes: @surajkota no, the `ackcompare.IsNil` should not be used here.

PR #183, opened 2021-09-07
Support nested paths in field config specifically list type members

mbaijal:
Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/932

Description of changes:
`getMemberByPath` checks the shape type as it loops through members to ensure lists are handled correctly. Currently I have simply ignored the empty element created by the splitting of the string on the single dot instead of handling the double dot notation explicitly. Open to comments. 
Also added a Unit Test for the same. 

Testing
1. Generated code for ApplicationAutoscaling and ensured CreationTime field now works as expected. Linked changes [in this PR](https://github.com/aws-controllers-k8s/applicationautoscaling-controller/pull/39).
2. Ensured that the SageMaker Controller can be generated using new changes with no diff as expected. 

### DIFF
--- a/pkg/model/sdk_helper.go
+++ b/pkg/model/sdk_helper.go
@@ -246,7 +246,9 @@ func (a *SDKAPI) GetOutputShapeRef(
 }
 
 // getMemberByPath returns a ShapeRef given a root Shape and a dot-notation
-// object search path
+// object search path. Given the explicit type check for list type members 
+// both ".." and "." notations work currently. 
+// TODO: Add support for other types such as map.
 func getMemberByPath(
 	shape *awssdkmodel.Shape,
 	path string,
@@ -254,6 +256,9 @@ func getMemberByPath(
 	elements := strings.Split(path, ".")
 	last := len(elements) - 1
 	for x, elem := range elements {
+		if elem == "" {
+			continue
+		}
 		if shape == nil {
 			return nil, false
 		}
@@ -264,7 +269,13 @@ func getMemberByPath(
 		if x == last {
 			return shapeRef, true
 		}
-		shape = shapeRef.Shape
+		elemType := shapeRef.Shape.Type
+		switch elemType {
+		case "list":
+			shape = shapeRef.Shape.MemberRef.Shape
+		default:
+			shape = shapeRef.Shape
+		}
 	}
 	return nil, false
 }
--- a/pkg/model/sdk_helper_test.go
+++ b/pkg/model/sdk_helper_test.go
@@ -134,6 +134,13 @@ func TestGetOutputShapeRef(t *testing.T) {
 			true,
 		},
 		{
+			// list type nested path search
+			"ListAliases",
+			"Aliases..RevisionId",
+			&stringshape,
+			true,
+		},
+		{
 			// no such op
 			"GetNonexisting",
 			"Foo",


pkg/model/sdk_helper.go
```diff
@@ -254,6 +254,9 @@ func getMemberByPath(
 	elements := strings.Split(path, ".")
 	last := len(elements) - 1
 	for x, elem := range elements {
+		if elem == "" {
+			continue
+		}
```
jaypipes, line 259: Yeah, this is fine for now. I suppose we could also warn or panic if the `..` notation was used to indicate a non-list/map field but I'm cool just ignoring for now.

pkg/model/sdk_helper.go
```diff
@@ -254,6 +254,9 @@ func getMemberByPath(
 	elements := strings.Split(path, ".")
```
jaypipes, line 254: Might be worth calling out in the docstring that `..` notation for list/map types is currently ignored.

pkg/model/sdk_helper.go
```diff
@@ -264,7 +267,13 @@ func getMemberByPath(
 		if x == last {
 			return shapeRef, true
 		}
-		shape = shapeRef.Shape
+		elemType := shapeRef.Shape.Type
+		switch elemType {
+		case "list":
+			shape = shapeRef.Shape.MemberRef.Shape
```
jaypipes, line 273: Eventually we should support `map` type as well, which would be:

```go
case "map":
    shape = shapeRef.Shape.ValueRef.Shape
```

jaypipes: You can do that in a followup PR, though, since this one is for lists only (and dealing with map types in the code generator is a little tougher anyway)

jaypipes (approved):

RedbackThomson (approved):
Is this worthy of some simple unit tests in the `sdk_helper_test.go` file?

pkg/model/sdk_helper.go
mbaijal: Actually, this check is required since the path string in case of a list is currently split into 3 elements and we need to ignore the second one - 

input Path - `ScalableTargets..CreationTime`
the elements array in line 254 - ["ScalableTargets", "", "CreationTime"]

and we need to make sure we don't process the second elem. Another option would be to directly parse for ".." and handle lists but that seemed unnecessary at the moment.

pkg/model/sdk_helper.go
mbaijal: Sounds good, let me create an issue for that such that it is tracked.

pkg/model/sdk_helper.go
mbaijal: sure, i will add the comment based on the discussion on the above comment.

pkg/model/sdk_helper.go
mbaijal: @jaypipes just making sure we are on the same page 👆🏻

mbaijal:
> Is this worthy of some simple unit tests in the `sdk_helper_test.go` file?

Probably, let me give that a go.

pkg/model/sdk_helper.go
jaypipes: Yep, we're all good.

pkg/model/sdk_helper.go
mbaijal: Added a comment and TODO.

A-Hilaly (approved):
LGTM :+1:

PR #184, opened 2021-09-08
Only return MetaObject from AWSResource.MetaObject() call

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/937

Description of changes:
* For implementation of AWSResource.MetaObject, we were returning the complete resource.
* This change only returns 'ObjectMeta' from AWSResource which is the implementation of 'metav1.Object' inside resource struct
* returning only the relevant members helps in metadata comparison correctly with `ackcompare.MetaV1ObjectEqual`

---------------

* Unit tests successful
* Successful local kind e2e test for newly generated dynamodb controller

### DIFF
--- a/templates/pkg/resource/resource.go.tpl
+++ b/templates/pkg/resource/resource.go.tpl
@@ -46,7 +46,7 @@ func (r *resource) RuntimeObject() k8srt.Object {
 // MetaObject returns the Kubernetes apimachinery/apis/meta/v1.Object
 // representation of the AWSResource
 func (r *resource) MetaObject() metav1.Object {
-	return r.ko
+	return r.ko.GetObjectMeta()
 }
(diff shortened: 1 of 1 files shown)


RedbackThomson:
I believe we are using `MetaObject` as part of the adoption reconciler. Would you mind verifying these still work as intended, as well? Just a simple smoke test should suffice.

vijtrip2:
> I believe we are using `MetaObject` as part of the adoption reconciler. Would you mind verifying these still work as intended, as well? Just a simple smoke test should suffice.

There should be no regression from this change at all but I will perform sagemaker adoption test nevertheless.

RedbackThomson:
Thanks! 
/lgtm

PR #190, opened 2021-09-10
Replace `primary_identifier_field_name` with `is_primary_key`

RedbackThomson:
Description of changes:
We have had quite a few issues with the logic of `SetResourceIdentifiers` struggling to find the correct identifier field name or matching it to the corresponding spec or status field. The current implementation solves the first of these issues by pointing directly to the identifier member in the input shape, but the logic to link this back to a spec or status field is still flawed. 
Rather than pointing to a member in the input shape and trying to infer the spec or status field, this new implementation directly denotes a field on a resource as the "primary key". The `SetResourceIdentifiers` logic can now directly find the target field and create the corresponding setter. The override logic for ARNs has now been moved to a `is_arn_primary_key` on the `ResourceConfig` object, since we cannot access the ARN as a `FieldConfig`.

This is a **backwards compatibility breaking change** as all existing `primary_identifier_field_name` configurations need to be refactored into `is_primary_key` fields on the respective resource's field. For example:
```yaml
operations:
  DescribeDBInstances:
    primary_identifier_field_name: DBInstanceIdentifier
  DescribeDBSubnetGroups:
    primary_identifier_field_name: DBSubnetGroupName
```

becomes:
```yaml
resources:
  DBInstance:
    fields:
      DBInstanceIdentifier:
        is_primary_key: true
  DBSubnetGroup:
    fields:
      Name:
        is_primary_key: true
```

### DIFF
--- a/pkg/generate/code/common.go
+++ b/pkg/generate/code/common.go
@@ -131,12 +131,6 @@ func FindPrimaryIdentifierFieldNames(
 ) (crField string, shapeField string) {
 	shape := op.InputRef.Shape
 
-	// Attempt to fetch the primary identifier override from the configuration
-	opConfig, ok := cfg.Operations[op.Name]
-	if ok {
-		shapeField = opConfig.PrimaryIdentifierFieldName
-	}
-
 	if shapeField == "" {
 		// For ReadOne, search for a direct identifier
 		if op == r.Ops.ReadOne {
@@ -151,8 +145,8 @@ func FindPrimaryIdentifierFieldNames(
 			default:
 				panic("Found multiple possible primary identifiers for " +
 					r.Names.Original + ". Set " +
-					"`primary_identifier_field_name` for the " + op.Name +
-					" operation in the generator config.")
+					"`is_primary_key` for the primary field in the " +
+					r.Names.Camel + " resource.")
 			}
 		} else {
 			// For ReadMany, search for pluralized identifiers
@@ -162,12 +156,12 @@ func FindPrimaryIdentifierFieldNames(
 		// Require override if still can't find any identifiers
 		if shapeField == "" {
 			panic("Could not find primary identifier for " + r.Names.Original +
-				". Set `primary_identifier_field_name` for the " + op.Name +
-				" operation in the generator config.")
+				". Set `is_primary_key` for the primary field in the " +
+				r.Names.Camel + " resource.")
 		}
 	}
 
-	if r.IsPrimaryARNField(shapeField) || shapeField == PrimaryIdentifierARNOverride {
+	if r.IsPrimaryARNField(shapeField) {
 		return "", PrimaryIdentifierARNOverride
 	}
 
--- a/pkg/generate/code/set_resource.go
+++ b/pkg/generate/code/set_resource.go
@@ -898,7 +898,7 @@ func SetResourceIdentifiers(
 	var primaryCRField, primaryShapeField string
 	isPrimarySet := primaryField != nil
 	if isPrimarySet {
-		memberPath, _ := findFieldInCR(cfg, r, primaryField.Names.Camel)
+		memberPath, _ := findFieldInCR(cfg, r, primaryField.Names.Original)
 		targetVarPath := fmt.Sprintf("%s%s", targetVarName, memberPath)
 		primaryKeyOut += setResourceIdentifierPrimaryIdentifier(cfg, r,
 			primaryField,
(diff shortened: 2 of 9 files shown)


RedbackThomson:
Seeing an issue with additional keys duplicating the primary key:
```go
	if identifier.NameOrID == "" {
		return ackerrors.MissingNameIdentifier
	}
	r.ko.Spec.ResourceID = &identifier.NameOrID

	f3, f3ok := identifier.AdditionalKeys["resourceID"]
	if f3ok {
		r.ko.Spec.ResourceID = &f3
	}
	f4, f4ok := identifier.AdditionalKeys["scalableDimension"]
	if f4ok {
		r.ko.Spec.ScalableDimension = &f4
	}
	f5, f5ok := identifier.AdditionalKeys["serviceNamespace"]
	if f5ok {
		r.ko.Spec.ServiceNamespace = &f5
	}

	return nil
}
```

brycahta:
Thanks a lot for this @RedbackThomson -- should unblock some of my work as well! 

I'm downloading this PR locally to test, but will post questions here prior to SIG for discussion:
1. **Interface** seems a little clunky. I agree with setting at the `resource` level, but I don't think `is_primary_key` is clearer or better than `is_identifier` (or similar).
  * Defining nested paths won't be clean
  * Should setting `arn` as primary key exist on same level (in yaml)?


2. **Implementation** doesn't seem to handle nested paths, can you confirm? Ex: I need `VpcEndpoint.vpcEndpointId` from *DescribeVpcEndpoints* response

3. Would it be safe to remove some of that custom identifier logic I introduced a little while back? If not, maybe it could be simplified in future PR

brycahta (approved):
LGTM

pkg/generate/code/set_resource.go
```diff
@@ -873,21 +875,42 @@ func SetResourceIdentifiers(
...
-		//  r.ko.Status.ACKResourceMetadata = &ackv1alpha1.ResourceMetadata{}
-		// }
-		// r.ko.Status.ACKResourceMetadata.ARN = identifier.ARN
-		arnOut := "\n"
-		arnOut += ackResourceMetadataGuardConstructor(fmt.Sprintf("%s.Status", targetVarName), indentLevel)
-		arnOut += fmt.Sprintf(
-			"%s%s.Status.ACKResourceMetadata.ARN = %s.ARN\n",
-			indent, targetVarName, sourceVarName,
-		)
+	// if r.ko.Status.ACKResourceMetadata == nil {
```
mbaijal, line 878: what's up with this commented out code ?

pkg/generate/code/common.go
```diff
@@ -162,12 +156,12 @@ func FindPrimaryIdentifierFieldNames(
...
 			panic("Could not find primary identifier for " + r.Names.Original +
-				". Set `primary_identifier_field_name` for the " + op.Name +
-				" operation in the generator config.")
+				". Set `is_primary_key` for the primary field in the " +
+				r.Names.Camel + " resource.")
 		}
 	}
 
-	if r.IsPrimaryARNField(shapeField) || shapeField == PrimaryIdentifierARNOverride {
+	if r.IsPrimaryARNField(shapeField) {
```
mbaijal, line 164: ApplicationAutoscaling does not utilize this, so I have not verified this snippet, looks ok though.

pkg/generate/config/field.go
```diff
@@ -170,11 +170,11 @@ type FieldConfig struct {
...
 	IsRequired *bool `json:"is_required,omitempty"`
-	// IsName indicates the field represents the name/string identifier field
-	// for the resource.  This allows the generator config to override the
-	// default behaviour of considering a field called "Name" or
+	// IsPrimaryKey indicates the field represents the primary name/string
+	// identifier field for the resource.  This allows the generator config to
+	// override the default behaviour of considering a field called "Name" or
 	// "{Resource}Name" or "{Resource}Id" as the "name field" for the resource.
-	IsName bool `json:"is_name"`
+	IsPrimaryKey bool `json:"is_primary_key"`
```
mbaijal, line 177: My local config has both `is_name` and `is_primary_key` since I had not looked at this code yet. But generator does not throw an error. We should have some validation in the future for extra fields too.

mbaijal (approved):
Haven't reviewed thoroughly but the code works for both ApplicationAutoscaling resources exactly as expected. Approving based on functionality.

pkg/generate/config/field.go
RedbackThomson: Yes, I did a global search for all service controllers using `is_name` and I believe SageMaker is the only example of them. I also agree that the generator should at least warn teams if they have unspecified fields.

pkg/generate/code/common.go
RedbackThomson: The SageMaker controller uses it (for ModelPackage). We have unit tests that cover it.

PR #192, opened 2021-09-10
Add hook into SetIdentifier for custom code before ReadOne is called from AdoptedResource.

mbaijal:
Issue #, if available:
This is currently required as a temporary fix to be able to use AdoptedResource feature for ScalableTarget in the ApplicationAutoscaling Service but is a useful hook to have. 

Description of changes:
Adds a hook into SetIdentifier for custom code before ReadOne is called from AdoptedResource.

Sample Usage:
as seen in [this PR ](https://github.com/aws-controllers-k8s/applicationautoscaling-controller/pull/42)
```
set_identifiers_pre_read_one:
       code: r.customSetIdentifierCode(identifier)
```

### DIFF
--- a/templates/pkg/resource/resource.go.tpl
+++ b/templates/pkg/resource/resource.go.tpl
@@ -80,6 +80,9 @@ func (r *resource) SetStatus(desired acktypes.AWSResource) {
 // resource identifier
 func (r *resource) SetIdentifiers(identifier *ackv1alpha1.AWSIdentifiers) error {
 {{- GoCodeSetResourceIdentifiers .CRD "identifier" "r.ko" 1}}
+{{- if $hookCode := Hook .CRD "set_identifiers_pre_read_one" }}
+{{ $hookCode }}
+{{- end }}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


templates/pkg/resource/resource.go.tpl
```diff
@@ -80,6 +80,9 @@ func (r *resource) SetStatus(desired acktypes.AWSResource) {
 // resource identifier
 func (r *resource) SetIdentifiers(identifier *ackv1alpha1.AWSIdentifiers) error {
 {{- GoCodeSetResourceIdentifiers .CRD "identifier" "r.ko" 1}}
+{{- if $hookCode := Hook .CRD "set_identifiers_pre_read_one" }}
```
vijtrip2, line 83: The position of hook is not pre read one. There is no read one call here.

I agree this can be a useful hook. maybe rename it to just "set_identifiers"

Just a thought, Should there be either hook code OR auto-generated code inside method body? 
This is such a simple method, no need to have both.

RedbackThomson:
I believe this is not necessary anymore. The recent replacement with https://github.com/aws-controllers-k8s/code-generator/pull/190 has fixed your use case for applicationautoscaling, and I imagine should fix most other issues. Happy to re-open this PR if hooks are needed in the future.

PR #196, opened 2021-09-16
Use docker expiration string from auth config

RedbackThomson:
Description of changes:
Just checking the existence of the authorization string in Docker config is not sufficient to determine whether the user is logged into ECR public

This PR will decode the expiration timestamp from the authorization string and compare it to the current UTC timestamp.


vijtrip2 (approved):
/lgtm

PR #201, opened 2021-09-21
add ability to skip fields for set output step

echen-98:
Spec patching can be problematic for certain fields. For example, modifying an ElastiCache Replication Group's `spec.LogDeliveryConfigurations` from value `A` to value `B` causes the controller to infinitely loop. `desired` will have value `B` and `latest` will have value `A`. They will then switch and this continues indefinitely.

Since we cannot disable spec patching, the point of this PR is to add a workaround where service teams can choose not to merge in a particular field of the API response, so the user's desired state will not be overwritten by an incorrect/not up to date server value upon spec patching.

### DIFF
--- a/pkg/generate/code/set_resource.go
+++ b/pkg/generate/code/set_resource.go
@@ -144,10 +144,32 @@ func SetResource(
 	out := "\n"
 	indent := strings.Repeat("\t", indentLevel)
 
+	// retrieve the operation config for the given operation
+	var operationConfig *ackgenconfig.OperationConfig
+	if val, ok := cfg.Operations[op.Name]; ok {
+		operationConfig = &val
+	}
+	var skipFields ackgenconfig.StringArray
+	if operationConfig != nil {
+		skipFields = operationConfig.SetOutputSkipFields
+	}
+
 	// Recursively descend down through the set of fields on the Output shape,
 	// creating temporary variables, populating those temporary variables'
 	// fields with further-nested fields as needed
 	for memberIndex, memberName := range outputShape.MemberNames() {
+		// skip the output field if specified in the operation config
+		skipField := false
+		for _, field := range skipFields {
+			if memberName == field {
+				skipField = true
+				break
+			}
+		}
+		if skipField {
+			continue
+		}
+
 		//TODO: (vijat@) should these field be renamed before looking them up in spec?
 		sourceAdaptedVarName := sourceVarName + "." + memberName
 
--- a/pkg/generate/config/operation.go
+++ b/pkg/generate/config/operation.go
@@ -45,6 +45,8 @@ type OperationConfig struct {
 	// An example of this is `Put...` or `Register...` API operations not being correctly classified as `Create` op type
 	// OperationType []string `json:"operation_type"`
 	OperationType StringArray `json:"operation_type"`
+	// Fields for which to skip generation of code for merging in API response
+	SetOutputSkipFields StringArray `json:"set_output_skip_fields"`
 }
 
 // IsIgnoredOperation returns true if Operation Name is configured to be ignored


vijtrip2:
Hi @echen-98 , Can you elaborate the description of problem a little bit more? Why does not Update call return `B` in the update output which will be patched back into etcd, making both latest and desired have value `B` ?

------------------

>desired will have value B and latest will have value A. They will then switch and this continues indefinitely

I do not understand why the desired and latest will switch infinitely. ACK runtime patches the spec only from Create or Update output. 
Is this resource's update async ? 
If yes, then Is there some information in UpdateResponse which tells that configuration is being updated on the server side?
If yes, then From my understanding, you should be able to use hooks to requeue the request/add conditions which tell user that update is in progress.

echen-98:
> Hi @echen-98 , Can you elaborate the description of problem a little bit more? Why does not Update call return `B` in the update output which will be patched back into etcd, making both latest and desired have value `B` ?
> 
> > desired will have value B and latest will have value A. They will then switch and this continues indefinitely
> 
> I do not understand why the desired and latest will switch infinitely. ACK runtime patches the spec only from Create or Update output.
> Is this resource's update async ?
> If yes, then Is there some information in UpdateResponse which tells that configuration is being updated on the server side?
> If yes, then From my understanding, you should be able to use hooks to requeue the request/add conditions which tell user that update is in progress.

@vijtrip2 to answer your question, it is an async update. Here are more details on why it loops:

1. Before any updates, `desired` has `A` and `latest` has `A`.
2. The user updates `desired` to `B`. After the next `ReadOne`, desired has `B` and `latest` has `A`, since the latest state has not changed yet.
3. The controller detects a difference, and invokes a modify call with `B` as the input.
4. The ElastiCache API returns a response. The value in the response is actually `A` (the old value), with the caveat that it also returns status `modifying`. In other words, the resource is in the process of moving from `A` to `B` but the API still returns `A`.
5. The consequence is that `latest` gets A. Once the spec is patched, `desired` will have `A` (due to patching) and `latest` will also have `A` (since that's what the API returned).
6. Once the modification has finished, the `ReadOne` API response will return `B`. At this point, `desired` has `A` and `latest` has `B` once this response is merged into `latest`.
7. Note that this is flipped from what we had before. The controller will detect a difference, and invoke a modify call with `A` as the input.
8. This entire process will continue to repeat.

echen-98:
> From my understanding, you should be able to use hooks to requeue the request/add conditions which tell user that update is in progress.

For the latter point, the user knows that an update is in progress since the resource will have the `Synced` condition false. Further, `Status.LogDeliveryConfigurations.Status` will show `modifying`.

Could you maybe elaborate on your first point regarding hooks/requeues? We are already requeue for a potential update request any time the synced condition is false. I'm not sure how to solve this issue without circumventing spec patching in some way.

vijtrip2:
@echen-98 , got it. Thanks for explaining it in the detail.

----------------

I feel this behavior is an exception case and you should be able to leverage existing hooks like :
1.  `sdk_update_pre_build_request` -> to check if `Status.LogDeliveryConfigurations.Status` is `modifying` then just throw requeue error

2.
```go
==sdk_update.go.tpl==
{{- if $hookCode := Hook .CRD "sdk_update_pre_set_output" }}
{{ $hookCode }}
{{- end }}
{{ GoCodeSetUpdateOutput .CRD "resp" "ko" 1 }}
	rm.setStatusDefaults(ko)
{{- if $setOutputCustomMethodName := .CRD.SetOutputCustomMethodName .CRD.Ops.Update }}
	// custom set output from response
	ko, err = rm.{{ $setOutputCustomMethodName }}(ctx, desired, resp, ko)
	if err != nil {
		return nil, err
	}
{{- end }}
{{- if $hookCode := Hook .CRD "sdk_update_post_set_output" }}
{{ $hookCode }}
{{- end }}
```

And from the above code for update fields, you can place your custom code for `setOutput` or use the hook `sdk_update_post_set_output` to make sure if the `LogDeliveryConfigurations.Status` is `modifying`, then do not update the `ko.Spec.LogDeliveryConfigurations`

-----------------------

Will these already existing solutions not work for you to handle this async case?

echen-98:
@vijtrip2 thanks for the suggestions.

With suggestion 1, our custom code already does this because any time `Status.LogDeliveryConfigurations.Status` is `modifying`, the replication group's `Status.Status` is also `modifying` (this is service behavior) and we [block/requeue all subsequent update requests](https://github.com/aws-controllers-k8s/elasticache-controller/blob/main/pkg/resource/replication_group/custom_update_api.go#L56-L59) until the replication group becomes available. The problem is not repeated modifications while the resource is modifying; rather, the problem is what happens within the initial modification.

This brings me to your second suggestion, which is to make use of the custom set output method and/or the `post_set_output` hook to "not update `ko.Spec.LogDeliveryConfigurations`". This occurred to me yesterday and it would be possible, but the issue is both of these come _after_ `ko.Spec.LogDeliveryConfigurations` has already been [set to the response from the modify API](https://github.com/aws-controllers-k8s/elasticache-controller/blob/main/pkg/resource/replication_group/sdk.go#L1232-L1268). Therefore, we would be re-setting the value to what it was originally, or essentially copying the state from `desired` over to `latest`. This solution would then result in the following flow for `sdkUpdate`:

1. Copy the `desired` `LogDeliveryConfigurations` state into `latest`
2. Merge in the API response `LogDeliveryConfigurations` to `latest`
3. Reset `latest`'s `LogDeliveryConfigurations` to the original value from `desired`

Steps 2 and 3 are redundant, and would result in a lot of excess code which would make our controller even less readable/maintainable and more error-prone than it already is. I'm just not convinced that the existing tools are good enough for this case, hence this PR.

echen-98:
> I feel this behavior is an exception case

We've already identified other fields that also face this issue. I think it's likely to happen with any field that is updated asynchronously, which is quite common for ElastiCache. 

I think we should also evaluate how/when spec patching is done but that's out of scope of this PR. For now we just need a workaround for fields which have issues with spec patching.

vijtrip2:
@echen-98 , Thanks for detailed explanation. I hear your point and agree that custom code will only bring redundancy in this case.

------------------

How are the error cases handled in this scenario? If this async update errors out, does the status.conditions reflect it?

------------------

* Can you please add unit tests that generate code with this functionality in `generator.yaml` ?
* I would request @jaypipes and @RedbackThomson to give this new configuration a look as well before merging the proposed code.

echen-98:
@vijtrip2 thanks! I'll get started on unit tests.

> How are the error cases handled in this scenario? If this async update errors out, does the status.conditions reflect it?

If the new field value is invalid, then the server should return some 4xx code and choose not to modify the replication group, which would result in a terminal condition. 

I'm not sure what happens if the server accepts the modification then fails later, and there's not really a way I can reliably test that since I can't force the server to fail the modification. My guess is that the replication group would roll back to its original state, with `status.status` `available` and synced condition `True`. Upon the next `ReadOne` the controller should detect that there is still a difference between `desired` and `latest`, then try the modification again.

jaypipes (changes requested):
As mentioned in private slack, @echen-98 I'd prefer you to use the `sdk_XXX_post_set_output` hooks for this functionality, at least for the time being.

Thanks!
-jay

PR #202, opened 2021-09-22
Update to controller-gen `v0.7.0`

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/951

Description of changes:
* Use latest controller-gen release, which unblocks aws-sdk-go upgrade for apigatewayv2 controller
* do not restrict map value type when using `reflect.DeepEqual` method

### DIFF
--- a/pkg/generate/code/compare_test.go
+++ b/pkg/generate/code/compare_test.go
@@ -376,3 +376,102 @@ func TestCompareResource_Lambda_Function(t *testing.T) {
 		),
 	)
 }
+
+func TestCompareResource_APIGatewayv2_Route(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+
+	g := testutil.NewModelForService(t, "apigatewayv2")
+
+	crd := testutil.GetCRDByName(t, g, "Route")
+	require.NotNil(crd)
+
+	expected := `
+	if ackcompare.HasNilDifference(a.ko.Spec.APIID, b.ko.Spec.APIID) {
+		delta.Add("Spec.APIID", a.ko.Spec.APIID, b.ko.Spec.APIID)
+	} else if a.ko.Spec.APIID != nil && b.ko.Spec.APIID != nil {
+		if *a.ko.Spec.APIID != *b.ko.Spec.APIID {
+			delta.Add("Spec.APIID", a.ko.Spec.APIID, b.ko.Spec.APIID)
+		}
+	}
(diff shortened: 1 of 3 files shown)


pkg/generate/code/compare_test.go
```diff
@@ -376,3 +376,102 @@ func TestCompareResource_Lambda_Function(t *testing.T) {
...
+		if !ackcompare.MapStringStringPEqual(a.ko.Spec.RequestModels, b.ko.Spec.RequestModels) {
+			delta.Add("Spec.RequestModels", a.ko.Spec.RequestModels, b.ko.Spec.RequestModels)
+		}
+	}
+	if ackcompare.HasNilDifference(a.ko.Spec.RequestParameters, b.ko.Spec.RequestParameters) {
+		delta.Add("Spec.RequestParameters", a.ko.Spec.RequestParameters, b.ko.Spec.RequestParameters)
+	} else if a.ko.Spec.RequestParameters != nil && b.ko.Spec.RequestParameters != nil {
+		if !reflect.DeepEqual(a.ko.Spec.RequestParameters, b.ko.Spec.RequestParameters) {
+			delta.Add("Spec.RequestParameters", a.ko.Spec.RequestParameters, b.ko.Spec.RequestParameters)
+		}
```
RedbackThomson, line 447: Just copying from the `DeepEqual` docs:
> Map values are deeply equal when all of the following are true: they are both nil or both non-nil, they have the same length, and either they are the same map object or their corresponding keys (matched using Go equality) map to deeply equal values. 

`DeepEqual` won't give granularity into which key was different - this may be a restriction for some packages that want to ignore diffs on a given key.

I think a better method for us would be to iterate through the keys, and `DeepEqual` them individually (basically recreating the `DeepEqual` functionality), adding each difference to the list of deltas.

RedbackThomson (commented):
Great work pushing out that nested maps PR in the controller-tools! I'm sure there will be a lot of happy developers elsewhere that benefit from that change.

I've left a comment regarding the use of `DeepEqual` below. LMK if you see the advantage in that work.

pkg/generate/code/compare_test.go
vijtrip2: Yes, I agree & I see value in that. But this should be done in a separate issue/PR. I will create an issue for it.

This will not just apply to `DeepEqual` but also to our existing `runtime/pkg/compare/map.go` methods

PR #210, opened 2021-09-28
Reference `output_wrapper_field_path` when generating CRD fields as part of the `Create*` request in `model.go`

brycahta:
Issue #, if available:
* `model.go` **does not** take into account any `output_wrapper_field_path` overrides when processing Status fields for a CRD

Description of changes:
* Refactored `GetOutputShape` api:
  * always return an *output* shape
  * make `getWrapperOutputShape` private
  * updated tests
* Added `GetOutputShape` invocation to `model.go` so that overrides are checked before creating the CRD

Local Testing:
* *generator.yaml*
```
operations:
  CreateVpcEndpoint:
```

* `make build-controller SERVICE=ec2` 
* **Result:** `VPCEndpointStatus` contains all unpacked fields. Assigning from aws sdk response (`vpc_endpoint/sdk.go`) includes `resp.VpcEndpoint` as prefix

### DIFF
--- a/pkg/generate/code/set_resource.go
+++ b/pkg/generate/code/set_resource.go
@@ -110,23 +110,14 @@ func SetResource(
 	if op == nil {
 		return ""
 	}
-	outputShape := op.OutputRef.Shape
+	outputShape, _ := r.GetOutputShape(op)
 	if outputShape == nil {
 		return ""
 	}
 
-	var err error
-	// We might be in a "wrapper" shape. Unwrap it to find the real object
-	// representation for the CRD's createOp/DescribeOP.
-
 	// Use the wrapper field path if it's given in the ack-generate config file.
 	wrapperFieldPath := r.GetOutputWrapperFieldPath(op)
 	if wrapperFieldPath != nil {
-		outputShape, err = r.GetWrapperOutputShape(outputShape, *wrapperFieldPath)
-		if err != nil {
-			msg := fmt.Sprintf("Unable to unwrap the output shape: %v", err)
-			panic(msg)
-		}
 		sourceVarName += "." + *wrapperFieldPath
 	} else {
 		// If the wrapper field path is not specified in the config file and if
--- a/pkg/generate/code/set_resource_test.go
+++ b/pkg/generate/code/set_resource_test.go
@@ -14,7 +14,6 @@
 package code_test
 
 import (
-	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -23,7 +22,6 @@ import (
 	"github.com/aws-controllers-k8s/code-generator/pkg/generate/code"
 	"github.com/aws-controllers-k8s/code-generator/pkg/model"
 	"github.com/aws-controllers-k8s/code-generator/pkg/testutil"
-	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
 )
 
 func TestSetResource_APIGWv2_Route_Create(t *testing.T) {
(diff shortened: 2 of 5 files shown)


pkg/generate/code/set_resource.go
```diff
@@ -110,23 +110,14 @@ func SetResource(
 	if op == nil {
 		return ""
 	}
-	outputShape := op.OutputRef.Shape
+	outputShape, _ := r.GetOutputShape(op)
```
RedbackThomson, line 113: You'll probably want to handle this error, as it's critical. A panic might be a fair response if this returns an error - you cannot find the shape to use.

RedbackThomson (changes requested):
> `model.go` does not take into account any `output_wrapper_field_path` overrides when processing Status fields for a CRD

I think what you're really referring to here is that `model.go` doesn't reference `output_wrapper_field_path` when generating CRD fields as part of the `Create*` request. I believe the code does use it in `set_resource.sdk` for `Describe*`.

Were you able to find a way for the `adaptedSourceVarName` to use the correct path?

Also it would be great to have a unit test for using `GetOutputShape` with `Create*`. That is, generating a list of Spec and Status fields and making sure it unwraps correctly (such as in your case with `VpcEndpoint`).

pkg/generate/code/set_resource.go
brycahta: `GetOutputShape` has a panic for that scenario. I kept logic consistent with previous behavior when refactoring

brycahta:
> > `model.go` does not take into account any `output_wrapper_field_path` overrides when processing Status fields for a CRD
> 
> I think what you're really referring to here is that `model.go` doesn't reference `output_wrapper_field_path` when generating CRD fields as part of the `Create*` request. I believe the code does use it in `set_resource.sdk` for `Describe*`.

You're right, that's clearer. I'll update PR title

> 
> Were you able to find a way for the `adaptedSourceVarName` to use the correct path?

Yessir, prior to these changes `GetOutputShape` would still try to unwrap a struct if it was the only member ref; however, I removed this because then the function would no longer be returning an *OutputShape*, but the nested field instead. Also, if no overrides are specified, `GetOutputShape` will return the original output shape,`createOp.OutputRef.Shape`.

> Also it would be great to have a unit test for using `GetOutputShape` with `Create*`. That is, generating a list of Spec and Status fields and making sure it unwraps correctly (such as in your case with `VpcEndpoint`).

Good call, will add in next revision

pkg/generate/code/set_resource.go
vijtrip2: But the method does throw other errors apart from that panic.
I think it makes sense to catch error and panic
(OR ugly version would be to panic for all errors inside `GetOutputShape` and remove `error` from return type)

vijtrip2 (commented):
Looks good.

pkg/generate/code/set_resource.go
vijtrip2: Fair enough... 
Since in the existing code outputshape is being returned as nil and then it is being handled in line 114, we do not need the error in method signature.

I would be onboard with removing error from method signature and keep the panic statement in method body

RedbackThomson (approved):
Thanks!

/lgtm

jaypipes:
@brycahta in future, please do create a GH issue for tracking these kinds of issues and include in the GH issue a simple test that demonstrates the buggy behaviour.

pkg/generate/code/set_resource.go
```diff
@@ -110,23 +110,14 @@ func SetResource(
...
 	wrapperFieldPath := r.GetOutputWrapperFieldPath(op)
 	if wrapperFieldPath != nil {
-		outputShape, err = r.GetWrapperOutputShape(outputShape, *wrapperFieldPath)
-		if err != nil {
-			msg := fmt.Sprintf("Unable to unwrap the output shape: %v", err)
-			panic(msg)
-		}
 		sourceVarName += "." + *wrapperFieldPath
 	} else {
 		// If the wrapper field path is not specified in the config file and if
```
jaypipes, line 123: Sorry, but why didn't this block of code pick up the `VpcEndpoint` shape as being the wrapper struct?

brycahta: This code did work as intended; however, because `model.go:164` executes first and **was not** checking for overrides initially, the code-generator would crash because it could not locate `VpcEndpoint` identifier. 

I added `GetOutputShape` call here after refactoring to clean up the code a bit-- the logic remains the same.

RedbackThomson: I would like to clarify that the original use for `output_wrapper_field_path` was working: You could unwrap a describe output structure into status and spec fields. 

What we are adding in this is support for `output_wrapper_field_path` on a create output. That is, taking a wrapper structure from a create and creating new status and spec fields from its members. [CreateVpcEndpointOutput](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#CreateVpcEndpointOutput) has a wrapper of `VpcEndpoint` - its members are what we want in the spec and status for the CRD.

Some of the code changes @brycahta has made are re-using existing parts of the code that worked with the describe output to also work with create. I believe this is one of those elements.

PR #212, opened 2021-09-30
Export `GroupVersionResource` and `GroupKind` variables

RedbackThomson:
Fixes https://github.com/aws-controllers-k8s/community/issues/998

Description of changes:
`GroupVersionResource` is used by dynamic clients to specify the CRD to be CRUD'd.
`GroupKind` is sometimes used in place where the version is not necessarily specified. I chose to export this mainly because it could be used down the line by other dynamic clients.

### DIFF
--- a/templates/pkg/resource/descriptor.go.tpl
+++ b/templates/pkg/resource/descriptor.go.tpl
@@ -18,7 +18,8 @@ const (
 )
 
 var (
-	resourceGK = metav1.GroupKind{
+	GroupVersionResource = svcapitypes.GroupVersion.WithResource("{{ ToLower .CRD.Plural }}")
+	GroupKind = metav1.GroupKind{
 		Group: "{{ .APIGroup }}",
 		Kind:  "{{ .CRD.Kind }}",
 	}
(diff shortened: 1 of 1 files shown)


vijtrip2 (approved):

A-Hilaly (approved):
:+1:
/lgtm

templates/pkg/resource/descriptor.go.tpl
```diff
@@ -32,7 +33,7 @@ type resourceDescriptor struct {
 // GroupKind returns a Kubernetes metav1.GroupKind struct that describes the
 // API Group and Kind of CRs described by the descriptor
 func (d *resourceDescriptor) GroupKind() *metav1.GroupKind {
```
A-Hilaly, line 35: Do we still need this method?

jaypipes: Yes, because the runtime doesn't import the resource package itself (it loads the resource manager factory from a registry) but still needs to know the GroupKind:

https://github.com/aws-controllers-k8s/runtime/blob/8e4dc5ecc41f9370e413c3be4ee88baf67ecbcd1/pkg/runtime/reconciler.go#L65-L72

jaypipes (approved):
:+1:

PR #216, opened 2021-10-07
Load service model name from generator

RedbackThomson:
Fixes aws-controllers-k8s/community#994

Description of changes:
This pull request supports including an optional `model_name` field in the `generator.yaml` file. The generator will use this argument to override the service name when looking up the API files in aws-sdk-go/models/apis.

Currently `ServiceIDClean` is used in all code generator templates as the import path for aws-sdk-go and when referencing the controller name (eg. `{{ .ServiceIDClean}}-controller`). This pull request will remove `ServiceIDClean` and instead use the service alias (as `ServiceAlias`) when referencing the controller and the aws-sdk-go packages. 

### DIFF
--- a/cmd/ack-generate/command/common.go
+++ b/cmd/ack-generate/command/common.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"sort"
 	"strings"
 	"syscall"
 	"time"
@@ -30,6 +31,7 @@ import (
 	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/generate/config"
 	ackmodel "github.com/aws-controllers-k8s/code-generator/pkg/model"
 	"github.com/aws-controllers-k8s/code-generator/pkg/util"
+	k8sversion "k8s.io/apimachinery/pkg/version"
 )
 
 const (
@@ -211,9 +213,19 @@ func getSDKVersionFromGoMod(goModPath string) (string, error) {
 	return "", fmt.Errorf("couldn't find %s in the go.mod require block", sdkModule)
 }
 
+// loadModelWithLatestAPIVersion finds the AWS SDK for a given service alias and
+// creates a new model with the latest API version.
+func loadModelWithLatestAPIVersion(svcAlias string) (*ackmodel.Model, error) {
+	latestAPIVersion, err := getLatestAPIVersion()
+	if err != nil {
+		return nil, err
+	}
+	return loadModel(svcAlias, latestAPIVersion)
+}
+
 // loadModel finds the AWS SDK for a given service alias and creates a new model
-// with the latest API version.
-func loadModel(svcAlias string) (*ackmodel.Model, error) {
+// with the given API version.
+func loadModel(svcAlias string, apiVersion string) (*ackmodel.Model, error) {
 	cfg, err := ackgenconfig.New(optGeneratorConfigPath, ackgenerate.DefaultConfig)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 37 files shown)


cmd/ack-generate/command/common.go
```diff
@@ -207,3 +212,67 @@ func getSDKVersionFromGoMod(goModPath string) (string, error) {
 	}
 	return "", fmt.Errorf("couldn't find %s in the go.mod require block", sdkModule)
 }
+
+// loadModelWithLatestAPIVersion finds the AWS SDK for a given service alias and
```
jaypipes, line 216: this and the code below, considering (in a future PR if you agree with me) perhaps putting this into an importable package under `pkg/`

pkg/generate/config/config.go
```diff
@@ -39,6 +39,9 @@ type Config struct {
 	// SetManyOutput function fails with NotFound error.
 	// Default is "return nil, ackerr.NotFound"
 	SetManyOutputNotFoundErrReturn string `json:"set_many_output_notfound_err_return,omitempty"`
+	// ModelName lets you specify the path used to identify the AWS service API
+	// in the aws-sdk-go's models/apis/ directory
```
jaypipes, line 43: suggest adding in the following: ". This field is optional and only needed for services such as the opensearchservice service where the model name is opensearch and the service package is called opensearchservice."

jaypipes (approved):
Teeny suggestion if you re-spin, but I'm good with this as-is. Thank you @RedbackThomson!

@A-Hilaly @brycahta since Vijay is out today and tomorrow, please review and lgtm if all good with this.

cmd/ack-generate/command/common.go
```diff
@@ -207,3 +212,67 @@ func getSDKVersionFromGoMod(goModPath string) (string, error) {
...
+func loadModelWithLatestAPIVersion(svcAlias string) (*ackmodel.Model, error) {
+	latestAPIVersion, err := getLatestAPIVersion()
+	if err != nil {
+		return nil, err
+	}
+	return loadModel(svcAlias, latestAPIVersion)
+}
+
+// loadModel finds the AWS SDK for a given service alias and creates a new model
+// with the given API version.
```
brycahta, line 227: what if `apiVersion` is empty string?

cmd/ack-generate/command/release.go
```diff
@@ -74,21 +73,7 @@ func generateRelease(cmd *cobra.Command, args []string) error {
...
-		}
-		sdkAPI, err = sdkHelper.API(newSvcAlias) // retry with serviceID
-		if err != nil {
-			return fmt.Errorf("service %s not found", svcAlias)
-		}
-	}
-	m, err := ackmodel.New(
-		sdkAPI, "", optGeneratorConfigPath, ackgenerate.DefaultConfig,
-	)
+	m, err := loadModel(svcAlias, "")
```
brycahta, line 76: nit: could you add a comment as to why the apiVersion is empty here? If it doesn't matter, then `loadModelWithLatestAPIVersion` would be clearer

pkg/generate/config/config.go
```diff
@@ -39,6 +39,12 @@ type Config struct {
 	// SetManyOutput function fails with NotFound error.
 	// Default is "return nil, ackerr.NotFound"
 	SetManyOutputNotFoundErrReturn string `json:"set_many_output_notfound_err_return,omitempty"`
+	// ModelName lets you specify the path used to identify the AWS service API
```
brycahta, line 42: nit: apply your comment for `ServiceModelName` here-- I thought it was very clear!

`// ModelName lets you specify the exact string used to identify the AWS service API...`

pkg/model/sdk_helper.go
```diff
@@ -93,8 +93,8 @@ func (h *SDKHelper) WithAPIVersion(apiVersion string) {
 }
 
 // API returns the aws-sdk-go API model for a supplied service alias
```
brycahta, line 95: nit: comment should be `... for a supplied service model name`. Same for below. Or `model service alias` depending on above discussion

pkg/generate/config/config.go
```diff
@@ -39,6 +39,12 @@ type Config struct {
...
+	// in the aws-sdk-go's models/apis/ directory. This field is optional and
+	// only needed for services such as the opensearchservice service where the
+	// model name is `opensearch` and the service package is called
+	// `opensearchservice`.
+	ModelName string `json:"model_name,omitempty"`
```
brycahta, line 47: Any thoughts on refactoring this name to include "service alias" in some capacity? Ex: `ModelSvcAlias` I think it's clearer both in code and for the interface

brycahta (changes requested):

cmd/ack-generate/command/release.go
RedbackThomson: @A-Hilaly Would love your input on this.

PR #213, opened 2021-10-04
Consolidate service naming nomenclature

RedbackThomson:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/994

Description of changes:
This pull request replaces all templates and variables with a defined nomenclature. The current code-generator uses service name terms interchangeably, which eventually lead to the linked issue. While PR #211 solved the logical issue of overriding the service model name, in some cases, it did not do a clear job of clarifying what names exist for any service and when they are applicable.

There are 3 different names that are used to programmatically reference a service. Here are the terms I am using in this PR to refer to them, and their given purpose:
- Service ID
  - The `metadata.serviceID` field from the model `api-2.json` file
- Service model name
  - The path to the model `api-2.json` file
- Service alias
  - The AWS SDK Go package path
  - Used by the ACK controller's for their name 

Some examples on when these differ:
- Step Functions uses `sfn` for its ID and alias, but `states` for its model name
- Elastic Load Balancing v2 uses `elbv2` for its alias, but `elasticloadbalancingv2` for its ID and model name

There are also 2 names defined in the controller metadata file that are not used in any parts of the code, but are used in documentation:
- Service name
  - The full name of the service as defined by the product's documentation page (eg. `Amazon Elastic Cloud Compute`)
-  Service short name
  - An informal, shortened name of the service (eg. `EC2`)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #218, opened 2021-10-08
remove --aws-account-id flag from deployment

jaypipes:
ACK runtime v0.15.0 removed support for the `--aws-account-id`
controller flag but our deployment templates were still adding this flag
to the controller entrypoint args, which causes tests to fail with:

```
unknown flag: --aws-account-id
```

Related: #213

See: https://github.com/aws-controllers-k8s/runtime/commit/60801010475969b411d192574c9a47b5acc81bb9

### DIFF
--- a/templates/config/controller/deployment.yaml.tpl
+++ b/templates/config/controller/deployment.yaml.tpl
@@ -26,8 +26,6 @@ spec:
       - command:
         - ./bin/controller
         args:
-        - --aws-account-id
-        - "$(AWS_ACCOUNT_ID)"
         - --aws-region
         - "$(AWS_REGION)"
         - --enable-development-logging
--- a/templates/helm/templates/deployment.yaml
+++ b/templates/helm/templates/deployment.yaml
@@ -37,8 +37,6 @@ spec:
       - command:
         - ./bin/controller
         args:
-        - --aws-account-id
-        - "$(AWS_ACCOUNT_ID)"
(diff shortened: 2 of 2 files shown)


brycahta:
> @jaypipes @brycahta need to remove from helm charts https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/helm/values.yaml.tpl#L41

_Functionally_ , this is fine today right? I agree it should be cleaned up along with its use here: [test-helm.sh](https://github.com/aws-controllers-k8s/test-infra/blob/main/scripts/test-helm.sh#L104)

RedbackThomson:
> > @jaypipes @brycahta need to remove from helm charts https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/helm/values.yaml.tpl#L41
> 
> No, the environment variables are fine. It's just the command-line flag that was at issue.

We still should remove the Helm value and corresponding environment variable from the templates - they serve no purpose anymore

jaypipes:
> I don't believe setting the environment variable has any effect in the AWS SDK Go - https://github.com/aws/aws-sdk-go/search?q=AWS_ACCOUNT_ID ? Through what mechanism would it override those results?

Sorry, was confusing with AWS_ACCESS_KEY_ID: https://github.com/aws/aws-sdk-go/blob/main/aws/credentials/env_provider.go

PR #222, opened 2021-10-14
Create custom list and map fields

RedbackThomson:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/881

Description of changes:
Supports creating custom list and map fields where the member shapes of those fields already exist in the SDK.
Rather than supplying a `from` in the field config, you can now supply a `custom_field` property which requires
either `list_of` or `map_of` with the name of an API shape as the value.

An example of generating a list (for an S3 Bucket):
```yaml
resources:
  Bucket:
    fields:
      AnalyticsConfigurations:
        custom_field:
          list_of: AnalyticsConfiguration
```

Produces the following:
```golang
AnalyticsConfigurations []*AnalyticsConfiguration `json:"analyticsConfigurations,omitempty"`
```

An example of generating a map (for an S3 Bucket):
```yaml
resources:
  Bucket:
    fields:
      AnalyticsConfigurations:
        custom_field:
          map_of: AnalyticsConfiguration
```

Produces the following:
```golang
AnalyticsConfigurations map[string]*AnalyticsConfiguration `json:"analyticsConfigurations,omitempty"` 
```


pkg/sdk/helper.go
```diff
@@ -112,14 +122,24 @@ func (h *SDKHelper) API(serviceModelName string) (*SDKAPI, error) {
...
-		return &SDKAPI{api, nil, nil, h.APIGroupSuffix}, nil
+		sdkapi := &SDKAPI{
+			API:            api,
+			APIGroupSuffix: h.APIGroupSuffix,
+			CustomShapes:   []*CustomShape{},
+			opMap:          nil,
+			typeRenames:    nil,
+		}
+
+		h.InjectCustomShapes(sdkapi)
```
brycahta, line 133: From the current implementation of Helper, I don't think it makes sense for one of its methods to be able to change/edit `sdkapi`.

Is there any reason this couldn't be a setter? ex: `func (a *SDKAPI) SetCustomShapes(cfg ackgenconfig.Config)`

pkg/sdk/helper.go
```diff
@@ -219,6 +240,26 @@ func (a *SDKAPI) GetOperationMap(cfg *ackgenconfig.Config) *OperationMap {
 	return &opMap
 }
 
+// GetCustomListRef finds a ShapeRef for a supplied custom slice field
+func (a *SDKAPI) GetCustomListRef(memberShapeName string) (*awssdkmodel.ShapeRef, bool) {
```
brycahta, line 244: Thoughts on refactoring this to be a deeper method? ex:

```
func (a*SDKAPI) GetCustomShapeRef(shapeName string) (*awssdkmodel.ShapeRef, bool) {
    customShape, found := getCustomListRef(shapeName)
    if !found {
        customShape, found = getCustomMapRef(shapeName)
    }
    return customShape, found
}
```

Aligns with existing method names and easier to use for clients

pkg/sdk/custom_shapes.go
```diff
@@ -0,0 +1,191 @@
...
+		sdkapi.API.Shapes[customShape.Shape.ShapeName] = customShape.Shape
+		sdkapi.CustomShapes = append(sdkapi.CustomShapes, customShape)
+	}
+
+	return nil
+}
+
+// getCustomListFieldMembers finds all of the custom list fields that need to
+// be generated as defined in the generator config.
+func (h *Helper) getCustomListFieldMembers() []string {
```
brycahta, line 76: this and `getCustomMapFieldMembers` should be `Config` getters

pkg/sdk/helper.go
```diff
@@ -88,12 +98,12 @@ func (h *SDKHelper) WithSDKVersion(version string) error {
...
 
 // WithAPIVersion sets the `apiVersion` field.
-func (h *SDKHelper) WithAPIVersion(apiVersion string) {
+func (h *Helper) WithAPIVersion(apiVersion string) {
 	h.apiVersion = apiVersion
 }
 
 // API returns the aws-sdk-go API model for a supplied service model name.
-func (h *SDKHelper) API(serviceModelName string) (*SDKAPI, error) {
+func (h *Helper) API(serviceModelName string) (*SDKAPI, error) {
```
brycahta, line 106: I know you didn't change any logic here, but why have the helper do this instead of a constructor for `SDKAPI` requiring serviceModelName?

RedbackThomson: I don't have any history for this method - they predate my introduction to the project. My best guess is that we don't care about the service name anywhere else in the `Helper`, but I think over time we have expanded it and now we have an awkward waterfall of passing it between methods.

pkg/generate/ack/controller.go
```diff
@@ -22,6 +22,7 @@ import (
 	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/generate/config"
 	"github.com/aws-controllers-k8s/code-generator/pkg/generate/templateset"
 	ackmodel "github.com/aws-controllers-k8s/code-generator/pkg/model"
+	ackoperations "github.com/aws-controllers-k8s/code-generator/pkg/operations"
```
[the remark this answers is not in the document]
RedbackThomson: The import aliasing in ACK seems to be a bit random. Some places use `ackmodel` others import model without an alias. I have just stuck to whatever the file has as convention. I don't think `ackmodel` is necessary, because we are clearly in the `ack` project, so `model` should be sufficient.

pkg/model/model.go
```diff
@@ -137,25 +137,45 @@ func (m *Model) GetCRDs() ([]*CRD, error) {
...
-				panic(msg)
+				if !found {
+					// This is a compile-time failure, just bomb out...
+					msg := fmt.Sprintf(
+						"unknown additional Spec field with Op: %s and Path: %s",
+						from.Operation, from.Path,
+					)
+					panic(msg)
+				}
+			} else if fieldConfig.CustomField != nil {
```
brycahta, line 157: It seems like the custom field is being added to both Spec and Status-- is that intended? Am I misreading?

RedbackThomson:
> Is the only way to use this custom field via Hooks in service-controller code?

At the moment, yes. Considering the type doesn't exist in the SDK, the generator won't really know how to use it - it's not linked to any of the operations and therefore not linked to any SDK methods. This could be smarter in the future, but not in this revision.

PR #224, opened 2021-10-15
update olm script to remove AdoptedResource from the bundle that is generated

acornett21:
Issue #, if available:
aws-controllers-k8s/community#1006
This does not close this issue, this is just one piece needed. The runtime also needs to be update.

Description of changes:
This PR updates the `olm` scripts in a way that the `AdpotedResource` is no longer generated in the bundle that gets produced on disk. Since the OLM script works off a `temp` directory, this does not update/affect any of the artifacts on disk in the repo. The bundle that this generates has been tested against an OpenShift cluster vai OLM install. 

### DIFF
--- a/scripts/olm-create-bundle.sh
+++ b/scripts/olm-create-bundle.sh
@@ -128,8 +128,8 @@ kustomize edit set image controller="$AWS_SERVICE_DOCKER_IMG"
 popd 1>/dev/null
 
 # remove crd/common from bases to prevent inclusion of AdoptedResource CRD from being generated in the bundle directory
-sed -i '' '/bases:/d' $tmp_kustomize_config_dir/crd/kustomization.yaml
-sed -i '' '/- common/d' $tmp_kustomize_config_dir/crd/kustomization.yaml
+sed -i.orig '/^bases:$/d' $tmp_kustomize_config_dir/crd/kustomization.yaml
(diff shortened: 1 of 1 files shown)


scripts/olm-create-bundle.sh
```diff
@@ -127,6 +127,10 @@ pushd $tmp_kustomize_config_dir/controller 1>/dev/null
 kustomize edit set image controller="$AWS_SERVICE_DOCKER_IMG"
 popd 1>/dev/null
 
+# remove crd/common from bases to prevent inclusion of AdoptedResource CRD from being generated in the bundle directory
+sed -i '' '/bases:/d' $tmp_kustomize_config_dir/crd/kustomization.yaml
+sed -i '' '/- common/d' $tmp_kustomize_config_dir/crd/kustomization.yaml
```
vijtrip2, line 132: I may be very wrong here but I have never seen this usage of `sed` command before.

I ran these commands on a dummy file locally and these commands did not cause any changes in my dummy file on zsh shell.

Can you please tell me the environment setting, where I can test this command to validate that it works?

vijtrip2: The integration tests run on debian image https://github.com/aws-controllers-k8s/test-infra/blob/main/prow/jobs/images/Dockerfile.olm-test

Can you please make sure the sed command you are using works on debian platform to get s3-olm-test to pass. : )

PR #225, opened 2021-10-19
Added caution note regarding `endpoint_url`

cparik:
Issue #, if available: 1015

Description of changes:
  Commented out endpoint_url. 
  The endpoint_url is for the AWS services such as s3, dynamodb, vpc, etc. See the [AWS SDK documentation](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/) for more information.
  It can be easily confused with the EKS endpoint. Adding EKS endpoint url here results in issue 1015.

### DIFF
--- a/templates/helm/values.yaml.tpl
+++ b/templates/helm/values.yaml.tpl
@@ -38,7 +38,12 @@ resources:
 aws:
   # If specified, use the AWS region for AWS API calls
   region: ""
-  endpoint_url: ""
+
+  # The following endpoint_url is for the AWS services such as s3, dynamodb, vpc, etc. 
+  # It can be easily confused with the EKS endpoint. Adding EKS endpoint url here results in issue 1015. 
+  # It has been intentionally commented out. 
+
+#  endpoint_url: ""
 
 # log level for the controller
 log:


templates/helm/values.yaml.tpl
```diff
@@ -38,7 +38,12 @@ resources:
 aws:
   # If specified, use the AWS region for AWS API calls
   region: ""
-  endpoint_url: ""
+
+  # The following endpoint_url is for the AWS services such as s3, dynamodb, vpc, etc. 
+  # It can be easily confused with the EKS endpoint. Adding EKS endpoint url here results in issue 1015. 
+  # It has been intentionally commented out. 
+
+#  endpoint_url: ""
```
jaypipes, line 46: please remove the comment in the above line (to set the value back to `""` instead of unset)

templates/helm/values.yaml.tpl
```diff
@@ -38,7 +38,12 @@ resources:
 aws:
   # If specified, use the AWS region for AWS API calls
   region: ""
-  endpoint_url: ""
+
```
jaypipes, line 41: please remove the empty newlines before and after the code comment you added in order to align with the other code comments in the file.

templates/helm/values.yaml.tpl
```diff
@@ -38,7 +38,12 @@ resources:
...
+  # The following endpoint_url is for the AWS services such as s3, dynamodb, vpc, etc. 
+  # It can be easily confused with the EKS endpoint. Adding EKS endpoint url here results in issue 1015. 
+  # It has been intentionally commented out.
```
jaypipes, line 44: Suggest rewording to:

```
# If specified, use the custom service API endpoint URL.
# NOTE: Do *not* set this to the EKS endpoint URL if you are installing
# the controller into an EKS cluster.
# See: https://docs.aws.amazon.com/general/latest/gr/rande.html
# See: https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/
```

PR #232, opened 2021-10-26
allow different Go types for input/output shapes

jaypipes:
Adds a new SetFieldConfig struct to the code generator config that
allows teams to instruct the code generator to handle the SetResource
generator specially for a particular `pkg/model.OpType`.

This is useful for at least a couple use cases:

1. Different Go types in input and output shapes

When the Go type of a field in the Create Input shape differs from the
Go type of the same-named field in another operation's Output shape, the
code generator doesn't know how to handle this situation and ends up
outputting invalid Go code.

An example of this behaviour is in the RDS controller's DBInstance
resource. The DBSecurityGroups field in the Create Input shape is of
type `[]*string`, but the same field in the Create, ReadMany and Update
output shapes is of type `[]*DBSecurityGroupMembership`. The
`DBSecurityGroupMembership` struct has two fields, `DBSecurityGroupName`
and `Status`. The only part of that struct that is relevant to the
`Spec.DBSecurityGroups` field is the `DBSecurityGroupName` field, which
corresponds to the string identifier of the DB security group provided
by the user in the `CreateDBInstanceRequest.DBSecurityGroups` field.

The code generator ends up calling `pkg/generate/code.SetResource` for
the DBInstance resource, gets to the `Spec.DBSecurityGroups` field and
just doesn't know how to handle the different Go types and so ends up
outputting nothing, which breaks compilation and has caused us to
manually comment out or fix the generated code.  operation.

See: https://github.com/aws-controllers-k8s/rds-controller/blob/dc3271fa7b455fc99c3531ce372ea221c9f5b8e7/pkg/resource/db_instance/sdk.go#L853-L864

In the RDS DBInstance DBSecurityGroups case, we'd add the following to
the RDS generator.yaml file to instruct the code generator how to
transform the response from the DescribeDBInstances Output shape for
setting the value of `Spec.DBSecurityGroups` from the set of
`DBSecurityGroupMembership.DBSecurityGroupName` values:

```yaml
resources:
  DBInstance:
    fields:
      DBSecurityGroups:
        set:
         - on: Create
           from: DBSecurityGroupName
         - on: ReadOne
           from: DBSecurityGroupName
```

2. Ignoring fields in output shapes containing stale data

For some service APIs, the returned value for a field in the Output
shape of an Update operation contains the *originally-set* value instead
of the value of the field that was set in the Update operation itself.

An example of this situation is the ElastiCache ModifyReplicationGroup
API call. If you pass some value for the LogDeliveryConfiguration field
in the Input shape (which comes from the Spec.LogDeliveryConfiguration
field of the ReplicationGroup resource), the value that is returned in
the Output shape's LogDeliveryConfiguration is actually the *old*
(stale) value for the field. The reason for this return of stale data is
because the log delivery configuration is asynchronously mutated.

For these cases, we want to be able to tell the code generator to ignore
these fields when outputting code for the
`pkg/generate/code.SetResource` function, but only for the Update
operation. In this case, we'd add the following to the ElastiCache
generator.yaml file:

```yaml
resources:
  ReplicationGroup:
    fields:
      LogDeliveryConfiguration:
        set:
         - on: Update
           ignore: true
```

Issue: aws-controllers-k8s/community#178


pkg/generate/ack/controller.go
```diff
@@ -102,7 +103,18 @@ var (
...
 		},
 		"GoCodeSetResourceForStruct": func(r *ackmodel.CRD, targetFieldName string, targetVarName string, targetShapeRef *awssdkmodel.ShapeRef, sourceVarName string, sourceShapeRef *awssdkmodel.ShapeRef, indentLevel int) string {
-			return code.SetResourceForStruct(r.Config(), r, targetFieldName, targetVarName, targetShapeRef, sourceVarName, sourceShapeRef, indentLevel)
+			f, ok := r.Fields[targetFieldName]
+			if !ok {
+				return ""
+			}
+			// We may have some special instructions for how to handle setting the
+			// field value...
+			setCfg := f.GetSetterConfig(model.OpTypeList)
```
vijtrip2, line 112: Why `model.OpTypeList` as operation? 
I think adding a comment for the same will be useful as well.

pkg/model/field.go
```diff
@@ -63,6 +63,23 @@ func (f *Field) IsRequired() bool {
...
+// and a supplied operation type, or nil if none exists.
+func (f *Field) GetSetterConfig(opType OpType) *ackgenconfig.SetFieldConfig {
+	if f.FieldConfig == nil {
+		return nil
+	}
+	rmMethod := ResourceManagerMethodFromOpType(opType)
+	for _, setCfg := range f.FieldConfig.Set {
+		// If the Method attribute is nil, that means the setter config applies to
+		// all resource manager methods for this field.
+		if setCfg.Method == nil || strings.EqualFold(rmMethod, *setCfg.Method) {
```
vijtrip2, line 76: What should be the strategy if there is a setConfig with empty method name and one setConfig with OpType from parameter?

I think currently a single one will be returned depending on which is defined earlier. Should a priority be added for method match first and then fallback to empty method?

vijtrip2 (approved):

pkg/generate/code/set_resource.go
```diff
@@ -219,6 +206,53 @@ func SetResource(
...
+			//         set:
+			//          - method: ReadOne
+			//            from: Bucket
+			//
+			// And this would indicate to the code generator that the Spec.Name
+			// field should be set from the value of the Output shape's Bucket
+			// field.
+			var found bool
+			if setCfg != nil && setCfg.From != nil {
+				sourceMemberShapeRef, found = outputShape.MemberRefs[*setCfg.From]
```
RedbackThomson, line 241: Do we want to supported nested field paths here, as well?

pkg/generate/code/set_resource.go
```diff
@@ -1123,6 +1186,7 @@ func setResourceForContainer(
 			targetFieldName,
 			targetVarName,
 			targetShapeRef,
+			targetSetCfg,
 			sourceVarName,
 			sourceShapeRef,
 			indentLevel,
```
RedbackThomson, line 1192: Oh wow these are getting out of hand :/

RedbackThomson (approved):

pkg/generate/code/set_resource.go
```diff
@@ -1257,26 +1328,71 @@ func setResourceForSlice(
...
-		containerFieldName = targetFieldName
+	// There may be a generator.yaml file that looks like this:
+	//
+	// resources:
+	//   DBInstance:
+	//     fields:
+	//       DBSecurityGroups:
+	//         set:
+	//          - method: ReadOne
+	//            from: DBSecurityGroupName
```
brycahta, line 1352: nit: Not a huge fan of `DBSecurityGroups` being used to reference both CRD field and Output Shape field, but can see why you can't do a syntactically correct path for `from` value. Since you are resolving a type mismatch issue here, what if the setconfig `from` reflected the path to the value of a given _type_  -- not the exact field name in Output Shape? Something like:

```
	// resources:
	//   DBInstance:
	//     fields:
	//       DBSecurityGroups:
	//         set:
	//          - method: ReadOne
	//            from_type: DBSecurityGroupMembership.DBSecurityGroupName
```

brycahta (commented):
Thanks for this Jay; left some comments and will test this locally with dhcpOptions

pkg/generate/code/set_resource.go
brycahta: Yeah at first glance looks like we might be able to refactor by passing `Field` type in directly instead of each of its fields or maybe extend `Field` type with these methods

pkg/generate/code/set_resource.go
brycahta: +1, I'm fairly certain I will need nested field paths support for DhcpOptions. To summarize, DhcpOptions Spec has a _slice of_  `*NewDHCPConfiguration`:

```
type NewDHCPConfiguration struct {
	Key *string `json:"key,omitempty"`
	Values []*string `json:"values,omitempty"`
}
```

and the CreateDhcpOptions Result output shape returns a _slice of_ `*DHCPConfiguration`:

```
type AttributeValue struct {
	Value *string `json:"value,omitempty"`
}

type DHCPConfiguration struct {
	Key *string `json:"key,omitempty"`
	Values []*AttributeValue `json:"values,omitempty"`
}
```

I need entries in `NewDHCPConfiguration.Values` to be populated with data from  `DHCPConfiguration.Values.Value`

brycahta (approved):
Discussed offline; nested fields will be in a follow-up patch

PR #237, opened 2021-11-09
Use loadModelWithLatestAPIVersion to load models for crossplane

EdgeJ:
This fixes an issue where the crossplane command would fail to find API
models for resources whose name differed between the model and API in
the SDK. Previous behavior was to continue using the svcAlias instead of
correctly using model_name from the configuration.

The changes here also have the added benefit of making the generateCrossplane function code more similar to other generation functions and utilize more of the common codebase.

Issue #, if available:
N/A

Description of changes:
Replace duplicated code in the generateCrossplane function with implementations in common.go and change variables to match the naming scheme used elsewhere.

### DIFF
--- a/cmd/ack-generate/command/crossplane.go
+++ b/cmd/ack-generate/command/crossplane.go
@@ -35,9 +35,6 @@ var crossplaneCmd = &cobra.Command{
 }
 
 func init() {
-	crossplaneCmd.PersistentFlags().StringVar(
-		&optOutputPath, "provider-dir", ".", "the directory of the Crossplane provider",
-	)
 	rootCmd.AddCommand(crossplaneCmd)
 }
 
--- a/cmd/ack-generate/command/crossplane.go
+++ b/cmd/ack-generate/command/crossplane.go
@@ -17,7 +17,6 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
-	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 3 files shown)


jaypipes (changes requested):
All looks good except you should make a corresponding change to the README in the templates/crossplane/ directory here:

https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/crossplane/README.md

which mentions the `--provider-dir` CLI flag. :)

cmd/ack-generate/command/crossplane.go
```diff
@@ -39,11 +34,9 @@ var crossplaneCmd = &cobra.Command{
 	RunE:  generateCrossplane,
 }
 
-var providerDir string
-
 func init() {
 	crossplaneCmd.PersistentFlags().StringVar(
-		&providerDir, "provider-dir", ".", "the directory of the Crossplane provider",
+		&optOutputPath, "provider-dir", ".", "the directory of the Crossplane provider",
```
jaypipes, line 39: You can/should actually remove this flag entirely since the existing `--output` flag already covers its use:

https://github.com/aws-controllers-k8s/code-generator/blob/31204bc5278e40cf544135b27871fc67d2dc0118/cmd/ack-generate/command/root.go#L118-L120

jaypipes (changes requested):

cmd/ack-generate/command/crossplane.go
EdgeJ: Done.

jaypipes (approved):
Fantastic, thanks @EdgeJ!

PR #239, opened 2021-11-09
set `ACK.ResourceSynced` condition upon reconciler success/failure

vijtrip2:
Description of changes:
* I found that `ACK.ResourceSynced` condition is currently only being set using custom hook in very few service controllers or when `RequeueOnSuccessSeconds` value is non zero
* I think `ACK.ResourceSynced` condition should be present for each ACK resource telling it's sync status
* This PR adds the `ACK.ResourceSynced` condition support in common generated code to support all ACK resources

### DIFF
--- a/templates/pkg/resource/sdk.go.tpl
+++ b/templates/pkg/resource/sdk.go.tpl
@@ -260,18 +260,28 @@ func (rm *resourceManager) updateConditions (
 		}
 	}
 
-{{- if $reconcileRequeuOnSuccessSeconds := .CRD.ReconcileRequeuOnSuccessSeconds }}
-	if syncCondition == nil && onSuccess {
+	if syncCondition == nil {
 		syncCondition = &ackv1alpha1.Condition{
 			Type:   ackv1alpha1.ConditionTypeResourceSynced,
 			Status: corev1.ConditionTrue,
 		}
 		ko.Status.Conditions = append(ko.Status.Conditions, syncCondition)
 	}
-{{- else }}
-	// Required to avoid the "declared but not used" error in the default case
-	_ = syncCondition
-{{- end }}
+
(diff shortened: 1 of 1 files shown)


templates/pkg/resource/sdk.go.tpl
```diff
@@ -260,18 +260,28 @@ func (rm *resourceManager) updateConditions (
...
-{{- end }}
+
+	var syncMessage, syncReason string
+	if onSuccess {
+		syncCondition.Status = corev1.ConditionTrue
+		syncReason = "resource synced successfully"
+		syncMessage = "resource synced successfully"
+	} else {
+		syncCondition.Status = corev1.ConditionFalse
+		syncReason = "resource failed to sync"
```
surajkota, line 278: not sure if this is a good way to enforce this
this would rewrite the condition set by the service teams, we should set this only if the condition was null

what is the assumption here? all operations, create/delete/update should have this condition set while returning the resource? is there any code change to do that?

surajkota (changes requested):

templates/pkg/resource/sdk.go.tpl
vijtrip2: Good point. I will think about the overwriting part.
But only relying on null will not work because then condition will not update when set once.

Not "assumption" but my expectation is that this condition be present on all ACK resources indicating their sync status. This will serve as basis of when to resolve reference of an ACK resource in another resource.

jaypipes:
@vijtrip2 I was working on a more comprehensive solution to condition handling in https://github.com/aws-controllers-k8s/community/issues/1030. Mind if I take this one?

jaypipes:
@vijtrip2 please see https://github.com/aws-controllers-k8s/runtime/pull/61

PR #240, opened 2021-11-10
Use loadModel instead of loadModelWithLatestAPIVersion for crossplane

EdgeJ:
getLatestAPIVersion doesn't seem to work correctly with crossplane, and
opgGenVersion is already used to find the configuration path for
resources, so just use that instead.

Issue #, if available:

Description of changes:

### DIFF
--- a/cmd/ack-generate/command/crossplane.go
+++ b/cmd/ack-generate/command/crossplane.go
@@ -49,9 +49,7 @@ func generateCrossplane(_ *cobra.Command, args []string) error {
 	}
 	svcAlias := strings.ToLower(args[0])
 	optGeneratorConfigPath = filepath.Join(optOutputPath, "apis", svcAlias, optGenVersion, "generator-config.yaml")
-	m, err := loadModelWithLatestAPIVersion(
-		svcAlias,
-	)
+	m, err := loadModel(svcAlias, optGenVersion)
 	if err != nil {
 		return err
 	}


AaronME:
@EdgeJ - Thanks for the quick fix. Glad you were on top of it.

I'm still getting `v1beta1` resources with this commit, and resources in the "services.k8s.aws" apiGroup.

loadModel is using the default ackgenerate.DefaultConfig and not the one created for Crossplane in [pkg/generate/crossplane/](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/crossplane/config.go)

See https://github.com/aws-controllers-k8s/code-generator/blob/d921da46936c50e4942493ace0e7daad2a0c31df/cmd/ack-generate/command/crossplane.go#L81 for reference.

It looks like this was changed in https://github.com/aws-controllers-k8s/code-generator/commit/79b8dd711d65d440720c5232c946a6e806363cce#diff-6c9b11a8ab6c3f28f4ad6ee4d3f7f5b1c9da81f13ac772b15c784aa95f0ded10 and we haven't caught it until now.

loadModel needs to support passing in alternative sdkHelpers and defaultConfigs, or else we will need to duplicate the flow for the crossplane resources.

EdgeJ:
> @EdgeJ - Thanks for the quick fix. Glad you were on top of it.
> 
> I'm still getting `v1beta1` resources with this commit, and resources in the "services.k8s.aws" apiGroup.
> 
> loadModel is using the default ackgenerate.DefaultConfig and not the one created for Crossplane in [pkg/generate/crossplane/](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/crossplane/config.go)
> 
> See
> 
> https://github.com/aws-controllers-k8s/code-generator/blob/d921da46936c50e4942493ace0e7daad2a0c31df/cmd/ack-generate/command/crossplane.go#L81
> 
> for reference.
> It looks like this was changed in [79b8dd7#diff-6c9b11a8ab6c3f28f4ad6ee4d3f7f5b1c9da81f13ac772b15c784aa95f0ded10](https://github.com/aws-controllers-k8s/code-generator/commit/79b8dd711d65d440720c5232c946a6e806363cce#diff-6c9b11a8ab6c3f28f4ad6ee4d3f7f5b1c9da81f13ac772b15c784aa95f0ded10) and we haven't caught it until now.
> 
> loadModel needs to support passing in alternative sdkHelpers and defaultConfigs, or else we will need to duplicate the flow for the crossplane resources.

Thanks for the insight @AaronME , I didn't realize the commit adding the model name configuration wiped out some necessary configs for crossplane.  I don't think it should be too difficult to add that back into with this PR, but as you pointed out, we may just need to duplicate some of the functionality in `loadModel` with crossplane specific configuration.

EdgeJ:
Added some flags to the root command to be able to control what default configs to use (right now, only the ACK default and Crossplane default) and to dynamically set the API group name.  I _think_ this should take care of the issues we were seeing.  At any rate, tests out well on my machine doing dry run builds of some existing resources.

RedbackThomson:
> Added some flags to the root command to be able to control what default configs to use (right now, only the ACK default and Crossplane default) and to dynamically set the API group name. I _think_ this should take care of the issues we were seeing. At any rate, tests out well on my machine doing dry run builds of some existing resources.

I don't expect that anyone would ever use the ACK defaults with this crossplane generator - so I am happy with just hardcoding that and removing the option. As for the API group name, is this something crossplane would likely override?

EdgeJ:
> > Added some flags to the root command to be able to control what default configs to use (right now, only the ACK default and Crossplane default) and to dynamically set the API group name. I _think_ this should take care of the issues we were seeing. At any rate, tests out well on my machine doing dry run builds of some existing resources.
> 
> I don't expect that anyone would ever use the ACK defaults with this crossplane generator - so I am happy with just hardcoding that and removing the option. As for the API group name, is this something crossplane would likely override?

I don't think it's too likely.  I'll go ahead and remove those flags and just hard code the crossplane command to set the specific configurations in loadModel instead.

cmd/ack-generate/command/root.go
```diff
@@ -43,6 +43,8 @@ var (
 	optGeneratorConfigPath string
 	optMetadataConfigPath  string
 	optOutputPath          string
+	cpAPIGroupSuffix       string
+	cpDefaultCfg           bool
```
jaypipes, line 47: And you could remove these variables.

cmd/ack-generate/command/common.go
```diff
@@ -250,6 +259,11 @@ func loadModel(svcAlias string, apiVersion string) (*ackmodel.Model, error) {
 			return nil, fmt.Errorf("service %s not found", svcAlias)
 		}
 	}
+
+	if cpAPIGroupSuffix != "" {
+		sdkAPI.APIGroupSuffix = cpAPIGroupSuffix
+	}
```
jaypipes, line 265: Instead of the conditionals above, how about adding new `apiGroup` and `defaultCfg` parameters to the `loadModel` function?

cmd/ack-generate/command/crossplane.go
```diff
@@ -42,16 +42,20 @@ func generateCrossplane(_ *cobra.Command, args []string) error {
...
 	defer cancel()
 	if err := ensureSDKRepo(ctx, optCacheDir, optRefreshCache); err != nil {
 		return err
 	}
 	svcAlias := strings.ToLower(args[0])
 	optGeneratorConfigPath = filepath.Join(optOutputPath, "apis", svcAlias, optGenVersion, "generator-config.yaml")
-	m, err := loadModelWithLatestAPIVersion(
-		svcAlias,
-	)
+	m, err := loadModel(svcAlias, optGenVersion)
```
jaypipes, line 58: and then here you would do:

```
m, err := loadModel(svcAlias, optGenVersion, "aws.crossplane.io", cpconfig.DefaultConfig)
```

This would mean you would not have to import the `pkg/generate/config/crossplane` package within `cmd/ack-generate/command/common.go`.

jaypipes (commented):
A few suggestions inline...

cmd/ack-generate/command/common.go
EdgeJ: Yeah, this is the best approach.  I think I didn't consider this before because I was trying to touch as few places as possible, but it turns out only one other package uses this function outside the ones I've already changed.

RedbackThomson (approved):

jaypipes (approved):
Nice. Much cleaner! :)

PR #243, opened 2021-11-16
Remove code setting a resource ARN from its spec primary identifier

A-Hilaly:
Originally, this code was added to be able to make `ReadOne` calls when
a resource was not created yet and its ARN is not known, by trying to
build the ARN from the primary identifier field.

Currently `ReadOne` calls uses `requiredFieldsMissingFromReadOneInput`
to know whether a resource was created or not - if a required field is
missing, the `sdkFind` function will return a `ackerr.NotFound`

This also fixes code generation for lambda `CodeSigningConfig` resource
which was panicking because `CodeSIgningConfig` doesn't have a spec
identifier field.

Tested with dynamodb and ECR.


jaypipes (changes requested):
Please remove the code in the comment above on lines 197-198 :)

A-Hilaly:
@jaypipes done

jaypipes (approved):
tres bien, merci monsieur!

PR #246, opened 2021-11-24
code generation changes for resolving referenced resources

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1063

Description of changes:
* Code generation changes for supporting resource reference
* Corresponding runtime PR https://github.com/aws-controllers-k8s/runtime/pull/62
* Generated code can be seen in `pkg/generate/code/resource_reference_test.go` file


pkg/generate/code/resource_reference.go
```diff
@@ -0,0 +1,233 @@
...
+					out += readReferencedResource(field, strings.Repeat("\t",
+						fIndentLevel+1), apiReaderVarName, contextVarName,
+						resVarName)
+					out += checkReferencedResource(field, strings.Repeat("\t",
+						fIndentLevel+1))
+					out += fmt.Sprintf("%s\tko.Spec.%s = obj.%s\n",
+						fIndent, field.Names.Camel,
+						field.FieldConfig.References.Path)
+					out += fmt.Sprintf("%s}\n", fIndent)
+				}
```
jaypipes, line 158: I think it would actually be easier and cleaner to have this generator function be a *template* that produces a function that looks up a *single* resource via a resource reference wrapper field.

For example, imagine a template that looked like this:

```handlebars
{{ range $field := .CRD.Fields }}
{{ if $field.HasReference }}
// resolveReferenceFor{{ $field.Names.Camel }} resolves references for {{ .CRD.Kind }}.Spec.{{ $field.Names.Camel }}
// and returns any error that occurred attempting to resolve the reference
func (rm *resourceManager) resolveReferenceFor(
	ctx context.Context,
	apiReader client.Reader,
	ns types.Namespace,
	ref *ackv1alpha1.AWSResourceReferenceWrapper,
	res acktypes.AWSResource,
) error {
	if ref != nil && ref.From != nil {
		ko := rm.concreteResource(res)
		nsName := types.NamespacedName{Namespace: ns, Name: *ref.From.Name}
		obj := acksvcv1alpha1.{{ .CRD.Kind }}{}
		err := apiReader.Get(ctx, nsName, &obj)
		if err != nil {
			return err
		}
		if ackcond.ResourceSynced(res) {
			return ackerr.ReferencedResourceNotSynced(*ref.From.Name)
		}
		if obj.{{ $field.FieldConfig.References.Path }} == nil {
			return ackerr.ReferencedResourceNotSynced(*ref.From.Name)
		}
		ko.Spec.{{ $field.Names.Camel }} = obj.{{ $field.FieldConfig.References.Path }}
}

{{ end }}
```

jaypipes (changes requested):
Hmm, I have a number of concerns... I actually think that the majority (maybe all?) of the generator functions should be replaced with one or more template interpolations. See inline for comments...

pkg/generate/config/field.go
```diff
@@ -267,6 +267,31 @@ type LateInitializeConfig struct {
...
+// When 'APIIDRef' field is present in custom resource manifest, reconciler will
+// read the referred 'API' resource and copy the value from 'Status.APIID' in
+// 'Integration' resource's 'APIID' field
+type ReferencesConfig struct {
+	// Resource mentions the K8s resource which is read to resolve the
+	// reference
+	Resource string `json:"resource"`
+	// Path refers to the the path of field which should be copied
+	// to resolve the reference
+	Path string `json:"path"`
```
muvaf, line 292: In Crossplane, we run into cases where the referenced field needed to be parsed, hence we know have the concept of [extractor functions](https://github.com/crossplane/crossplane-runtime/blob/af4e148/pkg/reference/reference.go#L93) that you can write and point to in the [code generator](https://github.com/crossplane/crossplane-tools#reference-resolvers) to use if the [default one](https://github.com/crossplane/crossplane-runtime/blob/af4e148/pkg/reference/reference.go#L97) doesn't work for that case. I don't know if that'd be necessary for ACK but would definitely be helpful for Crossplane 🙂

pkg/generate/config/field.go
```diff
@@ -267,6 +267,31 @@ type LateInitializeConfig struct {
...
+//```
+// The above configuration will result in generation of a new field 'APIIDRef'
+// of type 'AWSResourceReference' for ApiGatewayv2-Integration crd.
+// When 'APIIDRef' field is present in custom resource manifest, reconciler will
+// read the referred 'API' resource and copy the value from 'Status.APIID' in
+// 'Integration' resource's 'APIID' field
+type ReferencesConfig struct {
+	// Resource mentions the K8s resource which is read to resolve the
+	// reference
+	Resource string `json:"resource"`
```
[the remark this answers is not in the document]
jaypipes: @muvaf yes, eventually that's exactly where we want to be! It's a fast follow feature for us.

pkg/generate/config/field.go
jaypipes: :+1: great idea. we can certainly add that functionality in a future enhancement

muvaf: Sounds good. FWIW, we'll definitely use this struct in Crossplane pipeline as well to automate the reference resolver generation 🙂  Great work!

pkg/generate/config/field.go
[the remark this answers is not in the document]
muvaf: Just something to keep in mind; in Crossplane we have all AWS CRDs in a single provider, hence, the provider that extracts the information from referenced field already has the necessary RBAC to read it. But ACK controllers are deployed separately, so, there might be some additional mechanisms needed to allow RDS controller to `GET` an EC2 resource.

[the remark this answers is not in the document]
jaypipes: Yup, @muvaf that's exactly why we are doing the cross-controller references in a future followup ;)

templates/pkg/resource/manager.go.tpl
```diff
@@ -16,19 +16,27 @@ import (
 	ackrequeue "github.com/aws-controllers-k8s/runtime/pkg/requeue"
 	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
 	acktypes "github.com/aws-controllers-k8s/runtime/pkg/types"
+	acksvcv1alpha1 "github.com/aws-controllers-k8s/{{ .ServicePackageName }}-controller/apis/v1alpha1"
```
jaypipes, line 19: @vijtrip2 what would you say to putting all of the resource reference generated code into a new file called `pkg/resource/{resource}/references.go`?

templates/pkg/resource/manager_read_referenced_resource.go.tpl
```diff
@@ -0,0 +1,49 @@
...
+                //          "{{ .FieldConfig.References.Resource }}",
+                //          namespace, *arr.Name)
+            }
+            if obj.{{ .FieldConfig.References.Path }} == nil {
+                return ackerr.ResourceReferenceMissingTargetFieldFor(
+                          "{{ .FieldConfig.References.Resource }}",
+                          namespace, *arr.Name,
+                          "{{ .FieldConfig.References.Path }}")
+            }
+{{- end -}}
```
jaypipes, line 49: :+1: 

though, if you take my suggestion above to put all the resource references code into `pkg/resource/{resource}/references.go`, you might want to rename this template to `templates/pkg/resource/references_read_referenced_resource.go.tpl`

templates/pkg/resource/manager.go.tpl
vijtrip2: Yes, I will incorporate it and send another revision.

pkg/generate/code/resource_reference.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+	}
+	return out
+}
+
+// ReferenceFieldsPresent produces go code(logical condition) for finding whether
+// a non-nil reference field is present in a resource. This checks helps in deciding
+// whether ACK.ReferencesResolved condition should be added to resource status
+// Sample Code:
+// return false || ko.Spec.APIIDRef != nil
```
jaypipes, line 74: Any reason why you aren't returning the `return` in this function?

pkg/generate/code/resource_reference.go
```diff
@@ -0,0 +1,87 @@
...
+				out += fmt.Sprintf("%s\treturn "+
+					"ackerr.ResourceReferenceOrIDRequiredFor(\"%s\", \"%s%s\")\n",
+					indent, field.Names.Camel, field.Names.Camel,
+					referenceFieldSuffix)
+				out += fmt.Sprintf("%s}\n", indent)
+			}
+		}
+	}
+	return out
+}
```
jaypipes, line 68: I am OK with the above but still think it's a bit too ACK-specific (with the ackerr error messages). I'd like eventually for these Go code generating functions to be smaller and more generic but won't hold this particular PR up for that long-term goal.

jaypipes (changes requested):
A few minor comments inline. I'm OK with this approach for right now to move forward and complete the resource references solution. Longer-term, however, I think we will want more composable and reusable Go code generating functions.

pkg/model/crd.go
```diff
@@ -757,6 +757,17 @@ func (r *CRD) HasMember(
 	return inSpec, inStatus
 }
 
+// HasReferenceFields returns true if any of the fields in CRD is a reference
+// field. Otherwise returns false
+func (r *CRD) HasReferenceFields() bool {
+	for _, fields := range r.Fields {
+		if fields.IsReference() {
```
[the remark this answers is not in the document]
vijtrip2: No, :) . 
Since this is a method on CRD, i.e. whether CRD "HasReferenceFields", so we iterate over each field to see if there "IsReference" field or not.

pkg/generate/code/resource_reference.go
vijtrip2: I did not include the `return` statement for reusability of this code-generator function.

>My thought process was that if cross-plane wants to leverage this condition, they may use it in `if` block, and the `return` will make it useless for them.

jaypipes (approved):
alrighty, let's do this.

PR #247, opened 2021-11-29
Generate resource `IsSynced` functions

A-Hilaly:
Issue: Generate ACK.ResourceSynced handlers #1064

Currently all the controllers contains custom code that checks whether
a resource is synced or not. This code is mostly used by generic hooks
that sets the `ACK.ResourceSynced` to its correct value. [Generic example
hook](https://github.com/aws-controllers-k8s/rds-controller/blob/main/templates/hooks/db_cluster/sdk_read_many_post_set_output.go.tpl)

This patch adds a new configuration option that will allow controller
maintainers to generate `resource.IsSynced`functions based on some
conditional.

configuration example:

```yaml
resources:
  Function:
    synced:
      when:
        - path: Status.State
          is: AVAILABLE
        - path: Status.State
          in:
            - AVAILABLE
            - ACTIVE
```

generated code example:

```go
// IsSynced returns true if the supplied resource is synced.
func (r *resource) IsSynced() bool {

	if *r.ko.Status.State != "AVAILABLE" {
		return false
	}
	candidates1 := []string{"AVAILABLE", "ACTIVE"}
	if !ackutil.InStrings(*r.ko.Status.State, candidates1) {
		return false
	}

	return true
}

```

### DIFF
--- a/pkg/generate/code/synced.go
+++ b/pkg/generate/code/synced.go
@@ -1,0 +1,204 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package code
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/aws-controllers-k8s/code-generator/pkg/fieldpath"
+	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/generate/config"
+	"github.com/aws-controllers-k8s/code-generator/pkg/model"
+
+	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
+)
+
+// ResourceIsSynced returns the Go code that verifies whether a resource is synced or
+// not. This code is generated using ack-generate configuration.
+// See ack-generate/pkg/config.SyncedConfiguration.
+//
+//  Sample output:
+//
+//  	candidates0 := []string{"AVAILABLE", "ACTIVE"}
+//  	if !ackutil.InStrings(*r.ko.Status.TableStatus, candidates0) {
+//  		return false, nil
+//  	}
+//  	if r.ko.Spec.ProvisionedThroughput == nil {
+//  		return false, nil
+//  	}
+//  	if r.ko.Spec.ProvisionedThroughput.ReadCapacityUnits == nil {
+//  		return false, nil
+//  	}
+//  	candidates1 := []int{0, 10}
+//  	if !ackutil.InStrings(*r.ko.Spec.ProvisionedThroughput.ReadCapacityUnits, candidates1) {
+//  		return false, nil
+//  	}
+//  	candidates2 := []int{0}
+//  	if !ackutil.InStrings(*r.ko.Status.ItemCount, candidates2) {
+//  		return false, nil
+//  	}
+func ResourceIsSynced(
+	cfg *ackgenconfig.Config,
+	r *model.CRD,
+	// String
+	resVarName string,
+	// Number of levels of indentation to use
+	indentLevel int,
+) string {
+	out := "\n"
+	resConfig, ok := cfg.ResourceConfig(r.Names.Original)
+	if !ok || resConfig.Synced == nil || len(resConfig.Synced.When) == 0 {
+		return out
+	}
+
+	for _, condition := range resConfig.Synced.When {
+		if condition.Path == nil || *condition.Path == "" {
+			panic("Received an empty sync condition path. 'SyncCondition.Path' must be provided.")
+		}
+		if len(condition.In) == 0 {
+			panic("'SyncCondition.In' must be provided.")
+		}
+		fp := fieldpath.FromString(*condition.Path)
+		field, err := getTopLevelField(r, *condition.Path)
+		if err != nil {
+			msg := fmt.Sprintf("cannot find top level field of path '%s': %v", *condition.Path, err)
+			panic(msg)
+		}
+		candidatesVarName := fmt.Sprintf("%sCandidates", field.Names.CamelLower)
+		if fp.Size() == 2 {
+			out += scalarFieldEqual(resVarName, candidatesVarName, field.ShapeRef.GoTypeElem(), condition)
+		} else {
+			out += fieldPathSafeEqual(resVarName, candidatesVarName, field, condition)
+		}
+	}
+
+	return out
+}
+
+func getTopLevelField(r *model.CRD, fieldPath string) (*model.Field, error) {
+	fp := fieldpath.FromString(fieldPath)
+	if fp.Size() < 2 {
+		return nil, fmt.Errorf("fieldPath must contain at least two elements, received: %s", fieldPath)
+	}
+
+	head := fp.PopFront()
+	fieldName := fp.PopFront()
+	switch head {
+	case "Spec":
+		field, ok := r.Fields[fieldName]
+		if !ok {
+			return nil, fmt.Errorf("field not found in Spec: %v", fieldName)
+		}
+		return field, nil
+	case "Status":
+		field, ok := r.Fields[fieldName]
+		if !ok {
+			return nil, fmt.Errorf("field not found in Status: %v", fieldName)
+		}
+		return field, nil
+	default:
+		return nil, fmt.Errorf("fieldPath must start with 'Spec' or 'Status', received: %v", head)
+	}
+}
+
+// scalarFieldEqual returns Go code that compares a scalar field to a given set of values.
+func scalarFieldEqual(resVarName string, candidatesVarName string, goType string, condition ackgenconfig.SyncedCondition) string {
+	out := ""
+	fieldPath := fmt.Sprintf("%s.%s", resVarName, *condition.Path)
+
+	valuesSlice := ""
+	switch goType {
+	case "string":
+		// []string{"AVAILABLE", "ACTIVE"}
+		valuesSlice = fmt.Sprintf("[]string{\"%s\"}", strings.Join(condition.In, "\", \""))
+	case "int64", "PositiveLongObject", "Long":
+		// []int64{1, 2}
+		valuesSlice = fmt.Sprintf("[]int{%s}", strings.Join(condition.In, ", "))
+	case "bool":
+		// []bool{false}
+		valuesSlice = fmt.Sprintf("[]bool{%s}", condition.In)
+	default:
+		panic("not supported type " + goType)
+	}
+
+	// candidates1 := []string{"AVAILABLE", "ACTIVE"}
+	out += fmt.Sprintf(
+		"\t%s := %v\n",
+		candidatesVarName,
+		valuesSlice,
+	)
+	// 	if !ackutil.InStrings(*r.ko.Status.State, candidates1) {
+	out += fmt.Sprintf(
+		"\tif !ackutil.InStrings(*%s, %s) {\n",
+		fieldPath,
+		candidatesVarName,
+	)
+
+	// return false, nil
+	out += "\t\treturn false, nil\n"
+	// }
+	out += "\t}\n"
+	return out
+}
+
+// fieldPathSafeEqual returns go code that safely compares a resource field to value
+func fieldPathSafeEqual(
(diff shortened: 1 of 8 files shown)


pkg/generate/code/synced.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"fmt"
+	"strings"
+
+	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/generate/config"
+	"github.com/aws-controllers-k8s/code-generator/pkg/model"
+)
+
+// ResourceIsSynced returns the Go code that verifies whether a resource is synced or
+// not. This code generated used ack-generate configuration.
```
vijtrip2, line 25: nit:
```suggestion
// not. This code is generated using ack-generate configuration.
```

vijtrip2 (commented):
/approve

Good work and very clean code! 
This is a good base on which we can further build more functionality for determining synced state.
One attribute that I can see being really helpful is 'non-empty' i.e. Checking whether a field exists or not to determine synced state.

A-Hilaly:
> This is a good base on which we can further build more functionality for determining synced state.

Indeed!

> One attribute that I can see being really helpful is 'non-empty' i.e. Checking whether a field exists or not to determine synced state.

That's a good idea! i'll add it. Thinking this would be useful for resources that don't have a mutating state, and only need an ARN field

pkg/generate/config/resource.go
```diff
@@ -87,6 +90,26 @@ type ResourceConfig struct {
...
+	When []SyncedCondition `json:"when"`
+}
+
+// SyncedCondition represent one of the unique condition that should be fullfiled in
+// order to assert whether a resource is synced.
+type SyncedCondition struct {
+	// Path of the field. e.g Status.Processing
+	Path *string `json:"path"`
+	// Is contains the exact value `Path` should be equal to.
+	Is string `json:"is"`
```
A-Hilaly, line 107: I think i will change this type to `interface{}` so we can support `bool` and `int64` types as well

A-Hilaly: Mmmm maybe it should more have a `GoType` field. /cc @vijtrip2

vijtrip2: Can the string comparison not cover bool and int ?
What I mean is you can convert the value from `Path` into string and then compare with value from `Is`?

A-Hilaly: Mmmmm yes that is possible using `json.Marshall` or `strings.Atoi/Iota`, but I think that it would be easier to just detect the Path `shape.GoType` and generate code with simple `==` tokens

templates/pkg/resource/resource.go.tpl
```diff
@@ -88,3 +90,15 @@ func (r *resource) DeepCopy() acktypes.AWSResource {
 	koCopy := r.ko.DeepCopy()
 	return &resource{koCopy}
 }
+
+// IsSynced returns true if the resource is synced.
+func (r *resource) IsSynced() bool {
```
jaypipes, line 95: Couple things...

a) You should not add a new exported method to Resource without a corresponding change to the AWSResource interface in the runtime repo.

b) I actually do not think that the AWSResource interface is the right interface to modify, because there are some controllers that may need to call alternate AWS APIs to determine whether a resource is synced. And because of that need to potentially call an AWS API, you will need access to the aws-sdk-go SDKAPI object and that is in the AWSResourceManager interface (the one that implements ReadOne, Create, Update, Delete, etc).

pkg/generate/ack/hook.go
```diff
@@ -110,6 +112,11 @@ readOne call inside AWSResourceManager.LateInitialize() method
 The "late_initialize_post_read_one" hooks are called AFTER making the
 readOne call inside AWSResourceManager.LateInitialize() method
 
+The "is_synced_pre_check" hooks are called BEFORE the code that checks whether
+a resource is sync or not.
+
+The "is_synced_post_check" hooks are called AFTER the code that checks whether
+a resource is sync or not.
```
jaypipes, line 119: Let's leave these hook points out for the time being. I'm not yet convinced we will need them.

templates/pkg/resource/resource.go.tpl
```diff
@@ -88,3 +90,15 @@ func (r *resource) DeepCopy() acktypes.AWSResource {
...
+{{- if $hookCode := Hook .CRD "is_synced_pre_check" }}
+{{ $hookCode }}
+{{- end }}
+{{ GoCodeIsSynced .CRD "r.ko" 1}}
+{{- if $hookCode := Hook .CRD "is_synced_post_check" }}
+{{ $hookCode }}
+{{- end }}
+	return true
```
jaypipes, line 103: The default return should probably be `false`, no?

pkg/generate/config/resource.go
jaypipes, line 107: I know I'd originally suggested both an `is` and an `in` field, but thinking about this more, we only really need an `in` field. The `is` implementation is simply a single-element list of `in`...

jaypipes (changes requested):

templates/pkg/resource/resource.go.tpl
A-Hilaly: > b) I actually do not think that the AWSResource interface is the right interface to modify, because there are some controllers that may need to call alternate AWS APIs to determine whether a resource is synced. And because of that need to potentially call an AWS API, you will need access to the aws-sdk-go SDKAPI object and that is in the AWSResourceManager interface (the one that implements ReadOne, Create, Update, Delete, etc).

Fair points, I initially added `IsSynced` to the `AWSResourceDescriptor` interface, but I eventually realized that `resourceDescriptor` doesn't have access to the `Spec/Status` field of an object.
What do you think about the signature of `IsSynced(context.Context, AWSResource) bool` as part of the `AWSResourceManager`

jaypipes: > What do you think about the signature of IsSynced(context.Context, AWSResource) bool as part of the AWSResourceManager

++ @A-Hilaly

jaypipes:
@A-Hilaly what's the status on this PR? Need assistance or further reviews?

A-Hilaly:
> @A-Hilaly what's the status on this PR? Need assistance or further reviews?

No reviews required for this commit. I'll force push the new approach.

jaypipes:
> > @A-Hilaly what's the status on this PR? Need assistance or further reviews?
> 
> No reviews required for this commit. I'll force push the new approach.

Roger.

pkg/generate/config/resource.go
A-Hilaly: Kept the string type, and try to detect the type from the field path instead

templates/pkg/resource/manager.go.tpl
```diff
@@ -234,6 +239,17 @@ func (rm *resourceManager) lateInitializeFromReadOneOutput(
...
 
+// IsSynced returns true if the resource is synced.
+func (rm *resourceManager) IsSynced(ctx context.Context, res acktypes.AWSResource) (bool, error) {
+	r := rm.concreteResource(res)
+	if r.ko == nil {
+		// Should never happen... if it does, it's buggy code.
+		panic("resource manager's IsSynced() method received resource with nil CR object")
+	}
+{{ GoCodeIsSynced .CRD "r.ko" 1}}
+	return false, nil
```
vijtrip2, line 250: How about default being `true,nil`

I suggest this because I can see this method being called after successful `create/update + late-initialize` call to determine the synced status. In case of majority of AWS resources which have synchronous create operations, this default true will set Synced condition correctly and they will not need to explicitly add Synced conditions in their generator.yaml file.

Only issue with above approach will be the asynchronous create/update calls, which are minority of total AWS resources BTW. For those resources service teams can get rid of their Synced Condition hooks and add synced Config instead.

What do you think?

pkg/generate/code/synced.go
```diff
@@ -0,0 +1,203 @@
...
+) string {
+	out := "\n"
+	resConfig, ok := cfg.ResourceConfig(r.Names.Original)
+	if !ok || resConfig.Synced == nil || len(resConfig.Synced.When) == 0 {
+		return out
+	}
+
+	for index, condition := range resConfig.Synced.When {
+		if condition.Path == nil || *condition.Path == "" {
+			panic("Received an empty sync condition path. SyncCondition.Path must be provided.")
```
vijtrip2, line 67: nit:
```suggestion
			panic("Received an empty sync condition path. 'SyncCondition.Path' must be provided.")
```

pkg/generate/code/synced.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+		if len(condition.In) == 0 {
+			panic("SyncCondition.In must be provided.")
```
vijtrip2, line 70: nit: 
```suggestion
			panic("'SyncCondition.In' must be provided.")
```

pkg/generate/code/synced.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+		fp := fieldpath.FromString(*condition.Path)
+		field, err := getTopLevelField(r, *condition.Path)
+		if err != nil {
+			panic("err:" + err.Error())
+		}
+		candidatesVarName := fmt.Sprintf("candidates%d", index)
```
vijtrip2, line 77: nit: 
```suggestion
		candidatesVarName := fmt.Sprintf("%sCandidates", field.Names.CamelLower)
```

pkg/generate/code/synced.go
```diff
@@ -0,0 +1,203 @@
...
+		valuesSlice,
+	)
+	// 	if !ackutil.InStrings(*r.ko.Status.State, candidates1) {
+	out += fmt.Sprintf(
+		"\tif !ackutil.InStrings(*%s, %s) {\n",
+		fieldPath,
+		candidatesVarName,
+	)
+
+	// return true
```
vijtrip2, line 147: outdated comment

pkg/generate/code/synced.go
```diff
@@ -0,0 +1,203 @@
...
+		// return false, nil
+		out += "\t\treturn false, nil\n"
+		// }
+		out += "\t}\n"
+	}
+	out += scalarFieldEqual(resVarName, candidatesVarName, shapes[len(shapes)-1].GoTypeElem(), condition)
+	return out
+}
+
+func fieldPathContainsMapOrArray(fieldPath string, shapeRef *awssdkmodel.ShapeRef) bool {
```
vijtrip2, line 188: I think this is used right now, correct?

vijtrip2 (commented):
Looks good over all. Just few comments : )
Good work!

templates/pkg/resource/manager.go.tpl
```diff
@@ -234,6 +239,17 @@ func (rm *resourceManager) lateInitializeFromReadOneOutput(
 {{ GoCodeLateInitializeFromReadOne .CRD "observed" "latest" 1 }}
 }
 
+// IsSynced returns true if the resource is synced.
+func (rm *resourceManager) IsSynced(ctx context.Context, res acktypes.AWSResource) (bool, error) {
+	r := rm.concreteResource(res)
+	if r.ko == nil {
+		// Should never happen... if it does, it's buggy code.
+		panic("resource manager's IsSynced() method received resource with nil CR object")
```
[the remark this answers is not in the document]
A-Hilaly: The reconciler function catches panics and recovers from them, other methods like `ReadOne`, `Create`, ... use the same pattern.

templates/pkg/resource/manager.go.tpl
A-Hilaly: >this default true will set Synced condition correctly and they will not need to explicitly add Synced conditions in their generator.yaml file.

This makes a lot of sense actually. I can make the `return false, nil` part of the `GoCodeIsSynced` generation

vijtrip2 (commented):
/lgtm

Looks good to me. Happy to merge this. 
Please raise another PR for runtime. I had to revert your commit.

pkg/generate/code/synced.go
```diff
@@ -0,0 +1,204 @@
...
+func getTopLevelField(r *model.CRD, fieldPath string) (*model.Field, error) {
+	fp := fieldpath.FromString(fieldPath)
+	if fp.Size() < 2 {
+		return nil, fmt.Errorf("fieldPath must contain at least two elements, received: %s", fieldPath)
+	}
+
+	head := fp.PopFront()
+	fieldName := fp.PopFront()
+	switch head {
+	case "Spec":
```
jaypipes, line 98: Note that this hard-coding makes Crossplane unable to use any of this... because their top-level fields are under `Spec.ForProvider` which is specified in the [`PrefixConfig`](https://github.com/aws-controllers-k8s/code-generator/blob/6da05cd784c388f3b113778d034960398c2dd56e/pkg/generate/config/config.go#L34) of the generator configuration. This is one of those places we need to be aware of when refactoring the code generator. /cc @brycahta

PR #248, opened 2021-12-01
introduce pkg/fieldpath for processing field paths

jaypipes:
In attempting to implement the nested field path support for different
source and target field Go types, I realized I needed a support package
that made working with field paths easier, including taking a field path
and finding a ShapeRef within an Output shape that matched the field
path.

This patch introduces a new `pkg/fieldpath` package with a simple `Path`
struct containing a variety of utility methods, as shown in this example
Go code snippet:

```go
import (
        "fmt"

	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
        "github.com/aws-controllers-k8s/code-generator/pkg/fieldpath"
        "github.com/aws-controllers-k8s/code-generator/pkg/sdk"
)

...

p := fieldpath.FromString("Crawler.Schedule.ScheduleExpression")

// Will print "true"
fmt.Println(p.HasPrefix("Crawler.Schedule"))

// Will print "false"
fmt.Println(p.HasPrefix("crawler.Schedule"))

// Will print "true"
fmt.Println(p.HasPrefixFold("crawler.Schedule"))

// Will print "ScheduleExpression"
fmt.Println(p.Back())

// Will print "Crawler"
fmt.Println(p.Front())

// Find the shape within the GetCrawlerResponse shape that contains the
// field referred to by the path...
helper := sdk.NewHelper("/path/to/apis", genCfg)
api := helper.API("glue")
crawlerOutputShape := api.Shapes["GetCrawlerResponse"]

schedExpressionShape := p.ShapeRef(crawlerOutputShape)
// Will print "ScheduleExpression"
fmt.Println(schedExpressionShape.ShapeName)
```
Note that we support list and map type field paths using a single-dot
notation since we are only looking at field types and not **values**
within the list or map.

Related: aws-controllers-k8s/community#1078

### DIFF
--- a/pkg/fieldpath/path.go
+++ b/pkg/fieldpath/path.go
@@ -1,0 +1,243 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package fieldpath
+
+import (
+	"encoding/json"
+	"strings"
+
+	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
+)
+
+// Path provides a JSONPath-like struct and field-member "route" to a
+// particular field within a resource. Path implements json.Marshaler
+// interface.
+type Path struct {
+	parts []string
+}
+
+// String returns the dotted-notation representation of the Path
+func (p *Path) String() string {
+	return strings.Join(p.parts, ".")
+}
+
+// MarshalJSON returns the JSON encoding of a Path object.
+func (p *Path) MarshalJSON() ([]byte, error) {
+	// Since json.Marshal doesn't encode unexported struct fields we have to
+	// copy the Path instance into a new struct object with exported fields.
+	// See https://github.com/aws-controllers-k8s/community/issues/772
+	return json.Marshal(
+		struct {
+			Parts []string
+		}{
+			p.parts,
+		},
+	)
+}
+
+// Pop removes the last part from the Path and returns it.
+func (p *Path) Pop() (part string) {
+	if len(p.parts) > 0 {
+		part = p.parts[len(p.parts)-1]
+		p.parts = p.parts[:len(p.parts)-1]
+	}
+	return part
+}
+
+// Front returns the first part of the Path or empty string if the Path has no
+// parts.
+func (p *Path) Front() string {
+	if len(p.parts) == 0 {
+		return ""
+	}
+	return p.parts[0]
+}
+
+// PopFront removes the first part of the Path and returns it.
+func (p *Path) PopFront() (part string) {
+	if len(p.parts) > 0 {
+		part = p.parts[0]
+		p.parts = p.parts[1:]
+	}
+	return part
+}
+
+// Back returns the last part of the Path or empty string if the Path has no
+// parts.
+func (p *Path) Back() string {
+	if len(p.parts) == 0 {
+		return ""
+	}
+	return p.parts[len(p.parts)-1]
+}
+
+// PushBack adds a new part to the end of the Path.
+func (p *Path) PushBack(part string) {
+	p.parts = append(p.parts, part)
+}
+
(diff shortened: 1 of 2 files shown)


muvaf:
Hey @jaypipes ! FYI, we have been using the [`fieldpath`](https://github.com/crossplane/crossplane-runtime/blob/47bff13/pkg/fieldpath/paved.go#L201) package that can work with arbitrary JSONs from crossplane-runtime for a long while now whenever we need this functionality. It has support for indexed accesses as well, like `Object.AField[0].AnotherField`. It could be helpful here 🙂

pkg/fieldpath/path.go
```diff
@@ -0,0 +1,243 @@
...
+
+// PushBack adds a new part to the end of the Path.
+func (p *Path) PushBack(part string) {
+	p.parts = append(p.parts, part)
+}
+
+// Copy returns a new Path that is a copy of this Path
+func (p *Path) Copy() *Path {
+	return &Path{p.parts}
+}
```
A-Hilaly, line 93: This function doesn't really copy `p.parts`. Since go arrays are pointers, using `Copy` here will not instantiate a new array, it will just reference the original one.

Did you mean:

```go
    return &Path{append([]string{}, p.parts...)}
```

A-Hilaly (commented):
Neat!
Left one comment below

pkg/fieldpath/path.go
jaypipes: This is actually what I intended. See, I needed the `Copy` function in the `ShapeRef` function in order to call `PopFront` but not modify the original Path:

```go
	cp := p.Copy()
	cur := cp.PopFront()
	if compare.ShapeName != cur {
		return nil
	}
	// And then we walk through the path, searching through the supplied
	// ShapeRef for a member ShapeRef matching each path element.
	for !cp.Empty() {
		cur = cp.PopFront()
		if compare = memberShapeRef(compare, cur); compare == nil {
			return nil
		}
	}
```

In the latest revision, I added a test that ensures the Path isn't modified when calling `Path.ShapeRef`.

jaypipes:
> Hey @jaypipes ! FYI, we have been using the [`fieldpath`](https://github.com/crossplane/crossplane-runtime/blob/47bff13/pkg/fieldpath/paved.go#L201) package that can work with arbitrary JSONs from crossplane-runtime for a long while now whenever we need this functionality. It has support for indexed accesses as well, like `Object.AField[0].AnotherField`. It could be helpful here slightly_smiling_face

Hi @muvaf! Yes, I actually looked at CP's fieldpath package as well as k8s.io/structured-merge-diff/fieldpath package for inspiration here! :)

I ended up going this route for the *field definition* paths. In the code generator, we don't need to refer to the *values* that exist in the fields. Rather, we just need to refer to the field *definitions*. I may very well use the CP fieldpath package for referencing those field's *values* from within ACK's runtime repo (which has a pkg/compare/path.go file that could likely be replaced with CP's fieldpath package.

A-Hilaly:
@jaypipes ecr-controller tests might be failing in the upcoming hours, @vijtrip2 and I are trying to fix an issue with release/soak tests for ecr-controller. Sorry for the inconvenience

A-Hilaly (approved):
:+1:

pkg/fieldpath/path.go
```diff
@@ -0,0 +1,243 @@
...
+
+// Contains returns true if the supplied string, delimited on ".", matches
+// p.parts up to the length of the supplied string.
+// e.g. if the Path p represents "A.B":
+//  subject "A" -> true
+//  subject "A.B" -> true
+//  subject "A.B.C" -> false
+//  subject "B" -> false
+//  subject "A.C" -> false
+func (p *Path) Contains(subject string) bool {
```
vijtrip2, line 206: nit: IMO "Contains" implies a substring search, however this is more like prefix search. How about naming the method "BeginsWith" ?

pkg/fieldpath/path.go
```diff
@@ -0,0 +1,243 @@
...
+		if p.parts[i] != s {
+			return false
+		}
+	}
+
+	return true
+}
+
+// ContainsFold is the same as Contains but uses case-insensitive comparisons
+func (p *Path) ContainsFold(subject string) bool {
```
vijtrip2, line 223: Same as above

vijtrip2 (approved):

pkg/fieldpath/path.go
jaypipes: > 

@vijtrip2 yeah, that's a good point. I grabbed this code from the runtime pkg/compare/path.go file and didn't really change it. I can change this to `[HasPrefix](https://pkg.go.dev/strings#HasPrefix)` to match the semantics of the strings package function of the same name?

PR #250, opened 2021-12-08
allow iterating over a fieldpath's ShapeRefs

jaypipes:
Adds a few utility methods to `pkg/fieldpath.Path` to make it easier to
inspect associated ShapeRefs at different parts of the path:

* `Path.At(int)` returns the path part at the supplied index.
* `Path.CopyAt(int)` returns a new Path up to the supplied index.
* `Path.ShapeRefAt(*shapeRef, int)` returns a ShapeRef corresponding to
  the path part at the supplied index.
* `Path.IterShapeRefs(*shapeRef)` returns a slice of *ShapeRef that can
  be used to iterate over the ShapeRefs associated with each of the
  Path's parts.

### DIFF
--- a/pkg/fieldpath/path.go
+++ b/pkg/fieldpath/path.go
@@ -55,6 +55,15 @@ func (p *Path) Pop() (part string) {
 	return part
 }
 
+// At returns the part of the Path at the supplied index, or empty string if
+// index exceeds boundary.
+func (p *Path) At(index int) string {
+	if index < 0 || len(p.parts) == 0 || index > len(p.parts)-1 {
+		return ""
+	}
+	return p.parts[index]
+}
+
 // Front returns the first part of the Path or empty string if the Path has no
 // parts.
 func (p *Path) Front() string {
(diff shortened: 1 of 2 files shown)


pkg/fieldpath/path.go
```diff
@@ -55,6 +55,15 @@ func (p *Path) Pop() (part string) {
 	return part
 }
 
+// At returns the part of the Path at the supplied index, or empty string if
+// index exceeds boundary.
+func (p *Path) At(index int) string {
+	if index < 0 || len(p.parts) < 0 || index > len(p.parts)-1 {
```
vijtrip2, line 61: nit:
```suggestion
	if index < 0 || len(p.parts) == 0 || index > len(p.parts)-1 {
```

A-Hilaly (approved):
Thanks @jaypipes!

pkg/fieldpath/path.go
```diff
@@ -92,6 +101,18 @@ func (p *Path) Copy() *Path {
 	return &Path{p.parts}
 }
 
+// CopyAt returns a new Path that is a copy of this Path up to the supplied
+// index.
+//
+// e.g. given Path $A containing "X.Y", $A.CopyAt(0) would return a new Path
+// containing just "X". $A.CopyAt(1) would return a new Path containing "X.Y".
+func (p *Path) CopyAt(index int) *Path {
+	if index < 0 || len(p.parts) < 0 || index > len(p.parts)-1 {
```
vijtrip2, line 110: nit: I think length function will never return less than 0. You can just check for empty i.e." == 0"
```suggestion
	if index < 0 || len(p.parts) == 0 || index > len(p.parts)-1 {
```

vijtrip2 (commented):
/approve

pkg/fieldpath/path.go
jaypipes: doh. copy/pasta mistake on my part. nice catch! fixed.

PR #252, opened 2021-12-21
Adds helm schema validation

philnichol:
Issue #, if available:
Closes https://github.com/aws-controllers-k8s/community/issues/899

Description of changes:
This adds a JSON schema to validate inputs to the helm chart, so that values of an incorrect type, regex or enumeration will fail quickly, rather than after the helm deployment runs.

Please let me know if I've missed any of the enum options.

Also worth noting that regarding the types, this will break functionality for anyone using `"true"`, `"false"`, `1`, `0` as a boolean or `"42"` as an integer, so let me know if you want those added as additional options to those fields.

### DIFF
--- a/templates/helm/values.schema.json
+++ b/templates/helm/values.schema.json
@@ -1,0 +1,213 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema#",
+  "properties": {
+    "image": {
+      "description": "Container Image",
+      "properties": {
+        "repository": {
+          "type": "string",
+          "minLength": 1
+        },
+        "tag": {
+          "type": "string",
+          "minLength": 1
+        },
+        "pullPolicy": {
+          "type": "string",
+          "enum": ["IfNotPresent", "Always", "Never"]
+        },
+        "pullSecrets": {
+          "type": "array"
+        }
+      },
+      "required": [
+          "repository",
+          "tag",
+          "pullPolicy"
+      ],
+      "type": "object"
+    },
+    "nameOverride": {
+      "type": "string"
+    },
+    "fullNameOverride": {
+      "type": "string"
+    },
+    "deployment": {
+      "description": "Deployment settings",
+      "properties": {
+        "annotations": {
+          "type": "object"
+        },
+        "labels": {
+          "type": "object"
+        },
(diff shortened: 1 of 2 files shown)


templates/helm/values.schema.json
```diff
@@ -0,0 +1,213 @@
...
+        "affinity": {
+          "type": "object"
+        },
+        "priorityClassName": {
+          "type": "string"
+        }
+      },
+      "required": [
+          "containerPort",
+          "nodeSelector"
```
RedbackThomson, line 65: I don't believe node selector is required

templates/helm/values.schema.json
```diff
@@ -0,0 +1,213 @@
...
+    },
+    "log": {
+      "description": "Logging settings",
+      "properties": {
+        "enable_development_logging": {
+          "type": "boolean"
+        },
+        "level": {
+          "type": "string",
+          "enum": ["debug", "info"]
```
RedbackThomson, line 169: There may be more levels. Perhaps `error` and `fatal`? For now I might just recommend leaving it as an open-ended string.

templates/helm/values.schema.json
```diff
@@ -0,0 +1,213 @@
...
+    },
+    "installScope": {
+      "type": "string",
+      "enum": ["cluster", "namespace"]
+    },
+    "resourceTags": {
+      "type": "array",
+      "items": {
+        "type": "string",
+        "pattern": "^.*=.*$"
```
RedbackThomson, line 182: ❤️

RedbackThomson (changes requested):
This work is amazing! Thank you for this great contribution. 

Something else to add will be an entry in the code to tell the generator to copy this file alongside the `values.yaml` in the output - see https://github.com/aws-controllers-k8s/code-generator/blob/d1d34fc91d8b715ae08a5cef6defe04c623d9918/pkg/generate/ack/release.go#L29

philnichol:
@RedbackThomson thanks for the speedy review 😃 , have pushed a commit that hopefully addresses your comments

RedbackThomson (approved):
Sweet. Looks like we are ready to go!

/lgtm

PR #254, opened 2021-12-30
Hardening container

philnichol:
Issue #, if available:

Relates https://github.com/aws-controllers-k8s/community/issues/1112

Description of changes:
- ~~No longer runs as root, runs as nobody instead, since runtime is from scratch I've added a "dummy" /etc/shadow file~~
- ~~Runtime image is now "from scratch" since we don't need much other than ca-certs and the binary itself (eg. curl, vim, etc)~~
- Standard principle of least privilege security caps in deployment manifest (drop all plus explicit least privilege deployment/pod settings and capabilities)

This is a draft since there's still stuff missing, and not sure if you would want to go in a different direction

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,5 @@
 # Base image to use for the final stage
 ARG base_image=public.ecr.aws/amazonlinux/amazonlinux:2
-FROM public.ecr.aws/docker/library/alpine:3.15 as alpine
-
-# these are required for the runtime image to have standard CAs and run as non-root
-RUN apk add -U --no-cache ca-certificates
-RUN echo "nobody:x:65534:65534:Nobody:/:" > /etc_passwd
- 
 # Build the manager binary
 FROM public.ecr.aws/bitnami/golang:1.15.15 as builder
 
@@ -47,15 +41,9 @@ RUN GIT_VERSION=$service_controller_git_version && \
     -X ${VERSION_PKG}.BuildDate=${BUILD_DATE}" \
     -a -o $work_dir/bin/controller $work_dir/cmd/controller/main.go
 
-FROM scratch
-
+FROM $base_image
 ARG service_alias
 ARG work_dir=/github.com/aws-controllers-k8s/$service_alias-controller
 WORKDIR /
-
-COPY --from=alpine /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
-COPY --from=alpine /etc_passwd /etc/passwd
-
-COPY --chown=nobody --from=builder $work_dir/bin/controller $work_dir/LICENSE $work_dir/ATTRIBUTION.md /bin/
-USER nobody
+COPY --from=builder $work_dir/bin/controller $work_dir/LICENSE $work_dir/ATTRIBUTION.md /bin/
 ENTRYPOINT ["/bin/controller"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # Base image to use for the final stage
 ARG base_image=public.ecr.aws/amazonlinux/amazonlinux:2
-FROM alpine:3.6 as alpine
+FROM alpine:3.15 as alpine
 
 # these are required for the runtime image to have standard CAs and run as non-root
 RUN apk add -U --no-cache ca-certificates
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,11 @@
 # Base image to use for the final stage
 ARG base_image=public.ecr.aws/amazonlinux/amazonlinux:2
+FROM alpine:3.6 as alpine
+
+# these are required for the runtime image to have standard CAs and run as non-root
+RUN apk add -U --no-cache ca-certificates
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,5 +1,11 @@
 # Base image to use for the final stage
 ARG base_image=public.ecr.aws/amazonlinux/amazonlinux:2
+FROM alpine:3.15 as alpine
```
vijtrip2, line 3: In ACK, we use images from `public ecr` because of docker pull throttling

vijtrip2 (commented):
one single comment, otherwise looks good (I will test it locally once just for proof and then approve :) )

Dockerfile
philnichol: Thanks, really good point. And nice to know public ECR mirrors docker hub, great job! docker-rate-limit is a pain.
Should base_image ARG also be removed from Dockerfile, Makefile, CI, etc since it's no longer used? Or leave that for a separate PR?
And yeah, definitely worth testing locally with a few controllers 😄

vijtrip2: 1. I was unable to find legitimate/popular `scratch` image on `public ecr` :/ , so we'll probably continue to use `amazonlinux` version until a `scratch` image is published on `public ecr` . I did find that new release of `amazonlinux(public.ecr.aws/amazonlinux/amazonlinux:2018.03.0.20211222.0)` does not have security vulnerabilities so we will probably use that before migrating to `scratch`
2. Given the fact that ACK team will enable regular scanning on controller images, if this `amazonlinux` image needs further update, ACK team will be made aware of it.
3. I will try out your non-root solution with `amazonlinux` locally and comment on PR if it worked.

Thank you for your amazing contribution.
(Re: Golang vulnerabilities, ACK needs upstream controller runtime updates but there have been changes in upstream API(`Object` parameter type), which will need some refactoring in ACK runtime. I am looking into it.)

philnichol: Thanks :) Obviously your call, but scratch is a special container startpoint which contains nothing, don't think it actually counts as a pull unless running a really old version of Docker (it's literally an empty container, no FS, no shell, nothing, all that's in there is your go binary). bit more info here: https://hub.docker.com/_/scratch. It's quite popular among containers running go binaries since go produces a single self-contained executable including the runtime

vijtrip2: I will try to publish `scratch` image into ACK's `public-ecr-repo` and hopefully we will be able to use it instead of docker hub `scratch`. I will update here if I will successfully achieve it. :)

vijtrip2: excellent point! testing locally now with your changes as is.

Dockerfile
```diff
@@ -1,5 +1,11 @@
 # Base image to use for the final stage
 ARG base_image=public.ecr.aws/amazonlinux/amazonlinux:2
```
vijtrip2, line 2: nit: Please remove this `base_image` arg as it is no longer used.

vijtrip2 (approved):
This is excellent stuff! Thank you!

I tested it locally by building `ecr-controller` image, installing using new helm chart template, creating an `ecr repository` and deleting it. Everything works fine! :) 

I will let another ACK team member give this a look as well but LGTM.

----------

P.S. two more files need update, `Dockerfile.local` and `templates/config/controller/deployment.yaml.tpl` but I can handle that in separate PR.

philnichol:
thanks @vijtrip2 ! Have addressed your feedback and included the files you mentioned. Also just realised since this is security-related it probably should have been raised via a more private medium, apologies for that, won't happen again

jaypipes (changes requested):
Firstly, @philnichol thank you so much for your PR and for raising awareness of this important topic. I really appreciate it!

That said, I'd like to request that for this particular PR, we remove any changes to the Dockerfiles and *only* update the deployment.yaml files/templates.

I've just had a conversation with @micahhausler (who is my go-to person to talk about container build/toolchain security) and I'd like ACK to move away from the fat AL2 container image base and into the distroless-like EKS-D minimal AL2-based base non-root container image available here:

https://gallery.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-nonroot
(source code for the above image is here: https://github.com/aws/eks-distro-build-tooling/tree/main/eks-distro-base)

The reason I'd like to do this is twofold:

1) The EKS-D/EKS-Anywhere team has expertise in build toolchain provenance and security and all aspects of the EKS-D build toolchain are audited and owned by AMZN

2) We (the ACK team) should take advantage of the work that the EKS-D team has done over the last 2 years and reduce our overall surface area (for both supply-chain attacks as well as simply the amount of work it takes to keep a build chain up to date and secure).

Therefore, @philnichol if we could slim this particular PR down to just the changes to the deployment manifests (which are awesome, BTW!) and tackle the refactoring of the base image in a separate PR, that would be great. Thank you!

philnichol:
@jaypipes thanks for the feedback! Great point, makes sense to use the work that you guys already have on minimal container images 😄 , I've removed the Dockerfiles from this PR

vijtrip2 (approved):

jaypipes (approved):

jaypipes:
/lgtm

thanks @philnichol

PR #255, opened 2022-01-05
allow SetFieldConfig.From to refer to nested fields

jaypipes:
Often, the Go type of a Create Input shape field (which goes in the
resource's Spec struct) is different from the Go type of the same-named
field in an Output shape. This causes problems for the code generator,
which doesn't know how to set the value of the Spec field due to these
different source and target Go types.

The `pkg/generate/config.SetConfig` struct was designed to solve this
particular issue, however the original solution only worked properly
when the field in the Output shape that we wanted to set the Spec field
*from* was at the same "level" as the field we wanted to set the value
*to*.

Take for example the case of the IAM Role resource, which has a Create
Input shape with a string `PermissionsBoundary` field. When creating the
Role, the user specifies the PermissionsBoundary as an ARN field. In the
Create and Get Output shapes, however, the `PermissionsBoundary` field
is a struct containing a `PermissionsBoundaryArn` string field. We want
to set the value of the Role resource's `Spec.PermissionsBoundary` field
to the string value of this nested
`PermissionsBoundary.PermissionsBoundaryArn` field. This patch
facilitates that by modifying the `pkg/generate/code.SetResource`
function to look up the field in the Output shape via a field path
instead of a single field name map lookup.

Issue aws-controllers-k8s/community#1065


pkg/testdata/models/apis/iam/0000-00-00/generator.yaml
```diff
@@ -0,0 +1,37 @@
...
+            RoleName: Name
+        UpdateRole:
+          input_fields:
+            RoleName: Name
+        DeleteRole:
+          input_fields:
+            RoleName: Name
+    fields:
+      PermissionsBoundary:
+        set:
```
brycahta, line 35: Does the interface require an operation to be set? It looks to be the case [here](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/ack/controller.go#L112)

jaypipes: If `method` is missing, it means all operations :)

https://github.com/aws-controllers-k8s/code-generator/blob/d56efd462bcb087428d527e6a6a220caa6298eba/pkg/generate/config/field.go#L152-L157

brycahta (approved):

vijtrip2 (approved):

PR #256, opened 2022-01-06
Export ToNames in template func map

RedbackThomson:
Description of changes:
Export `names.New` as `ToNames` in the template function map. This will allow for mapping from AWS SDK types to their respective CRD Spec and Status field names. For example, in the EKS controller, we want to create a `awssdk.ResourceVpcConfig` constructor that uses `svcapitypes.ResourceVPCConfig` as input.

### DIFF
--- a/pkg/generate/ack/controller.go
+++ b/pkg/generate/ack/controller.go
@@ -23,6 +23,7 @@ import (
 	"github.com/aws-controllers-k8s/code-generator/pkg/generate/templateset"
 	"github.com/aws-controllers-k8s/code-generator/pkg/model"
 	ackmodel "github.com/aws-controllers-k8s/code-generator/pkg/model"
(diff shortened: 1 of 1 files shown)


pkg/generate/ack/controller.go
```diff
@@ -57,6 +58,9 @@ var (
 		"Dereference": func(s *string) string {
 			return *s
 		},
+		"ToNames": func(s string) names.Names {
+			return names.New(s)
+		},
```
jaypipes, line 63: Would it be possible to just use [`pkg/model.Field.Names`](https://github.com/aws-controllers-k8s/code-generator/blob/82c294c2e8fc6ba23baa0034520e84351bb7a32f/pkg/model/field.go#L35) instead?

RedbackThomson:
No longer needed, as `CRD.SpecFields` and `CRD.StatusFields` already contain the names and should be used instead.

PR #258, opened 2022-01-07
collect Field structs for members of structures

jaypipes:
When inferring the structure of a CRD and its fields, we create a
`pkg/model.Field` struct for each *top-level* field in the CRD's Spec or
Status. The `pkg/model.GetCRDs()` function ends up calling
`pkg/model.CRD:AddSpecField()` and `pkg/model.CRD:AddStatusField()` for
those field definitions that the API inference stage of the generation
pipeline discovers in the aws-sdk-go API model.

Because we were creating `pkg/model.Field` objects for only the
top-level fields in Spec and Status and the `pkg/model.Field` object is
what was tied to a `pkg/generate/config.FieldConfig` object, this meant
that only the top-level field definitions were able to be *configured*
via the generator.yaml file. This is a problem for more complex APIs
like EC2 which have resources with top-level fields that are structures
themselves. We were not able to configure nested fields properly because
we were not actually constructing a `pkg/model.Field` object for these
nested fields.

This PR enhances the API inference process inside the code generator to
detect when a new `pkg/model.Field` refers to an
`aws-sdk-go/private/model.Shape` that is of type "structure" and recurse
through that Shape's MemberRefs collection, building a map of
`pkg/model.Field` objects for each member field along with a fieldpath
that allows these nested struct member fields to later be "found" by the
code generator.

Issue: aws-controllers-k8s/community#1100

### DIFF
--- a/pkg/model/field.go
+++ b/pkg/model/field.go
@@ -47,6 +47,10 @@ type Field struct {
 	GoTypeWithPkgName string
 	ShapeRef          *awssdkmodel.ShapeRef
 	FieldConfig       *ackgenconfig.FieldConfig
+
+	// MemberFields is a map, keyed by the *renamed, cleaned, camel-cased name* of
+	// member fields when this Field is a struct type.
+	MemberFields map[string]*Field
 }
 
 // IsRequired checks the FieldConfig for Field and returns if the field is
@@ -162,6 +166,7 @@ func NewField(
 	shapeRef *awssdkmodel.ShapeRef,
 	cfg *ackgenconfig.FieldConfig,
 ) *Field {
+	memberFields := map[string]*Field{}
 	var gte, gt, gtwp string
 	var shape *awssdkmodel.Shape
 	if shapeRef != nil {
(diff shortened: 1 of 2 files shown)


RedbackThomson:
Will it be necessary in the future to have a similar process for slice element structs? 

/approve

jaypipes:
> Will it be necessary in the future to have a similar process for slice element structs?

Yes. And map types. Those PRs will follow this one.

pkg/model/field.go
```diff
@@ -170,6 +175,15 @@ func NewField(
 
 	if shape != nil {
 		gte, gt, gtwp = CleanGoType(crd.sdkAPI, crd.cfg, shape, cfg)
+		if shape.Type == "structure" {
+			// "unpack" the member fields composing this struct field...
+			for _, memberName := range shape.MemberNames() {
+				cleanMemberNames := names.New(memberName)
```
brycahta, line 181: Do we also need to check for renames within the struct?

jaypipes: @brycahta we *will*, yes, but since we don't *yet* support renames for nested struct fields, I will add that at a later time.

PR #260, opened 2022-01-11
allow ignore nested fieldpaths

jaypipes:
When implementing the update code paths for OpenSearchService, I found
out I needed to ignore the AutoTuneOptions.RollbackOnDisable nested
field within the UpdateDomainConfigInput shape. When I went to put the
following in the `generator.yaml` file:

```yaml
ignore:
  field_paths:
    - UpdateDomainConfigInput.AutoTuneOptions.RollbackOnDisable
```

What ended up happening is that the *entire* AutoTuneOptions field was
ignored instead of just the RollbackOnDisable field.

This patch fixes this situation by modifying the
`pkg/model.Model:ApplyIgnoreShapeRules()` method to account for nested
field paths and not just a two-level field path that has the Input or
Output shape name as the first part of the field path and the top-level
member field name as the second part of the field path.

### DIFF
--- a/pkg/model/model.go
+++ b/pkg/model/model.go
@@ -19,6 +19,7 @@ import (
 	"sort"
 	"strings"
 
+	ackfp "github.com/aws-controllers-k8s/code-generator/pkg/fieldpath"
 	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/generate/config"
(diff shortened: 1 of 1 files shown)


RedbackThomson:
Funky. I like it. Shape refs are a PITA but I guess we have to live with them.

/lgtm

PR #261, opened 2022-01-11
always use single dot fieldpath notation

jaypipes:
This patch completes earlier work that was done in the `pkg/fieldpath`
package that only uses a single dot notation to refer to a *field* with
type *list* or *map* within a field path.

We previously were using a double dot in field paths in `generator.yaml`
config files to refer to fields of type list or map. This was erroneous,
since referring to the *field* (not the *value* of the field) only
requires a single dot in order for the fieldpath to "locate" a field
within a path.

In this PR, we simplify and DRY up the inference code in
`pkg/model/model.go`'s processing of "nested fields" so that the field
path added for all Field objects uses this single-dot notation only.

### DIFF
--- a/pkg/generate/code/set_sdk.go
+++ b/pkg/generate/code/set_sdk.go
@@ -1145,16 +1145,14 @@ func setSDKForSlice(
 	//
 	//  f0elem.SetMyField(*f0iter)
 	containerFieldName := ""
-	sourceAttributePath := sourceFieldPath
 	if targetShape.MemberRef.Shape.Type == "structure" {
 		containerFieldName = targetFieldName
-		sourceAttributePath = sourceFieldPath + "."
 	}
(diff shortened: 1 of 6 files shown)


vijtrip2 (approved):
Much cleaner now. Like it! LGTM.

vijtrip2:
Just curious that if we need to refer to `value` of a nested field in list/map, we will still use double-dot notation, correct? 
(Late-Initialization code uses double-dot notation to copy field values for map types)

jaypipes:
> Just curious that if we need to refer to `value` of a nested field in list/map, we will still use double-dot notation, correct? (Late-Initialization code uses double-dot notation to copy field values for map types)

Maybe? I'd have to take a look at the code :)

RedbackThomson (approved):
/lgtm

PR #262, opened 2022-01-13
Non root controller image

vijtrip2:
Description of changes:

#### Before
* OpenShift enforces `nonRoot` security posture by running container images using random UID and does not allow specifying `runAsUser` field in deployment spec.
* Before this change, ACK generated artifacts enforced `nonRoot` security posture using `SecurityContext`'s `runAsUser` and `runAsNonRoot` field, while the container image ran as `root` by default

-----

#### Now

* With this change, ACK controller images now run by default with `non-root` User(1000)
* Removed `runAsUser` field from deployment templates. This change keeps the existing functionality because K8s pod inherit UserId from Image metadata and ACK controller still runs with non-root user(1000) [Tested and Validated]
* Removing `runAsUser` unblock OpenShift installation and allows OpenShift to run container using random UserId
* Keeping `runAsNonRoot` field, keeps the validation in place that ACK controller does not run as `root`. This validation helps security posture for both OpenShift and helm/kustomize installation. 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
 # Base image to use for the final stage
 ARG base_image=public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-nonroot:2021-12-01-1638322424
 # Build the manager binary
+# TODO(vijtrip2) move this builder image to public.ecr.aws/eks-distro-build-tooling/builder-base, when builder-base
+# supports golang 1.17
 FROM public.ecr.aws/bitnami/golang:1.17.5 as builder
 
 ARG service_alias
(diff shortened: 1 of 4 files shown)


acornett21:
@vijtrip2 I'm not sure OpenShift allows this, I think the deployment/container are a bonded pair and a change in either would require an `SecurityContextConstraints` to exist to allow this.

vijtrip2:
> @vijtrip2 I'm not sure OpenShift allows this, I think the deployment/container are a bonded pair and a change in either would require an `SecurityContextConstraints` to exist to allow this.

I saw post about bitnami non-root container that runs on open-shift and they are built as default non-root images. I have image for you that you can use to test this locally. Following up offline on slack.

acornett21:
@vijtrip2 It seems in this case the UID that OpenShift generates for the namespace does get propigated all the way to the node -> pod -> container and having `User 1000` in the container does not override that anywhere, which is what I was concerned of. I think unblocks us for now and we can iterate on this later if need be.

acornett21 (approved):

jaypipes (approved):
Nice, thanks for the deep dive on this @vijtrip2!

PR #265, opened 2022-01-14
unpack struct element/value fields of list/map

jaypipes:
A previous commit added functionality to the API/resource inference
process to "unpack" a struct field's member fields into the
`pkg/model.Field.MemberFields` map. However, I had failed to account for
lists or maps of structs. This patch accounts for the list or map of
structs fields, stepping into the list field's element struct type and
adding a Field definition to the containing Field's `MemberFields` map
for each member field in the element struct type. Same for containing
shapes of type "map".

With this patch, we're now able to "access" via field path any nested
field (including its corresponding `FieldConfig` object) which means
that in following patches we can finally resolve the "different Go types
for nested fields" problem we've had in the code generator.

Related issue: aws-controllers-k8s/community#1065

### DIFF
--- a/pkg/model/field.go
+++ b/pkg/model/field.go
@@ -154,15 +154,48 @@ func NewField(
 	if shapeRef != nil {
 		shape = shapeRef.Shape
 	}
+	// this is a pointer to the "parent" containing Shape when the field being
+	// processed here is a structure or a list/map of structures.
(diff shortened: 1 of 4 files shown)


brycahta (approved):
Looks great!! 🎉 thanks @jaypipes !

A-Hilaly (approved):

PR #268, opened 2022-01-18
support custom fields of list/scalar/map types

jaypipes:
Adds functionality for the generator.yaml file to contain a
`FieldConfig.Type` attribute that is the Go type override string for a
field. If the Go type of the field cannot be inferred via the Create
Input/Output shape *or* via the `FieldConfig.From` (SourceFieldConfig)
value, we can look at this new `FieldConfig.Type` value and construct a
`aws-sdk-go/private/model/api:ShapeRef` object manually.

Thus, with this patch, we can do something like this (example from the
iam-controller's generator.yaml file, which is what I tested this patch
on):

```yaml
resources:
  Role:
    fields:
      Policies:
        type: "[]*string"
```


vijtrip2 (approved):
Looks Good!

PR #269, opened 2022-01-18
Return `OpTypeList` for singular plurals

brycahta:
Issue #, if available: N/A, [heuristic](https://github.com/aws-controllers-k8s/code-generator/blob/89b0a106fea6268609fd666270518145099646f7/pkg/model/op.go#L110-L118) is returning `ReadOne` operation for `DescribeDhcpOptions` when it is a `ReadMany` call

Description of changes:
* returns `OpTypeList` instead of `OpTypeGet` for "singular plural" in `"Describe"` processing
* removes duplicate code
* updates comments on singular-plural description to be clearer and present in all affected areas

### DIFF
--- a/pkg/model/op.go
+++ b/pkg/model/op.go
@@ -111,7 +111,7 @@ func GetOpTypeAndResourceNameFromOpID(
 		resName := strings.TrimPrefix(opID, "Describe")
 		if pluralize.IsPlural(resName) {
 			if resourceExistsInConfig(resName, cfg) {
-				return OpTypeGet, resName
+				return OpTypeList, resName
 			}
 			return OpTypeList, pluralize.Singular(resName)
 		}
--- a/pkg/model/op_test.go
+++ b/pkg/model/op_test.go
@@ -144,7 +144,7 @@ func TestGetOpTypeAndResourceNameFromOpID_PluralSingular(t *testing.T) {
 		},
 		{
 			"DescribeDhcpOptions",
-			model.OpTypeGet,
+			model.OpTypeList,
 			"DhcpOptions",
 		},
 		{


pkg/model/op.go
```diff
@@ -124,19 +126,23 @@ func GetOpTypeAndResourceNameFromOpID(
 		}
 		resName := strings.TrimPrefix(opID, "Get")
 		if pluralize.IsPlural(resName) {
-			if resourceExistsInConfig(resName, cfg) {
-				return OpTypeGet, resName
+			// "pluralized singular" resources like EC2's DhcpOptions
+			// are explicitly defined in generator config to NOT singularize
+			// because there is no such thing as DhcpOption
+			if !resourceExistsInConfig(resName, cfg) {
+				return OpTypeList, pluralize.Singular(resName)
```
jaypipes, line 133: why are we returning DhcpOption here, then?

jaypipes (changes requested):
I was thinking this would be a single line changed (line 128, changed to OpTypeList instead of OpTypeGet to deal with the DhcpOption misnaming...) but I see a bunch of changes here.

pkg/model/op.go
brycahta: That's only the case if `DHCPOptions:` is not defined in the config. If it is, then `return OpTypeGet, resName` is hit on line 136 without singularizing the `resName`

brycahta:
> I was thinking this would be a single line changed (line 128, changed to OpTypeList instead of OpTypeGet to deal with the DhcpOption misnaming...) but I see a bunch of changes here.

Yeah, did some cleanup while here.. tried to break out by commits; sorry if it's still cluttered. No other functionality was added/updated.

brycahta:
> I was thinking this would be a single line changed (line 128, changed to OpTypeList instead of OpTypeGet to deal with the DhcpOption misnaming...) but I see a bunch of changes here.

Got it, backed out clarifications

PR #273, opened 2022-01-22
added buildmode=pie flag

philnichol:
Issue #, if available:
Closes https://github.com/aws-controllers-k8s/community/issues/1134

Description of changes:
Thanks in advance for your review. This adds the `-buildmode=pie` flag to get ASLR, improving the security posture of the binary.

Since it may increase memory usage, I wasn't sure if we should also increase the default memory requests/limits in the helm chart, so let me know if you think this should be included?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,6 +41,7 @@ RUN GIT_VERSION=$service_controller_git_version && \
     go build -ldflags="-X ${VERSION_PKG}.GitVersion=${GIT_VERSION} \
     -X ${VERSION_PKG}.GitCommit=${GIT_COMMIT} \
     -X ${VERSION_PKG}.BuildDate=${BUILD_DATE}" \
+    -buildmode=pie \
     -a -o $work_dir/bin/controller $work_dir/cmd/controller/main.go
 
 FROM $base_image
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -52,6 +52,7 @@ RUN GIT_VERSION=$service_controller_git_version && \
     go build -ldflags="-X ${VERSION_PKG}.GitVersion=${GIT_VERSION} \
     -X ${VERSION_PKG}.GitCommit=${GIT_COMMIT} \
     -X ${VERSION_PKG}.BuildDate=${BUILD_DATE}" \
+    -buildmode=pie \
     -modfile="${work_dir}/go.local.mod" \
     -a -o $work_dir/bin/controller $work_dir/cmd/controller/main.go
 


philnichol:
I'll see if I can replicate this locally to see the issue

jaypipes (approved):
:+1:  article lays out a good reasoning for enabling this.

vijtrip2:
I tested `s3 controller` locally using `kind-test` (which uses `docker` to build image instead of `buildah`) and tests are failing with same reason. The error in controller logs is 
```
standard_init_linux.go:228: exec user process caused: no such file or directory
```

Could it be because we are building final container image from a minimal base image now? I am not sure. needs more deep dive.

-----------

However if I execute controller binary (build with -buildmode=pie) it works fine on my Mac machine.

RedbackThomson:
I have just spent some time playing around with it to figure out what the problem is. We can recreate the controller logs by simply attempting to run the build container locally (`docker run aws-controllers-k8s:s3-v0.0.10-...`). This fails with the same error as @vijtrip2 saw above.

If you intercept the build before it packages it into the minimal build base, running it inside the golang build image, the binary works as expected. This means the pie mode build is working successfully in the builder, which leads me to believe the problem is with the packaging image (the rootless EKS-D base).

I switched the EKS-D base from the `-minimal-base-nonroot` variant to the `-minimal-base-glibc` and it works again. I'm not familiar enough with the Go glibc requirements, or with how these base images are constructed, to point to any specific dependency that is contained in one that isn't in the other. If anyone can take it from there, that would be awesome.

philnichol:
The more I read about this to try and find what might cause this, the less I'm convinced of the benefits it brings. Also hasn't quite been minimal effort as I incorrectly advertised 😉 

_(Russ Cox) Address space randomization is an OS-level workaround for a language-level problem, namely that simple C programs tend to be full of exploitable buffer overflows. Go fixes this at the language level, with bounds-checked arrays and slices and no dangling pointers, which makes the OS-level workaround much less important. In return, we receive the incredible debuggability of deterministic address space layout. I would not give that up lightly.

(Thomas Bushnell, BSG) You’re engaging in a common mistake: knowing that it is important to have seat belts on cars, because of the risks that cars are prone to, you’re now arguing for seat belts in restaurants (in case a car should crash into the restaurant), seat belts in trains (even though the risk/reward ratio is entirely different), seat belts on bicycles, and so forth._

https://groups.google.com/g/golang-nuts/c/Jd9tlNc6jUE

jaypipes:
> The more I read about this to try and find what might cause this, the less I'm convinced of the benefits it brings. Also hasn't quite been minimal effort as I incorrectly advertised wink
> 
> _(Russ Cox) Address space randomization is an OS-level workaround for a language-level problem, namely that simple C programs tend to be full of exploitable buffer overflows. Go fixes this at the language level, with bounds-checked arrays and slices and no dangling pointers, which makes the OS-level workaround much less important. In return, we receive the incredible debuggability of deterministic address space layout. I would not give that up lightly.
> 
> (Thomas Bushnell, BSG) You’re engaging in a common mistake: knowing that it is important to have seat belts on cars, because of the risks that cars are prone to, you’re now arguing for seat belts in restaurants (in case a car should crash into the restaurant), seat belts in trains (even though the risk/reward ratio is entirely different), seat belts on bicycles, and so forth._
> 
> https://groups.google.com/g/golang-nuts/c/Jd9tlNc6jUE

Interesting, thanks for diving deep on this @philnichol. So, I gather you are recommending that we close this PR out?

philnichol:
Yeah, from what I read unless we areusing unsafe, linking to C libraries or using CGO the benefit isn't really there. Go doesn't suffer from the risks of buffer overflows like some other languages. Since the people commenting in that thread literally wrote the go runtime I'd definitely trust their expertise on the matter over mine + a few bugs.

jaypipes:
No worries @philnichol, we appreciate all your effort regardless! :) Closing this one out then.

philnichol:
Thanks and apologies fo the noise!

PR #274, opened 2022-01-22
Support other auth mechanisms

philnichol:
Issue #, if available:
Relates https://github.com/aws-controllers-k8s/community/issues/1041

Description of changes:
This allows users of the helm chart to pass in a list of ENVVARs to the controller. These can include auth via AWS envvars.
Since the nature of these may be sensitive they are stored in a Secret rather than directly into the container spec.
Envar precedence will still be the ones set directly in the container spec, so there is no chance of these overwriting a setting elsewhere.
Assuming this is an acceptable way of doing this, I can raise a follow-up updating the docs with some other auth methods other than IRSA (inc AWS envvars and Kiam)

### DIFF
--- a/templates/helm/templates/deployment.yaml
+++ b/templates/helm/templates/deployment.yaml
@@ -73,6 +73,11 @@ spec:
           value: {{ .Values.log.level | quote }}
         - name: ACK_RESOURCE_TAGS
           value: {{ join "," .Values.resourceTags | quote }}
+        {{- if .Values.deployment.secretEnvVars }}
+        envFrom:
+        - secretRef:
+            name: {{ include "app.fullname" . }}
+        {{- end }}
         securityContext:
           allowPrivilegeEscalation: false
           privileged: false
--- a/templates/helm/templates/deployment.yaml
+++ b/templates/helm/templates/deployment.yaml
@@ -73,10 +73,10 @@ spec:
           value: {{ .Values.log.level | quote }}
         - name: ACK_RESOURCE_TAGS
           value: {{ join "," .Values.resourceTags | quote }}
-        {{- if .Values.deployment.secretEnvVars }}
+        {{- if .Values.deployment.env.fromSecret.vars }}
         envFrom:
         - secretRef:
-            name: {{ default (include "app.fullname" .) .Values.deployment.secretName }}
+            name: {{ default (include "app.fullname" .) .Values.deployment.env.fromSecret.name }}
         {{- end }}
         securityContext:
           allowPrivilegeEscalation: false
--- a/templates/helm/templates/deployment.yaml
+++ b/templates/helm/templates/deployment.yaml
@@ -76,7 +76,7 @@ spec:
         {{- if .Values.deployment.secretEnvVars }}
         envFrom:
         - secretRef:
-            name: {{ include "app.fullname" . }}
+            name: {{ default (include "app.fullname" .) .Values.deployment.secretName }}
         {{- end }}
         securityContext:
           allowPrivilegeEscalation: false
--- a/templates/helm/templates/secret.yaml
+++ b/templates/helm/templates/secret.yaml
@@ -1,8 +1,8 @@
-{{- if .Values.deployment.secretEnvVars -}}
+{{- if .Values.deployment.env.fromSecret.vars -}}
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ default (include "app.fullname" .) .Values.deployment.secretName }}
+  name: {{ default (include "app.fullname" .) .Values.deployment.env.fromSecret.name }}
   namespace: {{ .Release.Namespace }}
   labels:
     app.kubernetes.io/name: {{ include "app.name" . }}
@@ -13,7 +13,7 @@ metadata:
     helm.sh/chart: {{ include "chart.name-version" . }}
     control-plane: controller
 data:
-{{- range $key, $value := .Values.deployment.secretEnvVars }}
+{{- range $key, $value :=.Values.deployment.env.fromSecret.vars }}
   {{ $key }}: {{ $value | b64enc | quote }}
 {{- end }}
 {{- end -}}
--- a/templates/helm/templates/secret.yaml
+++ b/templates/helm/templates/secret.yaml
@@ -14,6 +14,6 @@ metadata:
     control-plane: controller
 data:
 {{- range $key, $value := .Values.deployment.secretEnvVars }}
-  {{ $key }}: {{ $value | quote }}
+  {{ $key }}: {{ $value | b64enc | quote }}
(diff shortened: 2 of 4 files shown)


vijtrip2:
Excellent idea... I will test this locally once to see if it works as expected. 🙌 

/ok-to-test

templates/helm/templates/secret.yaml
```diff
@@ -0,0 +1,19 @@
...
+  labels:
+    app.kubernetes.io/name: {{ include "app.name" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: Helm
+    app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+    k8s-app: {{ include "app.name" . }}
+    helm.sh/chart: {{ include "chart.name-version" . }}
+    control-plane: controller
+data:
+{{- range $key, $value := .Values.deployment.secretEnvVars }}
```
jaypipes, line 16: I believe you want `secretEnvVars` to be of type `object` not `array`, otherwise $key above is going to be numeric indexes...

templates/helm/values.yaml.tpl
```diff
@@ -28,6 +28,12 @@ deployment:
   # Which priorityClassName to set?
   # See: https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#pod-priority
   priorityClassName: ""
+
+  # List of key/value pairs to be put in secret to act as a source for environment variables for the controller.
+  # Can be used to authenticate without IRSA
+  # eg. - AWS_ACCESS_KEY_ID: "YOUR_AWS_ACCESS_KEY_ID"
+  # eg. - AWS_SECRET_ACCESS_KEY: "YOUR_AWS_SECRET_ACCESS_KEY"
+  secretEnvVars: []
```
jaypipes, line 36: should be:

```yaml
secretEnvVars: {}
```

templates/helm/templates/deployment.yaml
```diff
@@ -73,6 +73,11 @@ spec:
           value: {{ .Values.log.level | quote }}
         - name: ACK_RESOURCE_TAGS
           value: {{ join "," .Values.resourceTags | quote }}
+        {{- if .Values.deployment.secretEnvVars }}
+        envFrom:
+        - secretRef:
+            name: {{ include "app.fullname" . }}
```
jaypipes, line 79: Consider potentially having a separate variable storing the name of the Secret to use (instead of hard-coding to a Secret named the same as the application)

jaypipes (changes requested):
Unless I'm mistaken, you will want to use a dict/object instead of an array of object?

templates/helm/templates/secret.yaml
philnichol: 🤦 you're right, will update

philnichol:
@jaypipes thanks for the feedback, have (hopefully) addressed your comments, any issues let me know

templates/helm/values.yaml.tpl
```diff
@@ -28,6 +28,15 @@ deployment:
...
   priorityClassName: ""
+
+  # Map of key/value pairs to be put in secret to act as a source for environment variables for the controller.
+  # Can be used to authenticate without IRSA (although IRSA is recommended)
+  # eg. AWS_ACCESS_KEY_ID: "YOUR_AWS_ACCESS_KEY_ID"
+  # eg. AWS_SECRET_ACCESS_KEY: "YOUR_AWS_SECRET_ACCESS_KEY"
+  secretEnvVars: {}
+
+  # Name of the secret to store your envvars, defaults to app fullname
+  secretName: ""
```
jaypipes, line 39: OK, hear me out on this one... :)

How about we structure these configuration values like so:

```yaml
deployment:
  # env contains configuration for environment variables in the Deployment
  env:
    # fromSecret contains configuration for environment variables in the Deployment
    # read from a Secret that is automatically created by Helm
    fromSecret:
      # name is the name of the Secret that is created that contains environment
      # variables inserted into the Deployment (defaults to the application full name
      name: ""
      # vars contains a map of key/value pairs representing the environment variables
      # that will be inserted into the Secret and read into the Deployment
      vars: {}
```

with the above structure, we can neatly contain all the environment-variable setup into a sub-object. We may later want to add something like:

```yaml
deployment:
  # env contains configuration for environment variables in the Deployment
  env:
    # additionalVars is a map of key/value pairs to be added as-is to the Deployment
    additionalVars: {}
```

philnichol: that is way better, I'll update now :)

RedbackThomson: This seems to just be replicating the `env` structure inside the Helm values schema. In that case, why not just inject values in the same way we do for annotations and labels? Essentially just take whatever structure they put under `deployment.env` and indent it. See [this example](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml#L11) for how to achieve it. Then we by default support anything that the deployment env can support.

jaypipes: > This seems to just be replicating the `env` structure inside the Helm values schema. In that case, why not just inject values in the same way we do for annotations and labels? Essentially just take whatever structure they put under `deployment.env` and indent it. See [this example](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml#L11) for how to achieve it. Then we by default support anything that the deployment env can support.

Sorry, not following you @RedbackThomson. Are you suggesting just making `deployment.env` be an object with a schema matching the `Deployment.Spec.Env` struct from upstream?

philnichol: that would make it more flexible, happy with either, whichever you both prefer 😄

RedbackThomson: > Sorry, not following you @RedbackThomson. Are you suggesting just making `deployment.env` be an object with a schema matching the `Deployment.Spec.Env` struct from upstream?

Sorry yeah it's hard to explain. Essentially the same way we treat [`serviceAccount.annotations`](https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/helm/values.yaml.tpl#L76-L77) as a multi-line string, and then just [indent it](https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/helm/templates/service-account.yaml#L15-L17) to put it into the manifest. I propose we do the same with the `deployment.env` field.  

This is the [common practice](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml#L143) for Helm value files to let people have ultimate customization over fields with arbitrary values.

PR #278, opened 2022-01-28
Reference field name updates

vijtrip2:
Description of changes:
* Refactor adding ReferenceField in CRD inside AddSpecField method
* Create new "Field.GetReferenceFieldName"  method to generate ReferenceFieldName from corresponding Field
* Change the ReferenceFieldName from just adding "Ref" suffix to "Singular(FieldName - IdentifierSuffix) + Ref(s)"
* Uncomment the wait for ResourceSynced condition before referencing a resource
* Update comments and docs to use "APIRef" instead of "APIIDRef"


pkg/model/crd.go
```diff
@@ -190,6 +190,15 @@ func (r *CRD) AddSpecField(
...
 	r.Fields[fPath] = f
+
+	// If this field has a ReferencesConfig, Add the new
+	// Reference field inside Spec as well
+	if fConfig != nil && fConfig.References != nil {
+		referenceFieldNames := f.GetReferenceFieldName()
+		rf := NewReferenceField(r, referenceFieldNames, shapeRef)
+		r.SpecFields[referenceFieldNames.Original] = rf
+		r.Fields[referenceFieldNames.Camel] = rf
+	}
```
jaypipes, line 201: :+1:

pkg/model/field.go
```diff
@@ -96,7 +99,7 @@ func (f *Field) GetSetterConfig(opType OpType) *ackgenconfig.SetFieldConfig {
 //          path: Status.APIID
 // ```
 // For the above configuration, 'HasReference' for 'ApiId'(Original name) field
-// will return true because a corresponding 'ApiIdRef' field will be generated
+// will return true because a corresponding 'ApiRef' field will be generated
```
jaypipes, line 102: APIRef

pkg/model/field.go
```diff
@@ -111,6 +114,41 @@ func (f *Field) IsReference() bool {
...
+	}
+	if refNamePrefix == "" {
+		panic("The corresponding field name for a reference field cannot" +
+			" be just 'id(s)' or 'arn(s)' or 'name(s)'. Current value: " +
+			f.Names.Original)
+	}
+	refName := refNamePrefix
+	// If the shape of corresponding field is a list, singularize the refNamePrefix
+	// and add Refs at the end
+	if f.ShapeRef != nil && f.ShapeRef.Shape != nil && f.ShapeRef.Shape.Type == "list" {
```
jaypipes, line 144: yup, this should work perfectly for now. I don't know of any `map[string]*AWSResourceReferenceWrapper` fields (yet!) so I think we're good here.

pkg/model/model.go
```diff
@@ -132,16 +132,6 @@ func (m *Model) GetCRDs() ([]*CRD, error) {
...
-
-			// If this field has ReferencesConfig, add corresponding
-			// resource reference field in spec as well
-			fConfig := m.cfg.ResourceFields(crdName)[fieldName]
-			if fConfig != nil && fConfig.References != nil {
-				referenceFieldNames := names.New(fieldName + "Ref")
-				rf := NewReferenceField(crd, referenceFieldNames, memberShapeRef)
-				crd.SpecFields[referenceFieldNames.Original] = rf
-				crd.Fields[referenceFieldNames.Camel] = rf
-			}
```
jaypipes, line 144: good moving this to AddSpecField. ++

jaypipes (changes requested):
@vijtrip2 this is perfectly aligned with how I want the resource reference fields to be named. Thank you!

(I have a tiny request for one more unit test case, otherwise I'm good with this)

jaypipes (approved):
Thanks, Vijay!

jaypipes:
I will let @RedbackThomson lgtm this because I know he's a huge fan of this naming convention.

pkg/model/field.go
```diff
@@ -111,6 +114,41 @@ func (f *Field) IsReference() bool {
...
+		"Arn", "Arns", "ARN", "ARNs", "ARNS",
+	}
+	for _, suffix := range identifierSuffixes {
+		if strings.HasSuffix(f.Names.Original, suffix) {
+			refNamePrefix = strings.TrimSuffix(refNamePrefix, suffix)
+			break
+		}
+	}
+	if refNamePrefix == "" {
+		panic("The corresponding field name for a reference field cannot" +
```
RedbackThomson, line 137: LOL if we ever hit this condition.

RedbackThomson (approved):
@jaypipes might have actually won me over on this. Maybe. This implementation looks clean, though. ++

/lgtm

PR #280, opened 2022-01-28
Add storageversion kubebuilder marker in Crossplane pipeline

ulucinar:
Issue #, if available:

Description of changes:
For cases where we do not delete the old version, we need to mark the generated type with storage version.


ulucinar:
Thank you @A-Hilaly.

PR #283, opened 2022-02-03
Allow `SetFieldConfig` to refer to nested Input Shape Fields

brycahta:
Issue #, if available: [#1146](https://github.com/aws-controllers-k8s/community/issues/1146)

code-generator has issues generating Go code to set a Resource's Spec field when the Go type of a Create Input shape field is different from the Go type of the *same-named field* in the Output shape. The `pkg/generate/config.SetConfig` was introduced to solve this problem; however, the current implementation does not support `SetFieldConfig` for **nested** Create Input shape fields (Resource Spec fields). 

For example, when generating EC2 DHCPOptions Resource the `CreateDhcpOptionsInput` Shape contains an attribute, `Values []*string` but the `CreateDhcpOptionsOutput` Shape has the same-named field as a different type: `Values []*AttributeValue`. `AttributeValue` struct has a `Value *string` attribute which is what we want code-generator to use to populate `Values []*string`. In other words, set `CreateDhcpOptionsInput.DhcpConfigurations.Values []*string` using `CreateDhcpOptionsInput.DhcpOptions.DhcpConfigurations.Values.Value *string`.  The `DhcpConfigurations.Values` is the **nested** Create Input Shape field in this case.

This patch allows users to define a **nested** Input Shape Field (Resource Field in `generator.yaml`) in `SetFieldConfig` to resolve Shape mismatches occurring within top "level" fields.

### DIFF
--- a/pkg/generate/code/set_resource.go
+++ b/pkg/generate/code/set_resource.go
@@ -1238,6 +1238,29 @@ func SetResourceForStruct(
 	for memberIndex, memberName := range sourceShape.MemberNames() {
 		targetMemberShapeRef := targetShape.MemberRefs[memberName]
 		if targetMemberShapeRef == nil {
+			// This can happen when the targetShape is a primitive
+			// and the sourceShape is a struct like when EC2 resource DHCPOptions
+			// needs to set NewDhcpConfiguration.Values (targetShape = string) field from
+			// DhcpConfiguration.Values (sourceShape = AttributeValue). AttributeValue
+			// has a 'Value' member whereas the string shape does not. However,
+			// there is a possibility that the targetShape can be set from
+			// a different member within sourceShape using SetConfig.
+			if currentField, ok := r.Fields[currentPath]; ok {
+				setCfg := currentField.GetSetterConfig(model.OpTypeList)
+				if setCfg != nil && setCfg.From != nil {
+					// If the Field's SetConfig refers to a member in
+					// sourceShape, then update targetMemberShapeRef
+					// to the intended target, targetShape.
+					fp := fieldpath.FromString(*setCfg.From)
+					updatedSourceShapeRef := fp.ShapeRef(sourceShapeRef)
+					if updatedSourceShapeRef != nil && updatedSourceShapeRef.Shape != nil {
+						targetMemberShapeRef = targetShapeRef
+					}
+				}
+			}
+		}
+		if targetMemberShapeRef == nil {
+			// No member or SetConfig overrides
 			continue
 		}
 		memberVarName := fmt.Sprintf("%sf%d", targetVarName, memberIndex)
@@ -1251,6 +1274,7 @@ func SetResourceForStruct(
 		qualifiedTargetVar := fmt.Sprintf(
 			"%s.%s", targetVarName, cleanNames.Camel,
 		)
+
 		switch memberShape.Type {
 		case "list", "structure", "map":
 			{
@@ -1279,6 +1303,14 @@ func SetResourceForStruct(
 				)
 			}
 		default:
+			if _, found := targetShape.MemberRefs[memberName]; !found {
+				// This is only possible when a SetCfg is applied and
+				// targetShape is a primitive. tartgetShape has
+				// no members; therefore, remove the memberName from qualifiedTargetVar.
+				// Dereference sourceAdaptedVarName because primitives are being set.
+				qualifiedTargetVar = targetVarName
+				sourceAdaptedVarName = "*" + sourceAdaptedVarName
+			}
 			out += setResourceForScalar(
 				qualifiedTargetVar,
 				sourceAdaptedVarName,
--- a/pkg/generate/code/set_resource.go
+++ b/pkg/generate/code/set_resource.go
@@ -308,6 +308,7 @@ func SetResource(
 					setCfg,
 					sourceAdaptedVarName,
 					sourceMemberShapeRef,
+					f.Names.Camel,
 					indentLevel+1,
 				)
 				out += setResourceForScalar(
@@ -578,6 +579,7 @@ func setResourceReadMany(
 					setCfg,
 					sourceAdaptedVarName,
 					sourceMemberShapeRef,
+					f.Names.Camel,
 					indentLevel+2,
 				)
 				out += setResourceForScalar(
@@ -1156,6 +1158,8 @@ func setResourceForContainer(
 	sourceVarName string,
 	// ShapeRef of the source struct field
 	sourceShapeRef *awssdkmodel.ShapeRef,
+	// currentPath is the absolute path to targetFieldName for nested fields
+	currentPath string,
 	indentLevel int,
 ) string {
 	switch sourceShapeRef.Shape.Type {
@@ -1168,6 +1172,7 @@ func setResourceForContainer(
 			targetSetCfg,
 			sourceVarName,
 			sourceShapeRef,
+			currentPath,
 			indentLevel,
 		)
 	case "list":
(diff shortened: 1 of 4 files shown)


pkg/testdata/models/apis/ec2/0000-00-00/generator.yaml
```diff
@@ -68,8 +68,11 @@ operations:
     output_wrapper_field_path: VpcEndpoint
 
 resources:
-  DHCPOptions:
-
+  DhcpOptions:
+    fields:
+      DHCPConfigurations.Values:
+        set:
+          - from: AttributeValue.Value
```
A-Hilaly, line 75: Is it possible to add a test case for multiple nested fields? something similar to `A.B.C`

brycahta: Did you mean multiple nested fields for the *Resource* or *Shape*? Note, this is a nested example for **both**

pkg/generate/code/set_resource.go
```diff
@@ -1229,6 +1238,29 @@ func SetResourceForStruct(
 	for memberIndex, memberName := range sourceShape.MemberNames() {
 		targetMemberShapeRef := targetShape.MemberRefs[memberName]
 		if targetMemberShapeRef == nil {
```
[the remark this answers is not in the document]
brycahta: The intent is to make these checks look the same. I think it makes it easier to read the exception handling when having `targetMemberShapeRef == nil` and that there's still a chance to resolve the field (using `SetConfig`).

pkg/generate/code/set_resource.go
```diff
@@ -1229,6 +1238,29 @@ func SetResourceForStruct(
...
+			// This can happen when the targetShape is a primitive
+			// and the sourceShape is a struct like when EC2 resource DHCPOptions
+			// needs to set NewDhcpConfiguration.Values (targetShape = string) field from
+			// DhcpConfiguration.Values (sourceShape = AttributeValue). AttributeValue
+			// has a 'Value' member whereas the string shape does not. However,
+			// there is a possibility that the targetShape can be set from
+			// a different member within sourceShape using SetConfig.
+			if currentField, ok := r.Fields[currentPath]; ok {
+				setCfg := currentField.GetSetterConfig(model.OpTypeList)
```
vijtrip2, line 1249: any particular reason for hardcoding "OpTypeList" here?

brycahta: nope just limited by api here and didn't want to introduce refactoring changes. Same as [here](https://github.com/aws-controllers-k8s/code-generator/blob/2eb18c5a67e5955149a6bb1eaa92440cdfed31e3/pkg/generate/ack/controller.go#L112)

pkg/generate/code/set_resource.go
```diff
@@ -1156,6 +1158,8 @@ func setResourceForContainer(
 	sourceVarName string,
 	// ShapeRef of the source struct field
 	sourceShapeRef *awssdkmodel.ShapeRef,
+	// currentPath is the absolute path to targetFieldName for nested fields
+	currentPath string,
```
jaypipes, line 1162: Why not call this `targetFieldPath` then?

pkg/generate/code/set_resource.go
```diff
@@ -1156,6 +1158,8 @@ func setResourceForContainer(
 	sourceVarName string,
 	// ShapeRef of the source struct field
 	sourceShapeRef *awssdkmodel.ShapeRef,
+	// currentPath is the absolute path to targetFieldName for nested fields
```
jaypipes, line 1161: It is actually the field path for the target *containing* field.

pkg/generate/code/set_resource.go
```diff
@@ -1229,6 +1238,29 @@ func SetResourceForStruct(
...
+					// sourceShape, then update targetMemberShapeRef
+					// to the intended target, targetShape.
+					fp := fieldpath.FromString(*setCfg.From)
+					updatedSourceShapeRef := fp.ShapeRef(sourceShapeRef)
+					if updatedSourceShapeRef != nil && updatedSourceShapeRef.Shape != nil {
+						targetMemberShapeRef = targetShapeRef
+					}
+				}
+			}
+		}
```
jaypipes, line 1261: This block of code does not reference the member field path. `currentPath` will always refer to the containing shape's field path... I think you need to need to build up a `memberFieldPath` containing `currentPath` + '.' + `memberName`?

pkg/generate/code/set_resource.go
```diff
@@ -1229,6 +1238,29 @@ func SetResourceForStruct(
 	for memberIndex, memberName := range sourceShape.MemberNames() {
```
jaypipes, line 1238: I was expecting you to change the above to loop over the *targetShape*'s members, not the source shape...

pkg/generate/code/set_resource.go
jaypipes, line 1249: I actually don't see how passing `model.OpTypeList` hard-coded here is going to work...

pkg/testdata/models/apis/ec2/0000-00-00/generator.yaml
jaypipes, line 75: If you add:

`    method: CREATE`

to the above `generator.yaml` file, the test will fail because you are passing `OpTypeCreate` in the test but `OpTypeList` is hard-coded in the SetResourceForStruct method...

jaypipes (changes requested):
@brycahta I've got concerns about this... I don't actually see how it will work properly for non-OpTypeList methods or when there is more than a single level of nesting (since you aren't "building up" a field path for the nested fields anywhere...)

pkg/generate/code/set_resource.go
brycahta: Can you advise what should be done given `opType` is required for this api? I modeled after the similar behavior [here](https://github.com/aws-controllers-k8s/code-generator/blob/2eb18c5a67e5955149a6bb1eaa92440cdfed31e3/pkg/generate/ack/controller.go#L112).

Pass another parameter, opType, to setResource funcs?

pkg/generate/code/set_resource.go
brycahta: `currentPath` will always refer to current target shape so fetching the `SetConfig` is possible, and the path is being built on line 1295.

pkg/testdata/models/apis/ec2/0000-00-00/generator.yaml
brycahta: Good point; will update depending on convo above

pkg/generate/code/set_resource.go
brycahta: Can you elaborate on the _containing_ part? It's not clear to me

pkg/generate/code/set_resource.go
brycahta: I know we originally discussed that approach, but the implementation was more convoluted so I pivoted to what I thought would be more straightforward given the current codebase. 

I know we'll have a *code-generator* refactor task of iterating over resource fields instead of shape members in `SetResource` -- would you mind if I made this update to iterate over _targetShape_ members at that time? I think those patterns are similar and consistent and be better understood if delivered around the same time.

pkg/generate/code/set_resource.go
brycahta: > I actually don't see how passing model.OpTypeList hard-coded here is going to work...

This only works if `SetterConfig` applies to ALL methods or just `model.OpTypeList`. Either way-- the gap you pointed out is still there (what if on `Create` only).

brycahta:
closing in favor of https://github.com/aws-controllers-k8s/code-generator/pull/292

PR #287, opened 2022-02-09
Use resp variable in sdkFind

RealFatCat:
Issue #, if available:

Description of changes:
I've tried to generate sqs-controller with 0.16.5 runtime, and on running test got this error:
``` 
$ make test
go test -v ./...
?   	github.com/aws-controllers-k8s/sqs-controller/apis/v1alpha1	[no test files]
# github.com/aws-controllers-k8s/sqs-controller/pkg/resource/queue
pkg/resource/queue/sdk.go:68:6: resp declared but not used
make: *** [Makefile:19: test] Error 2
```
This patch should fix it.


jaypipes:
@RealFatCat I'm afraid that the sqs-controller is sadly not going to work properly even with this patch. This patch would allow the generation of the sqs-controller but the controller would unlikely work. Have you tried running the generated controller after this patch?

PR #288, opened 2022-02-10
allow resource references across service controllers

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1091

Description of changes:
* https://github.com/aws-controllers-k8s/community/pull/1092
* Adds a new `ServiceName` parameter in `ReferencesConfig` to indicate when the referenced resource is from another service controller
* Based on the `ServiceName` parameter imports the API-types from that service controller
* Adds the additional RBAC permissions to allow reading the referenced resource from another service controller

Tested by generating and running apigatewayv2-controller which refers the Subnet and SecurityGroup resource from ec2 controller. Please see the [proposal link](https://github.com/aws-controllers-k8s/community/pull/1092) for generated code.

### DIFF
--- a/pkg/generate/config/field.go
+++ b/pkg/generate/config/field.go
@@ -284,6 +284,12 @@ type LateInitializeConfig struct {
 // read the referred 'API' resource and copy the value from 'Status.APIID' in
 // 'Integration' resource's 'APIID' field
 type ReferencesConfig struct {
+	// ServiceName mentions the AWS service name where 'Resource' exists
+	// ServiceName is the package name for AWS service package in
+	// aws-sdk-go/models/apis/<package_name>
+	// When not specified, 'ServiceName' defaults to service name of controller
+	// which contains generator.yaml
+	ServiceName string `json:"service_name,omitempty"`
 	// Resource mentions the K8s resource which is read to resolve the
 	// reference
 	Resource string `json:"resource"`
--- a/pkg/generate/config/field.go
+++ b/pkg/generate/config/field.go
@@ -284,9 +284,15 @@ type LateInitializeConfig struct {
 // read the referred 'API' resource and copy the value from 'Status.APIID' in
 // 'Integration' resource's 'APIID' field
 type ReferencesConfig struct {
-	// ServiceName mentions the AWS service name where 'Resource' exists
-	// ServiceName is the package name for AWS service package in
-	// aws-sdk-go/models/apis/<package_name>
+	// ServiceName mentions the AWS service name where "Resource" exists.
+	// This field is used to generate the API Group for the "Resource".
+	//
+	// ServiceName is the go package name for AWS service in
+	// aws-sdk-go/service/<package_name>/api.go
+	// Ex: Use "opensearchservice" to refer "Domain" resource from
+	// opensearchservice-controller because it is the go package name for
+	// "aws-sdk-go/service/opensearchservice/api.go"
+	//
 	// When not specified, 'ServiceName' defaults to service name of controller
 	// which contains generator.yaml
 	ServiceName string `json:"service_name,omitempty"`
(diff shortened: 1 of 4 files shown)


pkg/generate/config/field.go
```diff
@@ -284,6 +284,12 @@ type LateInitializeConfig struct {
 // read the referred 'API' resource and copy the value from 'Status.APIID' in
 // 'Integration' resource's 'APIID' field
 type ReferencesConfig struct {
+	// ServiceName mentions the AWS service name where 'Resource' exists
+	// ServiceName is the package name for AWS service package in
+	// aws-sdk-go/models/apis/<package_name>
+	// When not specified, 'ServiceName' defaults to service name of controller
+	// which contains generator.yaml
+	ServiceName string `json:"service_name,omitempty"`
```
RedbackThomson, line 292: As per my [previous PR clarifying names](https://github.com/aws-controllers-k8s/code-generator/pull/213), this should be referred to as the `ServiceModelName`.

templates/pkg/resource/references.go.tpl
```diff
@@ -18,10 +18,28 @@ import (
...
 {{ end -}}
 	acktypes "github.com/aws-controllers-k8s/runtime/pkg/types"
-
 	svcapitypes "github.com/aws-controllers-k8s/{{ .ServicePackageName }}-controller/apis/{{ .APIVersion }}"
+{{ $servicePackageName := .ServicePackageName -}}
+{{ $apiVersion := .APIVersion -}}
+{{ if .CRD.HasReferenceFields -}}
+{{ range $fieldName, $field := .CRD.Fields -}}
+{{ if and $field.HasReference (not (eq $field.ReferencedServiceName $servicePackageName)) -}}
+    {{ $field.ReferencedServiceName }}apitypes "github.com/aws-controllers-k8s/{{ $field.ReferencedServiceName }}-controller/apis/{{ $apiVersion }}"
```
RedbackThomson, line 27: Here you are using the "Service package name", no the same as the "Service model name" that you ask for in the comments of the generator `ReferencesConfig`.

RedbackThomson (changes requested):

templates/pkg/resource/references.go.tpl
vijtrip2: Actually, I do need the ServicePackageName i.e. "Used by the ACK controller's for their name"

pkg/generate/config/field.go
vijtrip2: I will update the comment to mention the ServicePackagePath.

This field is an abstraction that helps service teams by not having them enter complete api-group string. ACK code-generator creates the api-group string from the "ServiceName" field.

I feel it is more readable/user-friendly to leave this as simple "ServiceName" and the comment will explain that it actually is ServicePackageName from `aws-sdk-go`. Let me know your thoughts.

Even if users make a mistake like `service_name: opensearch` instead of `opensearchservice`, the error will be caught at compile time.

RedbackThomson: I agree that it's more user friendly, sure. Please just update the code comments to make it clear for us, then 👍🏼

RedbackThomson:
Thank you!
/approve

templates/pkg/resource/references.go.tpl
```diff
@@ -18,10 +18,28 @@ import (
 	ackerr "github.com/aws-controllers-k8s/runtime/pkg/errors"
 {{ end -}}
 	acktypes "github.com/aws-controllers-k8s/runtime/pkg/types"
-
```
jaypipes, line 21: keep this separate newline please. it demarcates third-part imports (which the ACK runtime is) from "local" imports (which the specific controller is)

templates/pkg/resource/references.go.tpl
jaypipes, line 27: I presume for now we are requiring the controller's `go.mod` files to be manually updated with the imported referenced service controller's package?

vijtrip2: yes. I also mentioned it in the proposal PR. https://github.com/aws-controllers-k8s/community/pull/1092/files

jaypipes (approved):
w00t, thanks @vijtrip2 :)

PR #289, opened 2022-02-11
Refactor `SetResourceForStruct` to iterate over `targetShape` members

brycahta:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1146

After a recent update to `/pkg/model/field.go`, shapes _may_ contain **additional** fields/members with a "nested" format such as **DHCPOptions.Fields**: `[DHCPConfigurations TagSpecifications.Tags.Value Tags.Value TagSpecifications DHCPConfigurations.Values TagSpecifications.ResourceType TagSpecifications.Tags Tags.Key DHCPOptionsID OwnerID DHCPConfigurations.Key TagSpecifications.Tags.Key DryRun Tags]` The purpose of that update is to allow users to easily fetch nested shapes and their corresponding configurations using the "nested" field name as the key.

However, `SetResourceForStruct` will never have the opportunity to process and/or fetch shapes and configs for these "nested" fields because the algorithm processes fields **iff they exist in sourceShape**. *Note, sourceShape comes directly from `aws-sdk` in this context, so will NOT have these additional, nested fields.*

This patch iterates over `targetShape.Members` instead of `sourceShape` so these "nested" fields can be included in the lookup logic, but continues to use `sourceShape` index for consistency throughout codebase. There's no change in functionality, but this paves the way for a follow-up PR to check a targetShape's nested Field `SetConfig` if the Field cannot be found in the sourceShape: 

```
sourceMemberShapeRef := sourceShape.MemberRefs[targetMemberName]
		if sourceMemberShapeRef == nil {
			continue
		}

                // Future patch: check if targetMemberName has SetConfig logic that needs to be processed
```

### DIFF
--- a/pkg/generate/code/common.go
+++ b/pkg/generate/code/common.go
@@ -14,6 +14,7 @@
 package code
 
 import (
+	"fmt"
 	"strings"
 
 	awssdkmodel "github.com/aws/aws-sdk-go/private/model/api"
@@ -187,4 +188,15 @@ func FindPrimaryIdentifierFieldNames(
 		)
 	}
 	return crField, shapeField
+}
+
+// GetMemberIndex returns the index of memberName by iterating over
+// shape's slice of member names for deterministic ordering
+func GetMemberIndex(shape *awssdkmodel.Shape, memberName string) (int, error) {
+	for sourceIndex, sourceMemberName := range shape.MemberNames() {
+		if strings.EqualFold(sourceMemberName, memberName) {
+			return sourceIndex, nil
+		}
+	}
+	return -1, fmt.Errorf("Could not find %s in shape %s", memberName, shape.ShapeName)
 }
--- a/pkg/generate/code/common.go
+++ b/pkg/generate/code/common.go
@@ -193,9 +193,9 @@ func FindPrimaryIdentifierFieldNames(
 // GetMemberIndex returns the index of memberName by iterating over
 // shape's slice of member names for deterministic ordering
 func GetMemberIndex(shape *awssdkmodel.Shape, memberName string) (int, error) {
-	for sourceIndex, sourceMemberName := range shape.MemberNames() {
-		if strings.EqualFold(sourceMemberName, memberName) {
-			return sourceIndex, nil
+	for index, shapeMemberName := range shape.MemberNames() {
+		if strings.EqualFold(shapeMemberName, memberName) {
+			return index, nil
 		}
 	}
 	return -1, fmt.Errorf("Could not find %s in shape %s", memberName, shape.ShapeName)
--- a/pkg/generate/code/set_resource.go
+++ b/pkg/generate/code/set_resource.go
@@ -1226,14 +1226,20 @@ func SetResourceForStruct(
 	sourceShape := sourceShapeRef.Shape
 	targetShape := targetShapeRef.Shape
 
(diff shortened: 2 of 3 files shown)


pkg/generate/code/common.go
```diff
@@ -188,3 +189,14 @@ func FindPrimaryIdentifierFieldNames(
...
+// GetMemberIndex returns the index of memberName by iterating over
+// shape's slice of member names for deterministic ordering
+func GetMemberIndex(shape *awssdkmodel.Shape, memberName string) (int, error) {
+	for index, shapeMemberName := range shape.MemberNames() {
+		if strings.EqualFold(shapeMemberName, memberName) {
+			return index, nil
+		}
+	}
+	return -1, fmt.Errorf("Could not find %s in shape %s", memberName, shape.ShapeName)
+}
```
jaypipes, line 202: Nice job with this. It will allow the index numbers in variable names to remain consistent, cutting down on unnecessary code changes in generated controllers. thank you!

pkg/generate/code/set_resource.go
```diff
@@ -1226,53 +1226,61 @@ func SetResourceForStruct(
 	sourceShape := sourceShapeRef.Shape
 	targetShape := targetShapeRef.Shape
 
-	for memberIndex, memberName := range sourceShape.MemberNames() {
-		targetMemberShapeRef := targetShape.MemberRefs[memberName]
-		if targetMemberShapeRef == nil {
+	for _, targetMemberName := range targetShape.MemberNames() {
```
jaypipes, line 1229: This will be just fine for now (and is a perfect stepping stone for future commits, as you allude to in your PR summary). Long-term, I'd be keen to change this to loop over `r.Fields[targetMemberName].MemberFields`, though, which will contain all the additional custom fields as well.

pkg/generate/code/set_resource.go
```diff
@@ -1226,53 +1226,61 @@ func SetResourceForStruct(
...
+		if sourceMemberShapeRef == nil {
 			continue
 		}
-		memberVarName := fmt.Sprintf("%sf%d", targetVarName, memberIndex)
-		memberShapeRef := sourceShape.MemberRefs[memberName]
-		memberShape := memberShapeRef.Shape
-		cleanNames := names.New(memberName)
-		sourceAdaptedVarName := sourceVarName + "." + memberName
+		// Upstream logic iterates over sourceShape members and therefore uses
+		// the sourceShape's index; continue using sourceShape's index here for consistency.
```
jaypipes, line 1235: :+1:

pkg/generate/code/set_resource.go
```diff
@@ -1226,53 +1226,61 @@ func SetResourceForStruct(
...
+		sourceMemberIndex, err := GetMemberIndex(sourceShape, targetMemberName)
+		if err != nil {
+			continue
```
jaypipes, line 1238: change this to a `panic` so we can get early failure in code generation. I have a feeling it would be super tricky to troubleshoot a problem here if we just ignore and `continue`

jaypipes (changes requested):
@brycahta this is fantastic, thank you! One teeny little request to `panic` instead of `continue` when we cannot determine a source shape index, but otherwise this is ready to ship. A++

jaypipes (approved):
Sweet. thanks again, @brycahta! :)

PR #292, opened 2022-02-15
Use `SetConfig` in `SetResourceForStruct` when target shape has no members

brycahta:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1146

This is a follow-up to https://github.com/aws-controllers-k8s/code-generator/pull/289. A `targetShape` without members in `SetResourceForStruct` will not be processed today; however, this is **not** desired behavior for all APIs such as EC2's DHCPOptions resource. 

This patch will look up and consume `SetConfig` in `SetResourceForStruct` when `targetShape` has no members like when DHCPOptions' field, `DHCPConfigurations.Values []*string`, aligns (by name) with aws-sdk's output shape, `DhcpConfigurations.Values []*AttributeValue`. Instead of skipping the processing of `Values []*string` field, it will check if a `SetConfig` exists and if it's valid (referencing a *real* shape in sourceShape), then use that updated field to source the target field's value resulting in using `DhcpConfigurations.Values.Value *string`  to populate the resource's `DHCPConfigurations.Values []*string`.


pkg/generate/code/set_resource.go
```diff
@@ -1290,6 +1312,52 @@ func SetResourceForStruct(
...
+
+		// Check if target field has a SetConfig, validate SetConfig.From points
+		// to a shape within sourceShape, and generate Go code using
+		// said shape. Using the example above, SetConfig is set
+		// for NewDhcpConfiguration.Values and Setconfig.From points
+		// to AttributeValue.Value (string), which leads to generating Go
+		// code referencing DhcpConfiguration.Values.Value instead of 'Values'.
+
+		if targetField, ok := r.Fields[targetFieldPath]; ok {
+			setCfg := targetField.GetSetterConfig(model.OpTypeList)
```
brycahta, line 1333: meant to pass `op` param here-- will update in next revision

pkg/generate/code/set_resource.go
```diff
@@ -1290,6 +1312,52 @@ func SetResourceForStruct(
...
+					names := names.New(sourceMemberShapeRef.LocationName)
+					sourceAdaptedVarName = sourceVarName + "." + names.Camel
+					out += fmt.Sprintf(
+						"%sif %s != nil {\n", indent, sourceAdaptedVarName,
+					)
+					qualifiedTargetVar = targetVarName
+
+					// Use setResourceForScalar and dereference sourceAdaptedVarName
+					// because primitives are being set.
+					sourceAdaptedVarName = "*" + sourceAdaptedVarName
```
jaypipes, line 1347: This is fine for now, but I recommend we clean up the [`setResourceForScalar()`](https://github.com/aws-controllers-k8s/code-generator/blob/930850fc31c40fa7d44f2544ce6e676baab05c27/pkg/generate/code/set_resource.go#L1471-L1495) method implementation to be able to properly handle this only using the sourceShapeRef information.

pkg/generate/code/set_resource.go
```diff
@@ -1310,6 +1378,9 @@ func setResourceForSlice(
 	sourceVarName string,
 	// ShapeRef of the source slice field
 	sourceShapeRef *awssdkmodel.ShapeRef,
+	// targetFieldPath is the absolute path to targetFieldName
```
jaypipes, line 1381: it's not an absolute path. it's a field path :)

jaypipes (changes requested):
Correct that one hard-coded OpTypeList thing and I'm good with this, thank you @brycahta!

pkg/testdata/models/apis/ec2/0000-00-00/generator.yaml
```diff
@@ -68,8 +68,11 @@ operations:
     output_wrapper_field_path: VpcEndpoint
 
 resources:
-  DHCPOptions:
-
+  DhcpOptions:
+    fields:
+      DHCPConfigurations.Values:
+        set:
+          - from: AttributeValue.Value
```
jaypipes, line 75: You might consider adding:

```yaml
      method: CREATE
```

here in order to exercise the OpType passing down into the recursive SetResourceForXXX functions.

jaypipes (approved):
:+1:

PR #295, opened 2022-02-18
Field Config Fixes for Code Generator

nmvk:
- Support for nested list of secrets
  - Code was getting generated for single SecretRef

Generated code with the fix

```
             authenticationMode:
                description: Denotes the user's authentication properties, such as
                  whether it requires a password to authenticate.
                properties:
                  passwords:
                    items:
                      description: SecretKeyReference combines a k8s corev1.SecretReference
                        with a specific key within the referred-to Secret
                      properties:
                        key:
                          description: Key is the key within the secret
                          type: string
                        name:
                          description: Name is unique within a namespace to reference
                            a secret resource.
                          type: string
                        namespace:
                          description: Namespace defines the space within which the
                            secret name must be unique.
                          type: string
                      required:
                      - key
                      type: object
                    type: array
                  type_:
                    type: string
                type: object
```

- Fix the struct compare logic
  - Spec was appended to field path due to this compare
    config was ignored.
https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/code/compare.go#L128

Issue #, if available:

Description of changes:


pkg/generate/code/compare.go
```diff
@@ -562,7 +562,7 @@ func CompareStruct(
 		secondResAdaptedVarName := secondResVarName + "." + memberNameClean
 
 		var compareConfig *ackgenconfig.CompareFieldConfig
-		fieldConfig := fieldConfigs[memberFieldPath]
+		fieldConfig := fieldConfigs[memberFieldPath[5:len(memberFieldPath)]]
```
jaypipes, line 565: What's this for? (Note: if I'm asking this question, pretty sure other readers of the code will ask the same question without a code comment explaining why... :) )

pkg/model/model.go
```diff
@@ -541,7 +541,7 @@ func replaceSecretAttrGoType(
 		)
 		panic(msg)
 	}
-	attr.GoType = "*ackv1alpha1.SecretKeyReference"
+	attr.GoType = field.GoType
```
jaypipes, line 544: :+1:

jaypipes (changes requested):

pkg/generate/code/compare.go
nmvk: Agreed, updated it

memberFieldPath contains the field path along with the prefix `Spec.` hence we would need to substring to exclude `Spec.` to get correct field config.

https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/code/compare.go#L128 adds `Spec.` prefix.

pkg/generate/code/compare.go
```diff
@@ -562,7 +562,9 @@ func CompareStruct(
 		secondResAdaptedVarName := secondResVarName + "." + memberNameClean
 
 		var compareConfig *ackgenconfig.CompareFieldConfig
-		fieldConfig := fieldConfigs[memberFieldPath]
+		// memberFieldPath contains the field path along with the prefix "Spec." hence we
+		// would need to substring to exclude "Spec." to get correct field config.
```
jaypipes, line 566: I see... A more future proof way of doing this would be to remove the length of the [`cfg.PrefixConfig.SpecField`](https://github.com/aws-controllers-k8s/code-generator/blob/386dc0d66152f1779e4bb315d153be5412361123/pkg/generate/config/config.go#L71) value. For targets like Crossplane, they use [`Spec.ForProvider`](https://github.com/aws-controllers-k8s/code-generator/blob/386dc0d66152f1779e4bb315d153be5412361123/pkg/generate/crossplane/config.go#L21) instead of `Spec`, which is why this configuration field exists. :)

jaypipes (changes requested):

pkg/generate/code/compare.go
nmvk: Thanks! updated.

jaypipes (approved):
:tada: :)

PR #297, opened 2022-02-23
Add `make` targets for local builds

RedbackThomson:
Description of changes:
Adds a `go.local.mod`, and make targets using the new file, to more easily reference the local runtime package. Developers are now able to run `make local-build-ack-generate` to build and test changes to `code-generator` when updating `runtime`. 


vijtrip2 (commented):
Good improvement.

Any means of keeping the dependencies in go.local.mod and go.local.sum consistent with dependency updates in go.mod and go.sum.

Currently for code-generation, I just need to add a "replace" line in `go.mod` and just use existing make targets. With the new approach do you think there can be hidden surprises if go.local.mod and go.local.sum are not kept consistent.

RedbackThomson:
> With the new approach due you think there can be hidden surprises if go.local.mod and go.local.sum are not kept consistent

Hmm yeah if any of the other direct dependencies are not updating in the `go.local.mod` file, then it won't know to update those to be inline with `go.mod`. Not really sure how to reference the modules from `go.mod`, though. It's hard to google this stuff. So for now you would need to update dependencies in both places :/

Makefile
```diff
@@ -28,13 +30,20 @@ build-ack-generate:	## Build ack-generate binary
 	@go build ${GO_TAGS} ${GO_LDFLAGS} -o bin/ack-generate cmd/ack-generate/main.go
 	@echo "ok."
 
+local-build-ack-generate:	## Build ack-generate binary using the local go.mod
```
brycahta, line 33: nit: using the file name is clearer, `Build ack-generate binary using go.local.mod`

brycahta:
I agree that keeping these 2 `go.mod` in sync will be a pain. One idea is could the `make` target be a bash script that edits `go.mod` file directly to make it "localized"? And running `make build-controller` would undo the edits

RedbackThomson:
> One idea is could the make target be a bash script that edits go.mod file directly to make it "localized"

No doubt someone will commit this change and include it in a PR. I don't think our `go.mod` changes too much that this will legitimately be too annoying. I just checked the file history and we essentially haven't changed anything apart from the runtime version since `v0.4.0` or even before then.

vijtrip2 (approved):

PR #301, opened 2022-03-09
BugFix: fixes duplicate import inside references.go

vijtrip2:
Description of changes:

* Use a set of referenced service names for generating import statements inside `references.go.tpl`
* Use sorted fieldnames inside iteration for generating consistent code for `references.go.tpl`. Earlier the generated code was similar but not identical to previous generation. This will reduce cognitive load when reviewing auto-generated PRs

* Added cross service reference manifests for unit testing.
* Manually generated apigatewayv2 controller to test as well.

### DIFF
--- a/pkg/generate/code/resource_reference.go
+++ b/pkg/generate/code/resource_reference.go
@@ -35,7 +35,9 @@ func ReferenceFieldsValidation(
 	indentLevel int,
 ) string {
 	out := ""
-	for _, field := range crd.Fields {
+	// Sorted fieldnames are used for consistent code-generation
+	for _, fieldName := range crd.SortedFieldNames() {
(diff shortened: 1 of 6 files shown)


pkg/generate/code/resource_reference.go
```diff
@@ -35,7 +35,9 @@ func ReferenceFieldsValidation(
 	indentLevel int,
 ) string {
 	out := ""
-	for _, field := range crd.Fields {
+	// Sorted fieldnames are used for consistent code-generation
+	for _, fieldName := range crd.SortedFieldNames() {
```
RedbackThomson, line 39: 👍🏼

RedbackThomson (approved):
Totally makes sense. Good catch!

/lgtm

PR #303, opened 2022-03-10
Fix reference import templating

RedbackThomson:
Description of changes:
The current implementation attempts to put all service type imports on the same line. This pull request ensures each import gets a separate line per service and compiles correctly.


vijtrip2 (commented):
/lgtm

PR #304, opened 2022-03-11
Consider renaming when trying to find pluralized identifier shapes

A-Hilaly:
Description of changes:

Currently `FindPluralizedIdentifiersInShape` doesn't take into
consideration CRD field renames. This patch enhances the 
function to be able to lookup field renames and properly find the
matching identifier field.

Fixes https://github.com/aws-controllers-k8s/community/issues/1208


RedbackThomson (approved):
Good catch!
/lgtm

PR #308, opened 2022-03-21
Nil check for IsSync

nmvk:
Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/1231

Description of changes:
Nil check fix for IsSync needed for MemoryDB and
Fixed a compare ignore bug.

### DIFF
--- a/pkg/generate/code/synced.go
+++ b/pkg/generate/code/synced.go
@@ -121,7 +121,12 @@ func scalarFieldEqual(
 ) string {
 	out := ""
 	fieldPath := fmt.Sprintf("%s.%s", resVarName, *condCfg.Path)
-
+	// if r.ko.Status.Status == nil
+	out += fmt.Sprintf("\tif %s == nil {\n", fieldPath)
+	// return false, nil
+	out += "\t\treturn false, nil\n"
+	// }
+	out += "\t}\n"
 	valuesSlice := ""
 	switch goType {
 	case "string":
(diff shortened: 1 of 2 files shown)


A-Hilaly (approved):
Lookgs good! thanks @nmvk

pkg/generate/code/compare.go
```diff
@@ -564,7 +564,8 @@ func CompareStruct(
 		var compareConfig *ackgenconfig.CompareFieldConfig
 		// memberFieldPath contains the field path along with the prefix cfg.PrefixConfig.SpecField + "." hence we
 		// would need to substring to exclude cfg.PrefixConfig.SpecField + "." to get correct field config.
-		fieldConfig := fieldConfigs[memberFieldPath[len(cfg.PrefixConfig.SpecField) + 1 :len(memberFieldPath)]]
+		specFieldLen := len(strings.TrimPrefix(cfg.PrefixConfig.SpecField, "."))
```
A-Hilaly, line 567: Can we have this change in a different PR? I know it's a small change but it will be messy if we ever have to revert this commit.

A-Hilaly (changes requested):

vijtrip2 (approved):
Thanks for quick fix.

Would this handle multi-level nil check?

Ex: `ko.Status.FieldA.SubFieldB` should have nil check for both FieldA and SubFieldB

A-Hilaly:
@vijtrip2 the generated code is fine for nested fields. It looks like the issue is only with top level fields
```go
	if r.ko.Spec.ProvisionedThroughput == nil {
		return false, nil
	}
	if r.ko.Spec.ProvisionedThroughput.ReadCapacityUnits == nil {
		return false, nil
	}
	provisionedThroughputCandidates := []int{0, 10}
	if !ackutil.InStrings(*r.ko.Spec.ProvisionedThroughput.ReadCapacityUnits, provisionedThroughputCandidates) {
		return false, nil
	}
```

pkg/generate/code/compare.go
nmvk: Done

vijtrip2 (commented):
/lgtm

PR #309, opened 2022-03-21
Generate `Region()` method for `resourceIdentifiers`

A-Hilaly:
Follow up to https://github.com/aws-controllers-k8s/runtime/pull/79

This patch modifies the code-generator to generate a `Region` function
so that a resource can also satisfy the `Identifiers` interface. This
the patch also extends the `setStatusDefaults` so that it correctly
populates the `status.ResourceMetadata.Region` with the aws region in
which it is created/managed

Description of changes:
- Add `Region()` method to resourceIdentifiers template.
- Modifies `setStatusDefaults` to  be able to set the a default region.
- Update `runtime_test.go`

### DIFF
--- a/pkg/generate/ack/runtime_test.go
+++ b/pkg/generate/ack/runtime_test.go
@@ -46,6 +46,11 @@ func (ids *fakeIdentifiers) OwnerAccountID() *ackv1alpha1.AWSAccountID {
 	return &owner
 }
 
+func (ids *fakeIdentifiers) Region() *ackv1alpha1.AWSRegion {
+	region := ackv1alpha1.AWSRegion("us-west-2")
+	return &region
+}
+
 type fakeDescriptor struct{}
 
 func (fd *fakeDescriptor) GroupKind() *metav1.GroupKind {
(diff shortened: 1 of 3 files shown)


templates/pkg/resource/identifiers.go.tpl
```diff
@@ -31,3 +31,12 @@ func (ri *resourceIdentifiers) OwnerAccountID() *ackv1alpha1.AWSAccountID {
 	}
 	return nil
 }
+
+// Region returns the AWS region in which the resource should be managed, or
```
RedbackThomson, line 35:
```suggestion
// Region returns the AWS region in which the resource exists or will exist, or
```

RedbackThomson (approved):
Just one nit to bring it in-line with the latest runtime updates

jaypipes (approved):
:+1:

pkg/generate/ack/runtime_test.go
```diff
@@ -46,6 +46,11 @@ func (ids *fakeIdentifiers) OwnerAccountID() *ackv1alpha1.AWSAccountID {
 	return &owner
 }
 
+func (ids *fakeIdentifiers) Region() *ackv1alpha1.AWSRegion {
```
vijtrip2, line 49:
```suggestion
func (ids *fakeIdentifiers) Region() *ackv1alpha1.Region {
```

A-Hilaly: I kept it `AWSRegion` to be similar to the other variables `AWSAccountID` and `AWSResourceName`: https://github.com/aws-controllers-k8s/runtime/blob/main/apis/core/v1alpha1/common.go#L16-L17

PR #310, opened 2022-03-21
Fix Spec field length

nmvk:
Updated to trim the prefix "." to calculate the length of the `Spec`.

`cfg.PrefixConfig.SpecField` value is `.Spec` so old code `len(cfg.PrefixConfig.SpecField) + 1` constructed incorrect field path and ignored the config set in `generator.yaml`

Added unit test to validate that below block is not generated.

```
if !ackcompare.SliceStringPEqual(a.ko.Spec.AuthenticationMode.Passwords, b.ko.Spec.AuthenticationMode.Passwords) {
	delta.Add("Spec.AuthenticationMode.Passwords", a.ko.Spec.AuthenticationMode.Passwords, b.ko.Spec.AuthenticationMode.Passwords)
}
```

### DIFF
--- a/pkg/generate/code/compare.go
+++ b/pkg/generate/code/compare.go
@@ -564,7 +564,8 @@ func CompareStruct(
 		var compareConfig *ackgenconfig.CompareFieldConfig
 		// memberFieldPath contains the field path along with the prefix cfg.PrefixConfig.SpecField + "." hence we
 		// would need to substring to exclude cfg.PrefixConfig.SpecField + "." to get correct field config.
-		fieldConfig := fieldConfigs[memberFieldPath[len(cfg.PrefixConfig.SpecField) + 1 :len(memberFieldPath)]]
+		specFieldLen := len(strings.TrimPrefix(cfg.PrefixConfig.SpecField, "."))
+		fieldConfig := fieldConfigs[memberFieldPath[specFieldLen + 1: len(memberFieldPath)]]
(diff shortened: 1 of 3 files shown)


pkg/generate/code/compare.go
```diff
@@ -564,7 +564,8 @@ func CompareStruct(
 		var compareConfig *ackgenconfig.CompareFieldConfig
 		// memberFieldPath contains the field path along with the prefix cfg.PrefixConfig.SpecField + "." hence we
 		// would need to substring to exclude cfg.PrefixConfig.SpecField + "." to get correct field config.
-		fieldConfig := fieldConfigs[memberFieldPath[len(cfg.PrefixConfig.SpecField) + 1 :len(memberFieldPath)]]
+		specFieldLen := len(strings.TrimPrefix(cfg.PrefixConfig.SpecField, "."))
```
A-Hilaly, line 567: This fix makes sense, i still wonder what part of elasticache/memorydb delta generated code was broken? can we add some unit tests? :)

RedbackThomson:
Echoing @A-Hilaly here - Please update the PR to include an issue describing the bug this is fixing, or describe the error directly in the PR body. I don't have any context for what this line of code is doing, or should be doing.

pkg/generate/code/compare.go
nmvk: I added a new unit test :)

PR #311, opened 2022-03-22
add ServicePackageName to ack-user-config and ack-user secrets

acornett21:
Issue #, if available:
- Fixes: aws-controllers-k8s/community#1149

Description of changes:
Add in support for service specific configmaps and secret so that a cluster admin can configure each service differently, or the same if they so choose.

### DIFF
--- a/templates/config/controller/user-env.yaml.tpl
+++ b/templates/config/controller/user-env.yaml.tpl
@@ -10,8 +10,8 @@ spec:
       - name: controller
         envFrom:
           - configMapRef:
-              name: ack-user-config
+              name: ack-{{.ServicePackageName}}-user-config
               optional: false
           - secretRef:
-              name: ack-user-secrets
+              name: ack-{{.ServicePackageName}}-user-secrets
               optional: false


komish:
This seems reasonable enough. It does, however, mean that administrators who want the same configuration for all installed controllers would effectively have to duplicate configmaps and secrets, right? Was there a user request requesting the feature? 

Either way, this looks like it should accomplish this and seems useful 👍

RedbackThomson:
> This seems reasonable enough. It does, however, mean that administrators who want the same configuration for all installed controllers would effectively have to duplicate configmaps and secrets, right? Was there a user request requesting the feature?
> 
> Either way, this looks like it should accomplish this and seems useful 👍

The current method means that you cannot install multiple controllers with separate permissions, though - which is arguably a larger security concern. I think this PR does a good job of trading off that security concern for your ease of use concern.

acornett21:
> This seems reasonable enough. It does, however, mean that administrators who want the same configuration for all installed controllers would effectively have to duplicate configmaps and secrets, right? Was there a user request requesting the feature?

Yeah you are correct, it's a trade off, they would need to duplicate if they wanted the same configuration. And yes, this granularity was asked for by a user int he Kube slack channel.

jaypipes (approved):
:+1: doh, sorry for the delay in review/merge @acornett21!

PR #314, opened 2022-03-29
Support for nested resource references

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1212

This PR aims to add support for adding ResourceReferences for nested fields. I have tested these changes locally by generating and compiling *eks-controller* and *apigatewayv2-controller*, and also added unit tests.

In a nut-shell, this PR mainly updates the reference resolution to use `field.Path` instead of `field.Names` with proper nil checks. This PR also updates the TypeDefs to add new *AWSResourceReference* attribute, instead of only supporting Reference fields at top-level. More detailed desciption about the changes is present below.

Note: I have broken all the changes into multiple commits and i recommend to review this PR commit-by-commit. : )

Description of changes:
* Add new reusable go code generator method for producing Nil check logical condition on a *fieldPath*
* Wrap the *ReferenceFieldValidation* with proper Nil check for a fieldPath and correctly use *field.Path* for accessing the value instead of `field.Names.Camel`
* Include Nil checks inside `ReferenceFieldPresent` method for the `field.Path` and capture the ReferenceFieldPresent condition for each field inside parentheses to make it more readable.
* Make the ResourceFieldConfig lookup case-insensitive for better developer experience . This will allow developers to not worry about case-sensitivity when creating FieldConfig keys inside `generator.yaml`
* Update `crd.HasReference` method to work for nested fields as well.
* Update *TypeDef* generation inside `model.go` to also include new *AWSResourceReference* attribute if a nested field *ReferencesConfig* is present.
* Update `references.go.tpl` to correctly use *field.Path* instead of *field.Names* to resolve references.
* Update `references.go.tpl` to support string type-alias by performing type-casting before resolving a reference.(issue# https://github.com/aws-controllers-k8s/community/issues/1213)


pkg/model/model.go
```diff
@@ -459,44 +459,78 @@ func (m *Model) processNestedFieldTypeDefs(
...
 	if fp.Size() < 2 {
 		// TypeDef should only be updated with Reference fields for nested fieldPath.
 		// For non-nested fieldPath, the references are added directly to the resource
 		// Spec.
 		return
 	}
-	fieldName := fp.Pop()
-	// parent of fieldName determines the TypeDef
-	parentFieldName := fp.Pop()
+	// First part of nested reference fieldPath is the name of Spec field
```
jaypipes, line 478: When you say "is the name of the Spec field", are you referring to the field *within* the `Spec` field? i.e. the "top-level field" in the Spec?

Or are you referring to `Config.PrefixConfig.SpecField`? :) Because, you removed the call to `fp.Pop()` above and therefore the "first" field in the fieldpath is actually going to be the `Spec` itself...

pkg/model/model_eks_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	for _, td := range tds {
+		if td != nil && strings.EqualFold(td.Names.Original, "vpcConfigRequest") {
+			vpcConfigRequestTD = td
+			break
+		}
+	}
+	assert.NotNil(t, vpcConfigRequestTD)
+	securityGroupIdsAttr := vpcConfigRequestTD.GetAttributeIgnoreCase("SecurityGroupIds")
+	securityGroupRefsAttr := vpcConfigRequestTD.GetAttributeIgnoreCase("SecurityGroupRefs")
```
jaypipes, line 51: You may want to add:

```go
require.NotNil(securityGroupIdsAttr)
require.NotNil(securityGroupRefsAttr)
```

here, otherwise you will get nil-pointer access exceptions in the assertions below...

pkg/model/model_eks_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/aws-controllers-k8s/code-generator/pkg/model"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/aws-controllers-k8s/code-generator/pkg/testutil"
+)
+
+func TestEKS_WithNestedReference(t *testing.T) {
+	_ = assert.New(t)
```
jaypipes, line 29: this doesn't really make sense :) You should do something like this:

```go
assert := assert.New(t)
```

and then later on in the test, you don't need to supply the `t` argument when calling `assert.Equal()`..

pkg/model/model_eks_test.go
```diff
@@ -0,0 +1,56 @@
...
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package model_test
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/aws-controllers-k8s/code-generator/pkg/model"
```
jaypipes, line 20: please group the "local" imports together, below the 3rd party imports.

jaypipes (changes requested):
@vijtrip2 looking pretty good so far, thank you! Left some notes inline for you to address on your next round of revisions :)

pkg/model/field.go
```diff
@@ -178,6 +178,21 @@ func (f *Field) ReferencedResourceNamePlural() string {
...
+	fieldPathPrefix := strings.TrimSuffix(f.Path, f.Names.Camel)
+	return fmt.Sprintf("%s%s", fieldPathPrefix, f.GetReferenceFieldName().Camel)
+}
+
+// CamelCasedFieldPath replaces the period in fieldPath with
+// underscore. This method is useful for generating go method
+// name from the fieldPath.
+func (f *Field) CamelCasedFieldPath() string {
+	return strings.ReplaceAll(f.Path, ".", "_")
+}
```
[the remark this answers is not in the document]
vijtrip2: I renamed the method but keeping it on *model.Field instead of fieldPath because I need to use this method inside references.go.tpl template. 
The suggested refactor can be used inside go template if model.field.Path return fieldPath.Path instead of string but I believe that can be tackled in future PR if i need to use this method again.

pkg/model/model.go
vijtrip2: It's the former. I have updated the comment to be more precise.
> // First part of nested reference fieldPath is the name of top level Spec
> // field. Ex: For 'ResourcesVpcConfig.SecurityGroupIds' fieldpath the
> // specFieldName is 'ResourcesVpcConfig'

The fieldpath passed in this method is the key of ResourceConfig.Fields, which does not include Spec prefix.

pkg/model/model_eks_test.go
vijtrip2: Done. It was a leftover that i forgot to clean

jaypipes (approved):
Very nice work, Vijay. ++

brycahta (approved):
/lgtm

PR #315, opened 2022-03-29
updating olm kustomize to remove default values from env

acornett21:
Issue #, if available:
- Fixes: aws-controllers-k8s/community#1238

Description of changes:
Updating the olm kustomize file to go back to previous state before default values were added into the `env` section of the `deployment.yaml.tpl`. Ideally it would be best to be able to keep these defaults, but even moving everything into the `env` and away from the `envFrom` did not work properly in a cluster, so I feel this is the best option that we have.

### DIFF
--- a/templates/config/controller/olm-kustomization.yaml.tpl
+++ b/templates/config/controller/olm-kustomization.yaml.tpl
@@ -1,4 +1,22 @@
 {{ template "controller_kustomization" . }}
 
+patchesJson6902:
+  - target:
+      group: apps
+      version: v1
+      kind: Deployment
+      name: ack-{{ .ServicePackageName }}-controller
+    patch: |-
+        - op: replace
+          path: '/spec/template/spec/containers/0/env'
+          value: []
+        - op: add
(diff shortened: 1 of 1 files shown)


RedbackThomson (approved):
Thank you for figuring this out. It's a shame we can't do an intelligent merge strategy, but I'm happy with this for now. It just means we have to be aware of the OLM user config file whenever we add an environment variable.

acornett21:
> Thank you for figuring this out. It's a shame we can't do an intelligent merge strategy, but I'm happy with this for now. It just means we have to be aware of the OLM user config file whenever we add an environment variable.

@RedbackThomson I'm pretty sad `merge` did not work as well, when I moved everything into the `env` section ie the defaults plus the values from the config map, it resulted in this format

```
- name: ACK_LOG_LEVEL
          value: info
          valueFrom:
            configMapKeyRef:
              key: ACK_LOG_LEVEL
              name: ack-user-config
```

Which does not work in a cluster.

PR #316, opened 2022-03-29
Remove patch version from Dockerfile Golang base

RedbackThomson:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1241

Description of changes:
Removes the patch version from the Golang base image in our Dockerfiles. This should now use the latest Go 1.17 patch when building images - meaning we can capture the latest security fixes as soon as they become available in the bitnami image.


vijtrip2 (commented):
/lgtm

PR #318, opened 2022-04-05
Set Terminal Condition if ACK Terminal Error

nmvk:
Set terminal condition if ACK Terminal Error is thrown.

ACK Terminal errors can be used to set terminal conditions
if controller validations fail.

```
	if *latestStatus == "create-failed" {
		return nil, ackerr.TerminalError.New(errors.New("cluster cannot be updated as its status is not 'available'."))
	}

```

### DIFF
--- a/templates/pkg/resource/sdk.go.tpl
+++ b/templates/pkg/resource/sdk.go.tpl
@@ -4,6 +4,7 @@ package {{ .CRD.Names.Snake }}
 
 import (
 	"context"
+	"errors"
 	"reflect"
 	"strings"
 
@@ -218,8 +219,8 @@ func (rm *resourceManager) updateConditions (
 			syncCondition = condition
 		}
 	}
-
-	if rm.terminalAWSError(err) || err ==  ackerr.SecretTypeNotSupported || err == ackerr.SecretNotFound {
+	var termError *ackerr.TerminalError
+	if rm.terminalAWSError(err) || err ==  ackerr.SecretTypeNotSupported || err == ackerr.SecretNotFound || errors.As(err, &termError) {
 		if terminalCondition == nil {
 			terminalCondition = &ackv1alpha1.Condition{
 				Type:   ackv1alpha1.ConditionTypeTerminal,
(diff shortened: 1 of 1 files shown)


templates/pkg/resource/sdk.go.tpl
```diff
@@ -4,6 +4,7 @@ package {{ .CRD.Names.Snake }}
 
 import (
 	"context"
+	"github.com/pkg/errors"
```
RedbackThomson, line 7: Should just be `"errors"`

templates/pkg/resource/sdk.go.tpl
```diff
@@ -218,16 +219,16 @@ func (rm *resourceManager) updateConditions (
 			syncCondition = condition
 		}
 	}
-
-	if rm.terminalAWSError(err) || err ==  ackerr.SecretTypeNotSupported || err == ackerr.SecretNotFound {
+	var termError ackerr.TerminalError
+	if rm.terminalAWSError(err) || err ==  ackerr.SecretTypeNotSupported || err == ackerr.SecretNotFound || errors.As(err, &termError) {
```
RedbackThomson, line 223: Instead, use `if errors.Is(err, ackerr.TerminalError)` :)

(Honestly these other error checks should also be using that syntax)

templates/pkg/resource/sdk.go.tpl
```diff
@@ -218,16 +219,16 @@ func (rm *resourceManager) updateConditions (
...
 		if terminalCondition == nil {
 			terminalCondition = &ackv1alpha1.Condition{
 				Type:   ackv1alpha1.ConditionTypeTerminal,
 			}
 			ko.Status.Conditions = append(ko.Status.Conditions, terminalCondition)
 		}
 		var errorMessage = ""
-		if err == ackerr.SecretTypeNotSupported || err == ackerr.SecretNotFound {
+		if err == ackerr.SecretTypeNotSupported || err == ackerr.SecretNotFound || errors.As(err, &termError) {
```
RedbackThomson, line 231: Same here

RedbackThomson (changes requested):

templates/pkg/resource/sdk.go.tpl
RedbackThomson: Oops. Thank you for bringing to my attention that `errors.Is` works on value comparison, and not type assertion. `errors.As` is the proper use here.

templates/pkg/resource/sdk.go.tpl
```diff
@@ -218,16 +219,16 @@ func (rm *resourceManager) updateConditions (
 			syncCondition = condition
 		}
 	}
-
-	if rm.terminalAWSError(err) || err ==  ackerr.SecretTypeNotSupported || err == ackerr.SecretNotFound {
+	var termError ackerr.TerminalError
```
RedbackThomson, line 222: Should be an empty pointer?:
```suggestion
	var termError *ackerr.TerminalError
```

PR #321, opened 2022-04-06
add real versions to controller main.go

jaypipes:
I'd left a TODO in the main.go template for controllers to fill in their
GitVersion, GitCommit and BuildDate information properly. Just now
getting around to this...

Issue: aws-controllers-k8s/community#1171

### DIFF
--- a/templates/cmd/controller/main.go.tpl
+++ b/templates/cmd/controller/main.go.tpl
@@ -5,6 +5,7 @@ package main
 import (
 	"os"
 
+	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
 	ackcfg "github.com/aws-controllers-k8s/runtime/pkg/config"
 	ackrt "github.com/aws-controllers-k8s/runtime/pkg/runtime"
 	ackrtutil "github.com/aws-controllers-k8s/runtime/pkg/util"
@@ -16,7 +17,6 @@ import (
 	ctrlrtmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
 	svcsdk "github.com/aws/aws-sdk-go/service/{{ .ServicePackageName }}"
 
-	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
 	svcresource "github.com/aws-controllers-k8s/{{ .ServicePackageName }}-controller/pkg/resource"
 	svctypes "github.com/aws-controllers-k8s/{{ .ServicePackageName }}-controller/apis/{{ .APIVersion }}"
 {{- /* Import the go types from service controllers whose resources are referenced in this service controller.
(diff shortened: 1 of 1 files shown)


templates/cmd/controller/main.go.tpl
```diff
@@ -102,7 +103,11 @@ func main() {
 	)
 	sc := ackrt.NewServiceController(
 		awsServiceAlias, awsServiceAPIGroup, awsServiceEndpointsID,
-		ackrt.VersionInfo{},	// TODO: populate version info
+		ackrt.VersionInfo{
+			version.GitCommit,
```
vijtrip2, line 107: I see these values as empty in the generated controllers right now. What am i missing?

RedbackThomson: Yeah the LDFLAGS for individual controllers are not being loaded correctly. [You can see](https://github.com/aws-controllers-k8s/s3-controller/blob/main/Makefile#L7-L12) the version is pinned at `v0.0.0` and it's modifying variables that are not at that location (it's trying to reference them in main). Should probably model these flags after the [LDFLAGS in code-generator](https://github.com/aws-controllers-k8s/code-generator/blob/main/Makefile#L9-L15). These one load the version from the dirty git tag (which is clean when built through CI/CD), and properly use import paths for the variable replacement.

RedbackThomson: The Makefile is not included in `code-generator`, but IMO it should be. There isn't anything in there that is service-specific anyway, so I don't see any problem with doing so.

vijtrip2: >The Makefile is not included in code-generator, but IMO it should be.

For the most part, yes! But a service team may modify their Makefile and we would not wanna override it from code-generator in that case. Solution for that problem can be custom hooks for adding new targets to auto-generated Makefile.

vijtrip2: Got it now. These variables are set at build-time.
So this PR looks good to be merged however we still need to fix Makefile to pass the correct ldflags.

vijtrip2:
/approve

@RedbackThomson feel free to merge if you have no concerns.

RedbackThomson:
> /approve
> 
> 
> 
> @RedbackThomson feel free to merge if you have no concerns.

I'd prefer not to have to modify every repositories Makefiles. I would rather we have them code generated tbh

vijtrip2:
> I'd prefer not to have to modify every repositories Makefiles. I would rather we have them code generated tbh

I think we won't need to. This PR will correctly use LDFlags from controller/pkg/version in main.go and the [Dockerfile](https://github.com/aws-controllers-k8s/code-generator/blob/main/Dockerfile#L41) correctly sets those LDFLAGS when controller binary is built.

The LDFlags in controller Makefile are not used anywhere because binary is built in DockerFile which does not use Makefile.

RedbackThomson:
Oh woah I did not notice that! That's confusing, but yay. I'm happy, then!
/lgtm

PR #329, opened 2022-04-26
Add a new generator config to append a `Synced` printer column

A-Hilaly:
Issue: aws-controllers-k8s/community#1281

Description:

Since all ACK kubernetes resources have a synced condition status,
we need a way to instruct the code generator to generate kubebuilder
marker comments that will append the SYNCED field the `kubectl get`
response.

This patch extends Resource.PrintConfig to allow configuring the
ack-generate to generate such marker comments.


RedbackThomson (approved):
Cool. Maybe the generator config is unnecessary, but I'll see what the others think.

jaypipes (approved):
Yeah, this is cool for now. I think maybe we should just do AGE and SYNCED always for all resources, though, and not need to be configurable as @RedbackThomson alluded to. But, for now, all good.

PR #332, opened 2022-05-05
Add support for references within lists of structs

RedbackThomson:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1256

Description of changes:
This pull request adds support for references that occur within lists of structs. It updates the generated `references.go` files in each controller to now iterate through the structs to check for the existence of references and again for resolving their individual references. Rather than creating a new slice of structs with resolved references, it instead resolves the references in place - updating the base field value and leaving the reference untouched. 

This pull request has been tested against the APIGatewayv2 and EC2 resource reference tests, as well as manually against the EC2 changes that inspired the original issue.

### DIFF
--- a/pkg/generate/ack/controller.go
+++ b/pkg/generate/ack/controller.go
@@ -19,6 +19,7 @@ import (
 	ttpl "text/template"
 
 	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/config"
+	"github.com/aws-controllers-k8s/code-generator/pkg/fieldpath"
 	"github.com/aws-controllers-k8s/code-generator/pkg/generate/code"
 	"github.com/aws-controllers-k8s/code-generator/pkg/generate/templateset"
 	"github.com/aws-controllers-k8s/code-generator/pkg/model"
@@ -59,6 +60,9 @@ var (
 		},
 		"ResourceExceptionCode": func(r *ackmodel.CRD, httpStatusCode int) string {
 			return r.ExceptionCode(httpStatusCode)
+		},
+		"ConstructFieldPath": func(targetFieldPath string) *fieldpath.Path {
+			return fieldpath.FromString(targetFieldPath)
 		},
 		"GoCodeSetExceptionMessageCheck": func(r *ackmodel.CRD, httpStatusCode int) string {
 			return code.CheckExceptionMessage(r.Config(), r, httpStatusCode)
--- a/pkg/generate/ack/controller.go
+++ b/pkg/generate/ack/controller.go
@@ -15,6 +15,7 @@ package ack
 
 import (
 	"path/filepath"
+	"sort"
 	"strings"
 	ttpl "text/template"
 
@@ -263,6 +264,7 @@ func Controller(
 	for serviceName := range referencedServiceNamesMap {
 		referencedServiceNames = append(referencedServiceNames, serviceName)
 	}
+	sort.Strings(referencedServiceNames)
 	cmdVars := &templateCmdVars{
 		metaVars,
 		snakeCasedCRDNames,
--- a/pkg/generate/code/resource_reference.go
+++ b/pkg/generate/code/resource_reference.go
@@ -67,7 +67,7 @@ func ReferenceFieldsValidation(
 
 				fieldConfig, ok := crd.Fields[fieldNamePrefix]
 				if !ok {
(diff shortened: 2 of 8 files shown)


pkg/generate/code/resource_reference.go
```diff
@@ -99,27 +125,47 @@ func ReferenceFieldsPresent(
...
+			// Determine whether the field is nested
+			if fp.Size() > 0 {
+				// Determine whether the field is inside a slice
+				parentField, ok := crd.Fields[fp.String()]
+				if !ok {
+					panic(fmt.Sprintf("CRD %s has no Field with path %s", crd.Kind, fp.String()))
+				}
+
+				if parentField.ShapeRef.Shape.Type == "list" {
+					iteratorsOut += fmt.Sprintf("for _, iter%d := range %s.%s {\n", fieldIndex, fieldAccessPrefix, parentField.Path)
```
vijtrip2, line 150: This should be preceded by a nil check or it will throw nil pointer dereference error inside range `r.ko.Spec.FieldA.FieldB.ListFieldC`

There is a generator method already written to generate nil check for nested fields.

templates/pkg/resource/references.go.tpl
```diff
@@ -123,16 +138,44 @@ func resolveReferenceFor{{ $field.FieldPathWithUnderscore }}(
 	}
 	return nil
 }
+{{ else }}
+{{ $parentField := index .CRD.Fields $fp.String }}
+{{ if eq $parentField.ShapeRef.Shape.Type "list" -}}
```
vijtrip2, line 143: nit: I do not think you need this `if eq $parentField.ShapeRef.Shape.Type "list"` check(L#143) and it's corresponding else block(L#164).

Because after first `if` block ( `{{ if not $isList -}}` L#117)only lists will be passed down to `else if` and `else`. `else if`(L#127) checks for `nested lists` , and only `non-nested lists` will reach else block(L#141).

I don't think you need another shape check(L#143) inside `else` block.

pkg/generate/ack/controller.go
```diff
@@ -259,6 +264,7 @@ func Controller(
 	for serviceName := range referencedServiceNamesMap {
 		referencedServiceNames = append(referencedServiceNames, serviceName)
 	}
+	sort.Strings(referencedServiceNames)
```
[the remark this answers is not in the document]
RedbackThomson: Yeah, I was seeing the imports and schemas jumping around in my git diff

templates/pkg/resource/references.go.tpl
RedbackThomson: L#143 is not the same is `$isList` though. It checks that the parent of the field is a list. The first conditional is checking for single primitives, the second condition (the else if) is checking for lists of primitives. Within the third condition (the else) we are checking for lists of structs (the list is the parent field, not the field itself) and falling back to lists of primitives.

These are (some of) the conditions for resolving references - and which ones we support:
- [x] Single primitives (Condition 1)
- [x] List of primitives (Condition 2)
- [x] Singly-nested primitives (Condition 1)
- [x] Singly-nested struct (Condition 1)
- [x] Singly-nested list of primitives (Condition 2)
- [x] Singly-nested list of structs (Condition 3)
- [x] Recursively-nested primitives (Condition 1)
- [ ] Recursively-nested list of primitives
- [ ] Recursively-nested list of structs

Don't quote me on that. Perhaps some of these solutions can be generalised to match those, but it's also possible there are other cases I am not considering. Either way, I think this code will eventually need to be moved out of `references.go.tpl` and made with a string builder in Go. It's a bit too complex (especially if we use layers of iterators) for Go templates (and a PITA to debug).

RedbackThomson: BTW I'd love to draw a [Karnaugh map](https://en.wikipedia.org/wiki/Karnaugh_map) of the conditions and figure out a minimised conditional for this

pkg/generate/code/resource_reference.go
RedbackThomson: Yep. Not necessary for this particularly example I'm testing against, since `range ko.Spec.Routes` will just iterate 0 times in `Routes` is `nil`. But easy enough to add

jaypipes (approved):
Still :+1:  from me. Thanks @RedbackThomson!

PR #325, opened 2022-04-15
Refactor model package to use `Config` accessors

brycahta:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1184 (task 2; migrating access)

Description of changes:
* copy methods accessing `ackgenconfig` from model package to config package
* update methods in model package to leverage `ackgenconfig` accessors instead of directly accessing config
* align `ackgenconfig` accessors with config types (i.e. methods accessing `c.Resources` go in `pkg/config/resource.go`)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #335, opened 2022-05-19
update for default resource tags

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1261

Description of changes:
* Removes `service.k8s.aws/managed=true` and `services.k8s.aws/created=%UTC_NOW%` tags and adds the new `services.k8s.aws/controller-version=%CONTROLLER_VERSION%` tag as resource-tags input for ACK controller.
* Updates ACK runtime to `v0.19.0`
* Adds new `TagConfig` field inside `ResourceConfig`
* Adds the code-generation for implementation of `AWSResourceManager.EnsureTags()`


vijtrip2:
Note to reviewer: tests expected to fail due to `VersionInfo` refactoring from "pkg/runtime" to "pkg/types" inside ack runtime. PR -> https://github.com/aws-controllers-k8s/runtime/pull/87

Locally tested that this code-generator change works as expected.

RedbackThomson:
👍🏼 Will wait for the runtime PR to be merged

A-Hilaly (approved):
LGTM

A-Hilaly:
>Merging this as-is. It looks like we need to update the tests to modify the go.mod files for dynamodb/ecr/s3 controllers to the being-updated-to ACK runtime version?

@jaypipes We still don't have the part that generates `EnsureTags` methods for resources. This should wait for now

vijtrip2:
Note to reviewer: DynamoDB, S3, ECR tests are expected to fail because they need an update inside their `generator.yaml` to ignore tagging, or provide path for the tagging field. I am planning to update the generator.yaml for these controllers when their auto generation failure issue gets created by ack-bot.

Failures in this PR are because the default top level `Tags` field does not exist inside Backup(Dynamo), Bucket(S3) and PullThroughCache(ECR) resources.

pkg/model/crd.go
```diff
@@ -718,6 +718,119 @@ func (r *CRD) SortedFieldNames() []string {
 	return fieldNames
 }
 
+// GetIgnoreTagging returns whether ensuring controller tags should be ignored
+// for a resource or not.
+func (r *CRD) GetIgnoreTagging() bool {
```
brycahta, line 723: After this refactor https://github.com/aws-controllers-k8s/code-generator/pull/325, I was thinking we move any config-accessing logic outside of model and into `config`. Would you be opposed to following that style or is there something I'm missing why it needs to be implemented here?

pkg/model/crd.go
```diff
@@ -718,6 +718,119 @@ func (r *CRD) SortedFieldNames() []string {
...
+				}
+			}
+		}
+	}
+	return valueMemberName
+}
+
+// isListOfStruct method returns true is the shape of field is list of struct,
+// false otherwise
+func isListOfStruct(f *Field) bool {
```
brycahta, line 821: Can we make this a `Field` method instead? Ex: `func (f *Field) IsListOfStruct() bool {}`

vijtrip2: This was a very specific method for handling the tags and i did not wish to add it to Field struct.
Field struct has generic method to retrieve the field shape, Hence i left the private `isListOfStruct` close to other tag specific method for coherency.

pkg/testdata/models/apis/s3/0000-00-00/generator-with-tags.yaml
```diff
@@ -0,0 +1,38 @@
...
+    list_operation:
+      match_fields:
+        - Name
+    fields:
+      Tagging:
+        from:
+          operation: PutBucketTagging
+          path: Tagging
+#      Tagging.TagSet:
+#        has_tags: true
```
RedbackThomson, line 29: Old implementation detail? To be deleted?

templates/pkg/resource/manager.go.tpl
```diff
@@ -252,6 +258,53 @@ func (rm *resourceManager) IsSynced(ctx context.Context, res acktypes.AWSResourc
...
+func (rm *resourceManager) EnsureTags(
+    ctx context.Context,
+    res acktypes.AWSResource,
+    md acktypes.ServiceControllerMetadata,
+) error {
+{{- if $hookCode := Hook .CRD "ensure_tags" }}
+{{ $hookCode }}
+{{ else }}
+{{ $tagField := .CRD.GetTagField -}}
+{{ if $tagField -}}
```
RedbackThomson, line 277: `GetTagField` should panic if one can't be found, right? We don't want to generate an empty method here, otherwise? Therefore, we should be able to remove this condition

RedbackThomson (changes requested):

jaypipes (changes requested):
great start, @vijtrip2! few notes inline but overall looking good.

templates/pkg/resource/manager.go.tpl
vijtrip2: We wanna return a method with just `return nil` for the resources that do not support tagging. For example: apigatewayv2's route resource.

GetTagField will return `nil` when generator.yaml contains `tags: ignore: true` and then `EnsureTags` will only contain `return nil` statement.

pkg/generate/code/initialize_field.go
```diff
@@ -0,0 +1,98 @@
...
+//  ```
+
+func InitializeNestedStructField(
+	r *model.CRD,
+	sourceVarName string,
+	field *model.Field,
+	// apiPkgImportName contains the imported package name where the type definition
+	// for nested structs is present.
+	// ex: svcapitypes "github.com/aws-controllers-k8s/s3-controller/apis/v1alpha1"
+	apiPkgImportName string,
```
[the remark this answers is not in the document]
vijtrip2: yes, I will refactor variable name to 'apiPkgAlias'.

pkg/generate/code/initialize_field.go
```diff
@@ -0,0 +1,98 @@
...
+						" Shape type for %s is %s inside fieldpath %s", elemName,
+						elemShapeRef.Shape.Type, fieldPath))
+				}
+				out += fmt.Sprintf("%s%s.%s = &%s%s{}\n",
+					indent, elemAccessPrefix, elemName, importPath,
+					elemShapeRef.GoTypeElem())
+				elemAccessPrefix = fmt.Sprintf("%s.%s", elemAccessPrefix,
+					elemName)
+				index++
+			}
```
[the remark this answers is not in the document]
vijtrip2: here `sourceVar` parameter already contains the `.Spec` part.

`{{ $sourceVar := (print "r.ko" .CRD.Config.PrefixConfig.SpecField) -}}` inside manager.go.tpl sets up this parameter to be passed to both `CheckNilFieldPath` and `GoCodeInitializeNestedStructField`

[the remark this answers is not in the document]
vijtrip2: Now, i have added the part to determine Spec/Status from field name and add it accordingly.

pkg/generate/code/initialize_field.go
```diff
@@ -0,0 +1,107 @@
...
+	apiPkgAlias string,
+	// Number of levels of indentation to use
+	indentLevel int,
+) string {
+	out := ""
+	indent := strings.Repeat("\t", indentLevel)
+	fieldPath := field.Path
+	if fieldPath != "" {
+		fp := fieldpath.FromString(fieldPath)
+		if fp.Size() > 1 {
```
RedbackThomson, line 56: nit: Early exit on inverse (`if fp.Size() == 0`)

templates/pkg/resource/tags.go.tpl
```diff
@@ -0,0 +1,81 @@
...
+
+    svcapitypes "github.com/aws-controllers-k8s/{{ .ServicePackageName }}-controller/apis/{{ .APIVersion }}"
+)
+
+var (
+	_ = svcapitypes.{{ .CRD.Kind }}{}
+	_ = acktags.NewTags()
+)
+
+{{- if $hookCode := Hook .CRD "convert_tags" }}
```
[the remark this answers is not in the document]
RedbackThomson: Is that after `gofmt` has removed them all?

RedbackThomson (approved):
Just had 2 more nits. But I think the functionality and logic is all solid.

pkg/generate/code/initialize_field.go
vijtrip2: not needed, this is the only condition. if `fp.Size() == 0`, it never enters if block and jumps to return statement.

templates/pkg/resource/tags.go.tpl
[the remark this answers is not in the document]
vijtrip2: yeah you are right. fixed it now.

pkg/generate/code/initialize_field.go
```diff
@@ -0,0 +1,107 @@
...
+			// Use the front of fieldpath to determine whether the field is
+			// a spec field or status field.
+			elemAccessPrefix := sourceVarName
+			if _, found := r.SpecFields[front]; found {
+				elemAccessPrefix = fmt.Sprintf("%s%s", elemAccessPrefix,
+					r.Config().PrefixConfig.SpecField)
+			} else {
+				elemAccessPrefix = fmt.Sprintf("%s%s", elemAccessPrefix,
+					r.Config().PrefixConfig.StatusField)
+			}
```
jaypipes, line 82: :+1:

jaypipes (approved):
Rock on brother Vijay :) Really smooth work. ++

PR #339, opened 2022-06-09
Add helper function so go template can create slice

brucegucode:
Issue #, if available: no issue

Description of changes:
Create a small helper function so go template can create slice.
It will be used by rds controller soon. 

### DIFF
--- a/pkg/generate/ack/controller.go
+++ b/pkg/generate/ack/controller.go
@@ -169,9 +169,9 @@ var (
 		"CheckNilReferencesPath": func(f *ackmodel.Field, sourceVarName string) string {
 			return code.CheckNilReferencesPath(f, sourceVarName)
 		},
-    "Each": func (args ...interface{}) []interface{} {
-      return args
-    },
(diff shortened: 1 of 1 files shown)


pkg/generate/ack/controller.go
```diff
@@ -169,6 +169,9 @@ var (
 		"CheckNilReferencesPath": func(f *ackmodel.Field, sourceVarName string) string {
 			return code.CheckNilReferencesPath(f, sourceVarName)
 		},
+                "MakeSlice": func (args ...interface{}) []interface{} {
+                        return args
+                },
```
jaypipes, line 174: @brucegucode can you offer an example of how this type of generic function would be used in a template?

jaypipes: @brucegucode OK, I understand the use case better now. How about calling this function `Each`, so that the calling signature would look like this?

```handlebars
{{ range $operationName := Each "RestoreDBInstanceFromDBSnapshot" "CreateDBInstanceReadReplica" }}
```

PR #344, opened 2022-06-16
Use latest from sdk create/update for rm.onError()

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1336

Description of changes:
* After the sdkCreate/sdkUpdate call, the resource manager was still using the latest from ReadOne call when invoking `rm.onError` method
* Instead it should use the resource returned from sdkCreate and sdkUpdate call


jaypipes (approved):
As mentioned in the Chime call, nice detective work on this @vijtrip2 :)

PR #346, opened 2022-06-21
Add cycle detection with panic

RedbackThomson:
Fixes https://github.com/aws-controllers-k8s/community/issues/1327

Description of changes:
The majority of our code-generator templates and methods rely on our custom resource properties being represented as DAGs. That is, they can recurse through their children until they reach primitive values at the base. Without a major refactor, it is not possible to support a cyclic field reference. This pull request adds logic to detect a cycle as we are iterating over new fields and will panic to tell the user about the limitation.


pkg/model/field.go
```diff
@@ -269,16 +286,36 @@ func NewField(
...
+
+				// Check to see if we have seen this shape before in the stack
+				// and panic. Cyclic references are not supported
+				if _, ok := nestedParentFields[memberShape.ShapeName]; ok {
+					panic(fmt.Sprintf("Detected a cyclic type reference in %s"+
+						". Add the corresponding path to `ignore.field_paths`"+
+						" in the generator config to continue.",
+						containerShape.ShapeName))
+				}
+
```
vijtrip2, line 313:
```suggestion
				if path, ok := nestedParentFields[memberShape.ShapeName]; ok {
					panic(fmt.Sprintf("Detected a cyclic type reference in %s"+
						".  %s refers to same shape as %s. Add %s to `ignore.field_paths`"+
						" in the generator config to continue.",
						containerShape.ShapeName,
						memberPath,
						path,
						memberPath))
				}

```

vijtrip2:
Nice improvement!

vijtrip2:
Nice improvement!

pkg/model/field.go
```diff
@@ -229,6 +229,23 @@ func NewField(
...
+// newFieldRecurse recursively calls itself with protection against infinite
+// cycles
+func newFieldRecurse(
+	crd *CRD,
+	path string,
+	// A map of shape name to Fields. Keeps track of every shape that has been
+	// recursively discovered so that we can detect cycles. For example, the
+	// emrcontainers `Configuration` object contains a property of type
+	// `Configuration`.
+	parentFields map[string]*struct{},
```
A-Hilaly, line 245: Very nit: Usually sets are implemented using `map[string]bool` in Go

PR #349, opened 2022-06-24
Ensure failed secret dereferencing is recoverable

RedbackThomson:
Fixes https://github.com/aws-controllers-k8s/community/issues/1318

Description of changes:
If an ACK resource is evaluated before a secret becomes available, the controller should attempt to retry (with exponential backoff) until the secret exists. This pull request ensures that any error returned by `SecretValueFromReference` is retried by the controller.

Note: I chose not to update the error values in `SecretValueFromReference` because I thought it was more appropriate that the SDK logic chose the requeue action, not the secret fetching method.


jaypipes (approved):
++ thank you @RedbackThomson

PR #354, opened 2022-07-14
allow empty string as resourceTag entry

vijtrip2:
Description of changes:
* allow empty string inside resourceTags entry so that users can override the default ACK tag behavior using `--set resourceTags={}` param during helm installation
* tested locally that this change allows removing the default ACK tags behavior during helm installation


RedbackThomson:
Yes please
/lgtm

PR #357, opened 2022-07-27
Remove faulty suffix trim from Helm deployment

RedbackThomson:
Description of changes:
Removes a suffix trim character from the Helm deployment template. This trim was causing the string to concatenate with the next line and create an error. Removing this ensures it stays as a separate line if the `priorityClassName` is specified.


vijtrip2 (commented):
/lgtm

PR #359, opened 2022-08-19
Helm install check

kskzk:
Issue #, if available:  https://github.com/aws-controllers-k8s/community/issues/1436

Description of changes:
- Add helm version check function
- Add helm installation helper script
- Modify k8s_controller_gen_version_equals function and its comment

### DIFF
--- a/scripts/lib/common.sh
+++ b/scripts/lib/common.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 CONTROLLER_TOOLS_VERSION="v0.7.0"
+HELM_VERSION="v3.9"
 
 # setting the -x option if debugging is true
 if [[ "${DEBUG:-"false"}" = "true" ]]; then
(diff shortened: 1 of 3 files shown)


A-Hilaly (approved):
Neat :+1:  
Thank you @kskzk !

scripts/lib/common.sh
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 CONTROLLER_TOOLS_VERSION="v0.7.0"
+HELM_VERSION="v3.9"
```
A-Hilaly, line 4: Let's drop the version to `3.7` ? i guess 3.9 is not required.

A-Hilaly: Or maybe we need to update our prow jobs to use 3.9 helm binaries..

A-Hilaly: Helm is installed here https://github.com/aws-controllers-k8s/test-infra/blob/da4dad6c57818a0d4f87f8e2ac9e608612bfaff8/prow/jobs/images/Dockerfile.auto-generate-controllers and few other dockerfiles in the same directory

kskzk: Thanks for the review!
Which do you think is better? Do we drop the minimum required version to 3.7 for now? If so, I will modify the pull request.

A-Hilaly: I'm thinking that maybe we should start by dropping the version to 3.7 and dealing with the upgrade in a different PR

A-Hilaly:
Neat! Thank you @kskzk!
/approve
/lgtm

PR #369, opened 2022-10-21
Correct the generated code for nested parent field reference.

Vandita2020:
While generating the code for references for the Lambda function, more specifically for the S3 bucket field, we observe that there is use of array keywords with a non-array field e.g len(S3.Bucket) which is incorrect.

After debugging code-generator, we observe that there was an incorrect piece of code (most likely because of copy paste) that considered nested parent field as an array.

This patch fixes this issue by adding the right code for handling nested parent fields.

Unfortunately we cannot add unit tests for this change because this part of the code generation resides inside the template and it's not a part of the pkg/generate/code package.

Fixes https://github.com/aws-controllers-k8s/community/issues/1514

### DIFF
--- a/templates/pkg/resource/references.go.tpl
+++ b/templates/pkg/resource/references.go.tpl
@@ -143,6 +143,16 @@ func resolveReferenceFor{{ $field.FieldPathWithUnderscore }}(
 	}
 	return nil
 }
+{{ else if not $isList -}}
+    if ko.Spec.{{ $field.ReferenceFieldPath }} != nil &&
+        ko.Spec.{{ $field.ReferenceFieldPath }}.From != nil  {
+        arr := ko.Spec.{{ $field.ReferenceFieldPath }}.From
+{{ template "read_referenced_resource_and_validate" $field }}
+            referencedValue := string(*obj.{{ $field.FieldConfig.References.Path }})
+            ko.Spec.{{ $field.Path }} = &referencedValue
+    }
+    return nil
+}
 {{ else }}
 {{ $parentField := index .CRD.Fields $fp.String }}
 {{ if eq $parentField.ShapeRef.Shape.Type "list" -}}


templates/pkg/resource/references.go.tpl
```diff
@@ -117,7 +117,7 @@ func resolveReferenceFor{{ $field.FieldPathWithUnderscore }}(
 
 {{- $fp := ConstructFieldPath $field.Path -}}
 {{ $_ := $fp.Pop -}}
-{{ $isNested := gt $fp.Size 0 -}}
+{{ $isNested := gt $fp.Size 0 -}}
```
jaypipes, line 120: Mind removing the whitespace at the end of this line please, @Vandita2020? :)

templates/pkg/resource/references.go.tpl
```diff
@@ -143,6 +143,17 @@ func resolveReferenceFor{{ $field.FieldPathWithUnderscore }}(
 	}
 	return nil
 }
+{{ else if not $isList -}}
+//is nested
```
jaypipes, line 147: Please use a Go template comment instead of a Go comment:

```
{{/* nested field */}}
```

templates/pkg/resource/references.go.tpl
```diff
@@ -143,6 +143,17 @@ func resolveReferenceFor{{ $field.FieldPathWithUnderscore }}(
 	}
 	return nil
 }
+{{ else if not $isList -}}
```
jaypipes, line 146: Unless I'm mistaken, `isNested` could be `false` here.... so better to do this?

```
{{ if $isNested and (not $isList) -}}
```

jaypipes (changes requested):

templates/pkg/resource/references.go.tpl
Vandita2020: It's not necessary, it is always going to be true here because of L132.

templates/pkg/resource/references.go.tpl
Vandita2020: Done

templates/pkg/resource/references.go.tpl
Vandita2020: Done.

A-Hilaly (approved):
+++

templates/pkg/resource/references.go.tpl
jaypipes: Indeed, I missed that! :)

jaypipes (approved):

A-Hilaly:
Merging as the test failure are related to gopkg.in issues (see https://github.com/niemeyer/gopkg/issues/63)

PR #370, opened 2022-11-02
Helm with irsa

jeffeland:
Issue #, if available:

Description of changes:
Prepared the templates to generate the helm chart with the IRSA part when the iam controller is deployed.
That way the other controllers can be deployed via the helm chart only and not need extra steps to setup the IRSA.

### DIFF
--- a/templates/helm/templates/irsa.yaml.tpl
+++ b/templates/helm/templates/irsa.yaml.tpl
@@ -1,0 +1,30 @@
+{{- if and .Values.aws.accountId .Values.aws.oidcProvider }}
+apiVersion: iam.services.k8s.aws/v1alpha1
+kind: Role
+metadata:
+  name: {{ .Values.serviceAccount.name }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  description: IRSA role for ACK {{ .ServicePackageName }} controller deployment on EKS cluster using Helm charts
+  name: {{ .Values.serviceAccount.name }}
+  policies:
+    - "arn:aws:iam::aws:policy/AmazonEC2FullAccess"
+  assumeRolePolicyDocument: >
+    {
+      "Version": "2012-10-17",
+      "Statement": [
+        {
+          "Effect": "Allow",
+          "Principal": {
+            "Federated": "arn:aws:iam::{{ .Values.aws.accountId }}:oidc-provider/{{ .Values.aws.oidcProvider }}"
+          },
+          "Action": "sts:AssumeRoleWithWebIdentity",
+          "Condition": {
+            "StringEquals": {
+              "{{ .Values.aws.oidcProvider }}:sub": "system:serviceaccount:{{ .Release.Namespace }}:{{ .Values.serviceAccount.name }}"
+            }
+          }
+        }
+      ]
+    }
+{{- end }}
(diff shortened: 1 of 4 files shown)


templates/helm/templates/irsa.yaml.tpl
```diff
@@ -0,0 +1,30 @@
...
+apiVersion: iam.services.k8s.aws/v1alpha1
+kind: Role
+metadata:
+  name: {{ .Values.serviceAccount.name }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  description: IRSA role for ACK {{ .ServicePackageName }} controller deployment on EKS cluster using Helm charts
+  name: {{ .Values.serviceAccount.name }}
+  policies:
+    - "arn:aws:iam::aws:policy/AmazonEC2FullAccess"
```
jeffeland, line 11: This line need to be templated

templates/helm/values.yaml.tpl
```diff
@@ -53,6 +53,8 @@ resources:
 aws:
   # If specified, use the AWS region for AWS API calls
   region: ""
+  oidcProvider: ""
+  endpoint_url: ""
```
RedbackThomson, line 57: I believe this should be `accountID`?

templates/helm/values.schema.json
```diff
@@ -159,6 +159,12 @@
         "region": {
           "type": "string"
         },
+        "accountId": {
```
RedbackThomson, line 162: Could we please rename this to `accountID`? (Proper casing on initialisms)

templates/helm/templates/service-account.yaml
```diff
@@ -15,4 +15,7 @@ metadata:
   {{- range $key, $value := .Values.serviceAccount.annotations }}
     {{ $key }}: {{ $value | quote }}
   {{- end }}
+  {{- if and .Values.aws.accountId .Values.aws.oidcProvider }}
+    eks.amazonaws.com/role-arn: arn:aws:iam::{{ .Values.aws.accountId }}:role/ack-ec2-controller
```
RedbackThomson, line 19: Please change `ack-ec2-controller` to be the name of the role created before

templates/helm/templates/irsa.yaml.tpl
```diff
@@ -0,0 +1,30 @@
+{{- if and .Values.aws.accountId .Values.aws.oidcProvider }}
+apiVersion: iam.services.k8s.aws/v1alpha1
+kind: Role
+metadata:
+  name: {{ .Values.serviceAccount.name }}
```
RedbackThomson, line 5: We might want to make the role name a separate property from the service account, otherwise installing ACK controllers in 2 clusters within the same account would have conflicting roles.

RedbackThomson (changes requested):
Hey Jean-Francois,
I understand where you are coming from with this change but I am worried about the user experience. A naive user might not understand that they need to initially install and manually configure the IAM controller to use this functionality. I worry that we will receive complaints from people saying that the controller is broken even though they are providing a service account and valid annotation without seeing that the IAM role does not yet exist. I know that documentation would call this out, but I fear that most people are going to look at the Helm charts to be self-documenting and in this case there is no indication of this limitation. What are your thoughts about that?

RedbackThomson:
I still feel that this feature should not be enabled for all controllers - since these Helm charts assume that there is only one controller being deployed. We cannot assume that the IAM controller already exists in the cluster.

We are working on releasing a multi-controller Helm chart, which would include the IAM controller by default. I think this would be an excellent use-case for that chart, since it would automate the IRSA configuration for all of the other controllers. See the issue discussing the multi-controller Helm chart here - https://github.com/aws-controllers-k8s/community/issues/1360

PR #371, opened 2022-11-03
Normalize field names specific to comparison in generator config

A-Hilaly:
While generating code for iam-controller OIDC resource, we found a tiny bug causing the code-generator to generate comparison code for a field that was marked as `compare.is_ignored`.
This was found to be the result of the code-generator not properly normalizing field names.

This patch fixes this bug by normalizing the field name using camel case.
This not a good long term fix. Ideally normalization should happen earlier.

Co-authored-by: John Jacobs <jljaco@amazon.com>


jljaco (commented):
lgtm

PR #373, opened 2022-11-10
using go env GOARCH to detect system arch instead of uname

acornett21:
Issue #, if available:
N/A

Description of changes:
`uname` does not always match the friendly arch that operator-sdk, an example is that on a normal mac you get `x86_64` which fails to download since operator-sdk is released as `amd64`, there could be other examples as well.

Other things I thought of, since we might not want to rely on `go env` and be tied to go-tools:
- `uname -m | tr '[:upper:]' '[:lower:]' | sed 's/x86_64/amd64/g'`
- have this script take in arch as an arg

Let me know which approach would be best


A-Hilaly (approved):
LGTM, thanks Adam! the go env should be good.
/ok-to-test
/approve

PR #375, opened 2022-11-16
always ensure operation overrides work

jaypipes:
There were two bugs in the code generator's construction of the OperationMap during API inference.

The first bug was a non-deterministic behaviour that showed itself when an API had multiple Operations of a single operation type (say, READ_ONE). The code that was responsible for determining which Operation mapped to which operation type for a particular resource was looping over the `aws-sdk-go` `private/model/api:API`'s `Operations` field, which is a map. Depending on the parsing of the underlying api-2.json files, same-typed Operations with different IDs -- e.g. `DescribeStreamSummary` and `DescribeStream` -- were being mapped to a single operation type -- OpTypeGet. Depending on which Operation ID came last, that is the operation that was being mapped, regardless of which one appeared in the OperationConfig configuration option that handles overrides.

The second bug was that the `OpTypeFromString` method was too strict in its comparison logic. The strings "get", "READ_ONE", "read_one" and "Get" should all lead to OpTypeGet, but only "Get" was being matched.

Fixes Issue aws-controllers-k8s/community#1555

### DIFF
--- a/pkg/model/op.go
+++ b/pkg/model/op.go
@@ -150,32 +150,32 @@ func GetOpTypeAndResourceNameFromOpID(
 }
 
 func OpTypeFromString(s string) OpType {
-	switch s {
-	case "Create":
+	switch strings.ToLower(s) {
+	case "create":
 		return OpTypeCreate
-	case "CreateBatch":
+	case "createbatch":
 		return OpTypeCreateBatch
-	case "Delete":
+	case "delete":
 		return OpTypeDelete
-	case "Replace":
+	case "replace":
 		return OpTypeReplace
-	case "Update":
+	case "update":
 		return OpTypeUpdate
-	case "AddChild":
+	case "addchild":
 		return OpTypeAddChild
-	case "AddChildren":
+	case "addchildren":
 		return OpTypeAddChildren
-	case "RemoveChild":
+	case "removechild":
 		return OpTypeRemoveChild
-	case "RemoveChildren":
+	case "removechildren":
 		return OpTypeRemoveChildren
-	case "Get":
+	case "get", "readone", "read_one":
 		return OpTypeGet
-	case "List":
+	case "list", "readmany", "read_many":
 		return OpTypeList
-	case "GetAttributes":
+	case "getattributes":
 		return OpTypeGetAttributes
-	case "SetAttributes":
+	case "setattributes":
 		return OpTypeSetAttributes
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/model/sdk_api.go
```diff
@@ -86,6 +86,24 @@ func (a *SDKAPI) GetOperationMap(cfg *ackgenconfig.Config) *OperationMap {
...
+	// APIs have multiple operations of a particular OpType and we need to
+	// always be sure that the overridden operation is the one that we use
+	// during inference.
+	for opID, opCfg := range cfg.Operations {
+		if opCfg.ResourceName == "" {
+			continue
+		}
+		op, found := a.API.Operations[opID]
+		if !found {
+			panic("operation " + opID + " in generator.yaml 'operations:' object does not exist.")
```
jljaco, line 100: Nit: have this error read `... does not exist in the upstream API.`

pkg/model/sdk_api.go
```diff
@@ -86,6 +86,24 @@ func (a *SDKAPI) GetOperationMap(cfg *ackgenconfig.Config) *OperationMap {
 			opMap[opType][resName] = op
 		}
 	}
+
+	// We need to do a second pass over the operation overrides because some
+	// APIs have multiple operations of a particular OpType and we need to
+	// always be sure that the overridden operation is the one that we use
+	// during inference.
```
jljaco, line 93: Consider citing here a motivating example or simply include a reference to https://github.com/aws-controllers-k8s/community/issues/1555

jljaco (commented):
Just a few minor nits (a comment, an error message, and a typo)

Overall looks great, this is a great example of a well-documented bug and accompanying fix.  Thank you for including a clear and thorough PR description!

pkg/model/op.go
```diff
@@ -150,32 +150,32 @@ func GetOpTypeAndResourceNameFromOpID(
...
+	case "replace":
 		return OpTypeReplace
-	case "Update":
+	case "update":
 		return OpTypeUpdate
-	case "AddChild":
+	case "addchild":
 		return OpTypeAddChild
-	case "AddChildren":
+	case "cddchildren":
```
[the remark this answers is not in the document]
jaypipes: fixed.

pkg/model/sdk_api.go
jaypipes: done! :)

A-Hilaly:
Github is broken again..
/retest

A-Hilaly (approved):
Nice catch 👍

PR #383, opened 2022-12-13
ensure attribute fields have a ShapeRef

jaypipes:
When attempting to generate the SNS controller, I was getting the following failure:

```
[jaypipes@thelio code-generator]$ make build-controller SERVICE=sns
building ack-generate ... ok.
==== building sns-controller ====
Copying common custom resource definitions into sns
Building Kubernetes API objects for sns
Generating deepcopy code for sns
Generating custom resource definitions for sns
Building service controller for sns
Error: template: /home/jaypipes/go/src/github.com/aws-controllers-k8s/code-generator/templates/pkg/resource/delta.go.tpl:34:3: executing "/home/jaypipes/go/src/github.com/aws-controllers-k8s/code-generator/templates/pkg/resource/delta.go.tpl" at <GoCodeCompare .CRD "delta" "a.ko" "b.ko" 1>: error calling GoCodeCompare: runtime error: invalid memory address or nil pointer dereference
make: *** [Makefile:41: build-controller] Error 1
```

This failure turned out to be due to attribute-based API fields that do not have a `type: XXX` generator.yaml configuration option specified (as they do in the sqs-controller's generator.yaml) would end up getting a `pkg/model.Field` object created that had a `nil` ShapeRef. This caused problems in certain Go code generator functions (such as `pkg/generate/code.CompareResource`) that rely on `Field.ShapeRef` being non-nil.

Signed-off-by: Jay Pipes <jaypipes@gmail.com>

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #384, opened 2022-12-18
Fix panic in code generator when generating lambda controller

A-Hilaly:
The recent patch ([PR 383](https://github.com/aws-controllers-k8s/code-generator/pull/383)) introduced a regression that caused the
code generator to panic when generating the lambda controller.
This commit addresses the issue by adding a check to ensure that
the shapeRef reference is not nil before accessing the
Documentation field. This prevents the panic from occurring and
allows the code generator to function correctly.

By submitting this pull request, I confirm that my contribution is
made under the terms of the Apache 2.0 license.

### DIFF
--- a/templates/apis/crd.go.tpl
+++ b/templates/apis/crd.go.tpl
@@ -16,7 +16,7 @@ import (
 {{ .CRD.Documentation }}
 type {{ .CRD.Kind }}Spec struct {
 {{ range $fieldName, $field := .CRD.SpecFields }}
-{{ if $field.ShapeRef.Documentation -}}
+{{ if $field.ShapeRef -}}
     {{ $field.ShapeRef.Documentation }}
 {{ end -}}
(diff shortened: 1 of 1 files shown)


templates/apis/crd.go.tpl
```diff
@@ -16,7 +16,7 @@ import (
 {{ .CRD.Documentation }}
 type {{ .CRD.Kind }}Spec struct {
 {{ range $fieldName, $field := .CRD.SpecFields }}
-{{ if $field.ShapeRef.Documentation -}}
+{{ if $field.ShapeRef -}}
```
jljaco, line 19: Can you nil-check both `ShapeRef` and `ShapeRef.Documentation` here?

A-Hilaly: `ShapeRef.Documentation` is of type `string` (not `*string`), hence it will never cause nil pointer panics.

PR #387, opened 2023-01-04
Allow calling `GoCodeSetResourceForStruct` for non-CRD SDK shapes

A-Hilaly:
Issue: https://github.com/aws-controllers-k8s/community/issues/1609

Previously, the code-generator was unable to let us use `GoCodeSetResourceForStruct` for fields that are not part of a CRD top level fields. This caused issues when trying to generate custom code that transforms an eventbridge `v1alpha1.Target` to an `svcsdk.Target`.

This patch fixes the issue by allowing users to call `GoCodeSetResourceForStruct` for any field, not just CRD top level fields. In addition, unit tests have been added to ensure the correct behavior.

This patch is needed to allow us to properly execute the template written here https://github.com/embano1/eventbridge-controller/blob/aminemicha/templates/hooks/rule/sdk_file_end.go.tpl


RedbackThomson:
I don't think I understand how this solves the problem? This seems to be throwing away error handling, but I don't see how it's able to access fields that aren't in the spec or status? It still uses `f.Fields`?

A-Hilaly:
>I don't think I understand how this solves the problem? This seems to be throwing away error handling, but I don't see how it's able to access fields that aren't in the spec or status? It still uses f.Fields?

Because the statement bellow will not return a positive `ok` unless if one `targetFieldName` is one of the spec/status fields.
```go
        f, ok := r.Fields[targetFieldName]
	if !ok {
		return ""
	}
```

Reading the code again i just realized that i need to add the `setCfg.ignore` part.. adding that in a different commit

pkg/generate/code/set_resource_test.go
```diff
@@ -3273,6 +3273,336 @@ func TestSetResource_EC2_DHCPOptions_NestedSetConfig(t *testing.T) {
...
+	require.True(ok)
+	sdkTargetsShape, ok := putTargetsOP.InputRef.Shape.MemberRefs["Targets"]
+	require.True(ok)
+	sdkTargetShape := &sdkTargetsShape.Shape.MemberRef
+
+	expected := `	if sdkTarget.Arn != nil {
+		krTarget.ARN = sdkTarget.Arn
+	}
+	if sdkTarget.BatchParameters != nil {
+		krTargetf1 := &svcapitypes.BatchParameters{}
```
jaypipes, line 3301: Normally, this would just be:

```go
f1 := &svcapitypes.BatchParameters{}
```

not

```go
krTargetf1 := &svcapitypes.BatchParameters{}
```

I think you are incorrectly specifying the target variable name...

pkg/generate/code/set_resource_test.go
```diff
@@ -3273,6 +3273,336 @@ func TestSetResource_EC2_DHCPOptions_NestedSetConfig(t *testing.T) {
...
+func TestSetResource_EventBridge_Rule_SetResourceForStruct(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+
+	g := testutil.NewModelForService(t, "eventbridge")
+
+	crd := testutil.GetCRDByName(t, g, "Rule")
+	require.NotNil(crd)
+
+	// Getting KR api Target shape
```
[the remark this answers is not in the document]
A-Hilaly: Kubernetes Resource, changing to CRD.

pkg/generate/code/set_resource_test.go
```diff
@@ -3273,6 +3273,336 @@ func TestSetResource_EC2_DHCPOptions_NestedSetConfig(t *testing.T) {
...
+		if sdkTarget.SqsParameters.MessageGroupId != nil {
+			krTargetf15.MessageGroupID = sdkTarget.SqsParameters.MessageGroupId
+		}
+		krTarget.SQSParameters = krTargetf15
+	}
+`
+	assert.Equal(
+		expected,
+		code.SetResourceForStruct(
+			crd.Config(), crd, "", "krTarget", krTargetShape, nil, "sdkTarget", sdkTargetShape, "", model.OpTypeList, 1,
```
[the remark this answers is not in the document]
A-Hilaly: The thing is that Target is not a CRD nor a CRD Field - There is no need to use `resp` and `ko` as they really represent what the function is used for

pkg/generate/code/set_resource_test.go
A-Hilaly: The thing is if i change `krTarget` to an empty string, L3298 will become `.ARN = resp.Arn` which is not correct

embano1:
Pinging @A-Hilaly for who can review and LGTM since this is blocking the ACK EventBridge controller 🙏🏻

RedbackThomson:
Setting the test naming conventions aside, I am happy with this as a simple feature/fix. 
/approve

pkg/generate/code/set_resource_test.go
```diff
@@ -3273,6 +3273,336 @@ func TestSetResource_EC2_DHCPOptions_NestedSetConfig(t *testing.T) {
...
+	}
+	if resp.DeadLetterConfig != nil {
+		kof2 := &svcapitypes.DeadLetterConfig{}
+		if resp.DeadLetterConfig.Arn != nil {
+			kof2.ARN = resp.DeadLetterConfig.Arn
+		}
+		ko.DeadLetterConfig = kof2
+	}
+	if resp.EcsParameters != nil {
+		kof3 := &svcapitypes.EcsParameters{}
```
jaypipes, line 3332: Hmm, this should be `ECSParameters`... we need to [add ECS to the initialisms](https://github.com/aws-controllers-k8s/pkg/blob/main/names/names.go#L101).

pkg/generate/code/set_resource_test.go
```diff
@@ -3273,6 +3273,336 @@ func TestSetResource_EC2_DHCPOptions_NestedSetConfig(t *testing.T) {
...
+				}
+				if kof3f0iter.Weight != nil {
+					kof3f0elem.Weight = kof3f0iter.Weight
+				}
+				kof3f0 = append(kof3f0, kof3f0elem)
+			}
+			kof3.CapacityProviderStrategy = kof3f0
+		}
+		if resp.EcsParameters.EnableECSManagedTags != nil {
+			kof3.EnableECSManagedTags = resp.EcsParameters.EnableECSManagedTags
```
jaypipes, line 3351: Interesting that they used "ECS" instead of "Ecs" in this field name...

pkg/generate/code/set_resource_test.go
```diff
@@ -3273,6 +3273,336 @@ func TestSetResource_EC2_DHCPOptions_NestedSetConfig(t *testing.T) {
...
+		if resp.SqsParameters.MessageGroupId != nil {
+			kof15.MessageGroupID = resp.SqsParameters.MessageGroupId
+		}
+		ko.SQSParameters = kof15
+	}
+`
+	assert.Equal(
+		expected,
+		code.SetResourceForStruct(
+			crd.Config(), crd, "", "ko", koShape, nil, "resp", targetShape, "", model.OpTypeList, 1,
```
jaypipes, line 3601: Turns out that the [`targetField`](https://github.com/aws-controllers-k8s/code-generator/blob/c6651c200ba136e5c7f50ad8be751fae060a38e6/pkg/generate/code/set_resource.go#L1258-L1259) parameter isn't ever used by the `SetResourceForStruct` function :( Only the [`targetFieldPath`](https://github.com/aws-controllers-k8s/code-generator/blob/c6651c200ba136e5c7f50ad8be751fae060a38e6/pkg/generate/code/set_resource.go#L1270-L1271) parameter is ever used.

My request is that you:

1) remove the unused `targetField` parameter (since it is just confusing) from the `SetResourceForStruct` function
2) Pass `fX` for the *targetVarName* parameter (normally this would be something like "f12" if `SetResourceForStruct` is being called recursively from other functions in `pkg/generate/code/set_resource.go` *or* it would be the name of the field if `SetResourceForStruct` was called from a top-level template in `sdk_file_end` hooks...)

jaypipes (changes requested):

jaypipes (approved):
Much better. Thank you sir! :)

PR #388, opened 2023-01-06
Add Helm support for the `--deletion-policy` CLI arg

RedbackThomson:
Description of changes:
Adds a new value in `values.yaml` named `deletionPolicy` to support the controller CLI argument `--deletion-policy`. Supports values of `delete` and `retain`

### DIFF
--- a/templates/helm/values.schema.json
+++ b/templates/helm/values.schema.json
@@ -203,6 +203,10 @@
         "pattern": "(^$|^.*=.*$)"
       }
     },
+    "deletionPolicy": {
+      "type": "string",
+      "enum": ["delete", "retain"]
+    },
     "serviceAccount": {
       "description": "ServiceAccount settings",
       "properties": {
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/aws-controllers-k8s/code-generator
 go 1.17
 
 require (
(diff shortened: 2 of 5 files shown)


RedbackThomson:
/hold
Requires release of https://github.com/aws-controllers-k8s/runtime/pull/107 and a runtime version update

templates/helm/values.schema.json
```diff
@@ -203,6 +203,10 @@
         "pattern": "(^$|^.*=.*$)"
       }
     },
+    "deletionPolicy": {
+      "type": "string",
+      "enum": ["delete", "retain"]
```
jljaco, line 208: Since this enum is already enforced by the runtime [here](https://github.com/RedbackThomson/ack-runtime/blob/1e572bced2a58f806c54e04fe09c12eac307c6b0/apis/core/v1alpha1/deletion_policy.go#L25), what do you think about just making this a free-form string so that there is one fewer place where an enum needs to be updated if the values ever change?

RedbackThomson: I want this here because I'd like the client-side Helm validation to fail if they try to add anything other than those values. Otherwise they would have to check the controller pod logs to understand what went wrong

jaypipes (approved):
:+1:

RedbackThomson:
/unhold
Updated with the latest runtime `v0.22.1`

PR #390, opened 2023-01-14
add: specific selector for operator

codekow:
Issue #, if available: [#1632](https://github.com/aws-controllers-k8s/community/issues/1632)

Description of changes:

Improve label selection for service and deployment association with pods. When multiple operators are deployed you may get inconsistent results on the service.

Related [s3-controller/pull/94](https://github.com/aws-controllers-k8s/s3-controller/pull/94)

### DIFF
--- a/templates/config/controller/deployment.yaml.tpl
+++ b/templates/config/controller/deployment.yaml.tpl
@@ -1,8 +1,6 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  labels:
-    app.kubernetes.io/name: ack-system
   name: ack-system
 ---
 apiVersion: apps/v1
--- a/templates/config/controller/deployment.yaml.tpl
+++ b/templates/config/controller/deployment.yaml.tpl
@@ -16,11 +16,13 @@ spec:
   selector:
     matchLabels:
       control-plane: controller
+      name: ack-{{ .ServicePackageName }}-controller
   replicas: 1
   template:
     metadata:
       labels:
         control-plane: controller
+        name: ack-{{ .ServicePackageName }}-controller
     spec:
       containers:
       - command:
--- a/templates/config/controller/deployment.yaml.tpl
+++ b/templates/config/controller/deployment.yaml.tpl
@@ -3,7 +3,6 @@ kind: Namespace
 metadata:
   labels:
     control-plane: controller
-    app.kubernetes.io/name: ack-system 
   name: ack-system
 ---
 apiVersion: apps/v1
--- a/templates/config/controller/deployment.yaml.tpl
+++ b/templates/config/controller/deployment.yaml.tpl
@@ -3,6 +3,7 @@ kind: Namespace
 metadata:
   labels:
     control-plane: controller
+    app.kubernetes.io/name: ack-system 
   name: ack-system
 ---
 apiVersion: apps/v1
@@ -12,18 +13,19 @@ metadata:
   namespace: ack-system
   labels:
     control-plane: controller
-    name: ack-{{ .ServicePackageName }}-controller
+    app.kubernetes.io/name: ack-{{ .ServicePackageName }}-controller
+    app.kubernetes.io/part-of: ack-system
 spec:
   selector:
     matchLabels:
       control-plane: controller
-      name: ack-{{ .ServicePackageName }}-controller
+      app.kubernetes.io/name: ack-{{ .ServicePackageName }}-controller
   replicas: 1
   template:
     metadata:
       labels:
         control-plane: controller
-        name: ack-{{ .ServicePackageName }}-controller
+        app.kubernetes.io/name: ack-{{ .ServicePackageName }}-controller
     spec:
       containers:
       - command:
--- a/templates/config/controller/deployment.yaml.tpl
+++ b/templates/config/controller/deployment.yaml.tpl
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: Namespace
 metadata:
   labels:
-    control-plane: controller
+    app.kubernetes.io/name: ack-system
   name: ack-system
 ---
 apiVersion: apps/v1
@@ -11,19 +11,16 @@ metadata:
   name: ack-{{ .ServicePackageName }}-controller
   namespace: ack-system
   labels:
-    control-plane: controller
     app.kubernetes.io/name: ack-{{ .ServicePackageName }}-controller
     app.kubernetes.io/part-of: ack-system
 spec:
   selector:
     matchLabels:
-      control-plane: controller
       app.kubernetes.io/name: ack-{{ .ServicePackageName }}-controller
   replicas: 1
   template:
     metadata:
       labels:
-        control-plane: controller
         app.kubernetes.io/name: ack-{{ .ServicePackageName }}-controller
     spec:
       containers:
--- a/templates/config/controller/deployment.yaml.tpl
+++ b/templates/config/controller/deployment.yaml.tpl
@@ -12,6 +12,7 @@ metadata:
   namespace: ack-system
   labels:
     control-plane: controller
+    name: ack-{{ .ServicePackageName }}-controller
 spec:
   selector:
     matchLabels:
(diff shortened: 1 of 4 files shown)


codekow:
Addresses https://github.com/aws-controllers-k8s/community/issues/1632

codekow:
See https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels

Considering above, I have updated the minimal labels to get basic selection functionality without removing any existing labels.

templates/config/controller/deployment.yaml.tpl
```diff
@@ -3,6 +3,7 @@ kind: Namespace
 metadata:
   labels:
     control-plane: controller
+    app.kubernetes.io/name: ack-system
```
jaypipes, line 6: I would leave this one out, since the Namespace isn't part of the "application" but rather all the parts of the "ack-system" application are Deployments.

jaypipes (changes requested):
One tiny request otherwise I dig it. :)

jaypipes (commented):
Also, one other thing, @codekow: we have corresponding Helm templates in [`templates/helm`](https://github.com/aws-controllers-k8s/code-generator/tree/main/templates/helm) too.

templates/config/controller/deployment.yaml.tpl
codekow: @jaypipes Does every ack controller install or attempt to install in the `ack-system` namespace?

codekow: > Also, one other thing, @codekow: we have corresponding Helm templates in [`templates/helm`](https://github.com/aws-controllers-k8s/code-generator/tree/main/templates/helm) too.

I'll take a look at the helm charts. I can finish up this PR and associate another PR for the helm charts if it starts getting too complex.

jaypipes: > @jaypipes Does every ack controller install or attempt to install in the `ack-system` namespace?

Yep!

jaypipes: > > Also, one other thing, @codekow: we have corresponding Helm templates in [`templates/helm`](https://github.com/aws-controllers-k8s/code-generator/tree/main/templates/helm) too.
> 
> I'll take a look at the helm charts. I can finish up this PR and associate another PR for the helm charts if it starts getting too complex.

Cool. Let me know if I can help you in any way. Happy to assist! :)

codekow: > > @jaypipes Does every ack controller install or attempt to install in the `ack-system` namespace?
> 
> Yep!

That is the reason I would consider the namespace as part of the `ack-system` since any controller has the namespace as a dependency?

@jaypipes What do you think?

jaypipes: I still prefer to leave it off the Namespace. That said, I'm just one opinion... @A-Hilaly @acornett21 @RedbackThomson @jljaco what say you? :)

acornett21: > > @jaypipes Does every ack controller install or attempt to install in the `ack-system` namespace?
> 
> Yep!

@jaypipes This is not always the case, in OpenShift `ack-system` is the recommended namespace that a cluster admin should select, but the user can also choose the `openshift-operators` namespace from the drop-down in the UI as well. Since this is the case, `Namespace` should be excluded.

acornett21:
I'd like to see the `deployment.label` section of the `CSV`  that's output for the S3 (or any other controller).

The below will generate a file at `..olm/bundle/manifests/ack-s3-controller.clusterserviceversion.yaml` in the S3 controller.
```
export SERVICE=s3
ACK_GENERATE_OLM=true make build-controller SERVICE=$SERVICE
```

I don't think we have any `post` hocks to change this, but I'd like to be certain.

jaypipes (approved):
:+1:

codekow:
> export SERVICE=s3
> ACK_GENERATE_OLM=true make build-controller SERVICE=$SERVICE

Checking `..olm/bundle/manifests/ack-s3-controller.clusterserviceversion.yaml`...

```
spec:
  install:
    spec:
      deployments:
      - label:
          app.kubernetes.io/name: ack-s3-controller
          app.kubernetes.io/part-of: ack-system
          control-plane: controller
        name: ack-s3-controller
        spec:
          replicas: 1
          selector:
            matchLabels:
              app.kubernetes.io/name: ack-s3-controller
              control-plane: controller
          strategy: {}
          template:
            metadata:
              labels:
                app.kubernetes.io/name: ack-s3-controller
                control-plane: controller
```

codekow:
FYI @acornett21  ^^

codekow:
Do we want to leave the label: `control-plane: controller`?

@jaypipes @acornett21

jaypipes:
> Do we want to leave the label: `control-plane: controller`?
> 
> @jaypipes @acornett21

We don't need to, no. I put that on there originally when I copied some stuff from the node-termination-handler project. I think that label is no longer used by anything.

acornett21:
> Do we want to leave the label: `control-plane: controller`?
> 
> @jaypipes @acornett21

I pretty sure no project that's dependent on ACK in OpenShift uses this label, so I'm fine removing it as well.

codekow:
@jaypipes @acornett21 

Old label removed. I believe the merge is ready. Let me know if there is anything else to resolve with this PR. 

Thanks for your help! :smile:

acornett21 (approved):

templates/config/controller/deployment.yaml.tpl
RedbackThomson: I feel strongly that we should leave the annotation off the namespace. My three primary reasons:
1. Installing through Helm, we do not install it with a namespace (Helm takes care of the namespace) and therefore it will not have the annotation
2. The namespace is not exclusively for the ACK controllers. It's the default path for the controllers, but can be used by any other K8s objects, so it would be incorrect to label it as the ACK controller namespace
3. I have not found any other operators that annotate the namespace

RedbackThomson:
@codekow I just reopened a comment regarding the namespace annotation. I know there was some disagreement, but I feel relatively strongly that it's not the right move for now. I'd also rather leave it off, and add it later, than have an annotation that we need to support for backward compatibility sake.

templates/config/controller/deployment.yaml.tpl
jaypipes: Thanks @RedbackThomson for chiming in here. :) That echoes my feelings as well.

templates/config/controller/deployment.yaml.tpl
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: Namespace
 metadata:
   labels:
-    control-plane: controller
+    app.kubernetes.io/name: ack-system
```
jaypipes, line 5: Yeah, I think we should remove this.

jaypipes (changes requested):

templates/config/controller/deployment.yaml.tpl
A-Hilaly: Agreed with @acornett21, @jaypipes  and @RedbackThomson here, ACK controllers can be installed in any namespace and can manage resources from any namespace, even the caches can watch resources outside of the `ack-system` namespace. https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/runtime/cache/cache.go#L48-L59

templates/config/controller/deployment.yaml.tpl
codekow: Label removed 👍

templates/config/controller/deployment.yaml.tpl
codekow: @RedbackThomson @A-Hilaly @jaypipes @acornett21 

Agree with all points. I work in OpenShift and everything you said aligns. Thanks for going through the discussion.

codekow:
@jaypipes @A-Hilaly @RedbackThomson 

I believe all discussed items have been addressed. The PR should be ready to test.

A-Hilaly (approved):
++ Thank you @codekow !

jaypipes (approved):
Nice stuff here, let's get this merged! :)

RedbackThomson:
Great. Thank you!
/lgtm

PR #393, opened 2023-01-27
fix required fields that are renamed

jaypipes:
A previous commit that used the `github.com/aws-controllers-k8s/pkg/names` package instead of the local `pkg/names` package changed the implementation of the `Field.IsRequired` method slightly by no longer referring to the `names.ModelOriginal` value. This ModelOriginal value was really just a hack from a while ago because we didn't have support for reverse-lookups of renamed fields to original Input/Output struct member names.

This patch adds that reverse lookup and fixes the IsRequired method to properly look up the original struct member name in the Input shape's required list.

### DIFF
--- a/pkg/config/resource.go
+++ b/pkg/config/resource.go
@@ -145,10 +145,11 @@ type SyncedCondition struct {
 // Example usage from the AmazonMQ generator config:
 //
 // resources:
-//   Broker:
-//     hooks:
-//       sdk_update_pre_build_request:
-//        code: if err := rm.requeueIfNotRunning(latest); err != nil { return nil, err }
+//
+//	Broker:
+//	  hooks:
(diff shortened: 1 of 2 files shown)


jaypipes:
@A-Hilaly @RedbackThomson @jljaco Please note the changes to comments are from `gofmt` changing in go1.19, which is what I'm using locally now and is integrated into my vim, so when I save a file, it automatically `gofmt`s the code...

jljaco (approved):
Looks good to me, I'll leave it to @A-Hilaly or @RedbackThomson for final approval

RedbackThomson (approved):
Looks fair to me!

A-Hilaly (commented):
Thank you @jaypipes! nice catch.

PR #394, opened 2023-01-27
support field custom documentation strings

jaypipes:
Adds support for extending the documentation strings for fields and setting docstrings for custom fields.

A new `FieldConfig.Documentation` configuration option is now supported. When present, this string *extends* any documentation provided for the existing field by the doc-2.json files. If no such documentation is provided by the existing field OR if the field is a custom field, the `FieldConfig.Documentation` configuration option is used for the field's docstring in the Go type files.

Example usage for an existing nested field (from a hypothetical ec2 generator.yaml file). The
`LaunchTemplateData.HibernationOptions.Configured` field has a docstring already that contains the following:

```
// If you set this parameter to true, the instance is enabled for hibernation.
//
// Default: false
```

Setting the `generator.yaml` file to the following:

```yaml
resources:
  LaunchTemplate:
    fields:
      LaunchTemplateData.HibernationOptions.Configured:
        documentation: XXX extended docs XXX
```

would cause the field's Documentation string to look like this:

```
// If you set this parameter to true, the instance is enabled for hibernation.
//
// Default: false
//
// XXX extended docs XXX",
```

Issue aws-controllers-k8s/community#1223

### DIFF
--- a/pkg/config/field.go
+++ b/pkg/config/field.go
@@ -13,7 +13,9 @@
 
 package config
 
-import "strings"
+import (
+	"strings"
+)
 
 // SourceFieldConfig instructs the code generator how to handle a field in the
 // Resource's SpecFields/StatusFields collection that takes its value from an
@@ -49,23 +51,23 @@ import "strings"
 // CreateFunction API call, which has a `Code` member of its Input shape that
 // looks like this:
 //
-// "Code": {
-//   "ImageUri": "string",
-//   "S3Bucket": "string",
-//   "S3Key": "string",
-//   "S3ObjectVersion": "string",
-//   "ZipFile": blob
-// },
+//	"Code": {
+//	  "ImageUri": "string",
+//	  "S3Bucket": "string",
+//	  "S3Key": "string",
+//	  "S3ObjectVersion": "string",
+//	  "ZipFile": blob
+//	},
 //
 // The GetFunction API call's Output shape has a same-named field called
 // `Code` in it, but this field looks like this:
 //
-// "Code": {
-//   "ImageUri": "string",
-//   "Location": "string",
-//   "RepositoryType": "string",
-//   "ResolvedImageUri": "string"
-// },
+//	"Code": {
+//	  "ImageUri": "string",
+//	  "Location": "string",
+//	  "RepositoryType": "string",
+//	  "ResolvedImageUri": "string"
+//	},
 //
 // This presents a conundrum to the ACK code generator, which, as noted above,
 // assumes the data types of same-named fields in the Create Operation's Input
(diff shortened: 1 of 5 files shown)


jaypipes:
@RedbackThomson @jljaco @A-Hilaly please note the changes to comment formatting is from using go1.19 gofmt changes...

RedbackThomson:
I have some concerns about how unruly the `generator.yaml` file will become if there are going to be lots of custom fields. (cough *S3*). Rather, I'd prefer if we could track the documentation in a separate file and instead add pointers to that file on a per-resource level. That way documentation could stay within the `apis/v1alpha1` directory? So either we could have:
```yaml
resource:
  Cluster:
    documentation_file: docs/cluster.yaml
```

and then a `apis/v1alpha1/docs/cluster.yaml` file containing:
```yaml
field1: >
  Multi-line
  Documentation
  Here
field2: ...
```

Or we could have a per-controller version of that where we define a top-level config in the `generator.yaml` like so:
```yaml
documentation_file: docs.yaml
```

Then at `apis/v1alpha1/docs.yaml`
```yaml
Cluster:
  field1: ...
Addons:
  field1: ...
```

jaypipes:
> I have some concerns about how unruly the `generator.yaml` file will become if there are going to be lots of custom fields. (cough _S3_). Rather, I'd prefer if we could track the documentation in a separate file and instead add pointers to that file on a per-resource level. That way documentation could stay within the `apis/v1alpha1` directory? So either we could have:
> 
> ```yaml
> resource:
>   Cluster:
>     documentation_file: docs/cluster.yaml
> ```
> 
> and then a `apis/v1alpha1/docs/cluster.yaml` file containing:
> 
> ```yaml
> field1: >
>   Multi-line
>   Documentation
>   Here
> field2: ...
> ```
> 
> Or we could have a per-controller version of that where we define a top-level config in the `generator.yaml` like so:
> 
> ```yaml
> documentation_file: docs.yaml
> ```
> 
> Then at `apis/v1alpha1/docs.yaml`
> 
> ```yaml
> Cluster:
>   field1: ...
> Addons:
>   field1: ...
> ```

@RedbackThomson I guess I see where you're coming from if the controller has a LOT of custom fields. s3-controller only actually has 4 fields though :) So not too onerous to have the docs in a single generator.yaml file IMHO.

Then again, I do like your suggestion about having a `documentation_file` top-level configuration option. I could also see a possible `doc_from` configuration option where, say for custom fields, you tell ack-generate where to find the documentation for a particular field (basically give it the Shape and Member path in the doc-2.json file to use)

RedbackThomson:
> @RedbackThomson I guess I see where you're coming from if the controller has a LOT of custom fields. s3-controller only actually has 4 fields though :) So not too onerous to have the docs in a single generator.yaml file IMHO.
> 
> Then again, I do like your suggestion about having a `documentation_file` top-level configuration option. I could also see a possible `doc_from` configuration option where, say for custom fields, you tell ack-generate where to find the documentation for a particular field (basically give it the Shape and Member path in the doc-2.json file to use)

Both sound great! How about:
```yaml
resources:
  Cluster:
    fields:
      MyCustomField:
        documentation: # Supports XOR `from_sdk` and `from_custom`
          from_sdk:
            operation: ...
            path: ...
          from_custom: ... 
```

A-Hilaly:
I'm a bit mixed regarding using new files different than `generator.yaml`, but i also agree that adding documentation in the "fields" config can be super verbose, how about doing something similar to what's below?
```yaml
resources:
  User:
    fields:
       newField:
          documentation:
             from: X
  Group:
     fields:
        newField:
           documentation:
              from: X

documentation:
  X: is an option that does ... # This can be reused many times (thinking resources that have similar fields like ARN etc...)
  Y: is an option that does ...
```

A-Hilaly (approved):
👍

RedbackThomson:
Even though I still think there should be some UX changes to this feature, I am going to merge this because I want to use it to inject deprecation notices into fields. 
/lgtm

PR #397, opened 2023-02-01
Use `go env GOARCH` instead of `uname -i`

A-Hilaly:
Since `-i` flag is not supported for the `uname` binary published in the
darwin operating systems, we're switching to `go env GOARCH` to keep the
script working on both linux and darwin OSs.


jljaco (approved):

RedbackThomson:
Ok yes love this
/lgtm

PR #401, opened 2023-02-04
Refactor reference resolution into Go code

RedbackThomson:
Description of changes:
Refactors the `resolveReferenceFor<Field>` Go template code into direct Go code. This way we can be more intelligent about nested field references and support unit testing its output. 

Breaking change:
The previous Go template produced valid, but ultimately faulty, code for references within lists of structs. These are not supported (see https://github.com/aws-controllers-k8s/community/issues/1291). As such, this update produces a panic when it detects those types of references.


pkg/generate/code/resource_reference.go
```diff
@@ -179,6 +185,187 @@ func ReferenceFieldsPresent(
...
+
+		ref := cur.ShapeRef
+
+		indent := strings.Repeat("\t", indentLevel+idx)
+
+		if ref.Shape.Type == "structure" {
+			fieldAccessPrefix = fmt.Sprintf("%s.%s", fieldAccessPrefix, fp.At(idx))
+
+			outPrefix += fmt.Sprintf("%sif %s != nil {\n", indent, fieldAccessPrefix)
+			outSuffix = fmt.Sprintf("%s}\n%s", indent, outSuffix)
```
A-Hilaly, line 266: I like this! reminds me that we can make it a wrapper function

pkg/generate/code/resource_reference.go
```diff
@@ -179,6 +185,187 @@ func ReferenceFieldsPresent(
...
+	out += fmt.Sprintf("%s\tif refResourceTerminal {\n", indent)
+	out += fmt.Sprintf("%s\t\treturn ackerr.ResourceReferenceTerminalFor(\n", indent)
+	out += fmt.Sprintf("%s\t\t\t\"%s\",\n", indent, field.FieldConfig.References.Resource)
+	out += fmt.Sprintf("%s\t\t\tnamespace, *arr.Name)\n", indent)
+	out += fmt.Sprintf("%s\t}\n", indent)
+	out += fmt.Sprintf("%s\tif !refResourceSynced {\n", indent)
+	out += fmt.Sprintf("%s\t\treturn ackerr.ResourceReferenceNotSyncedFor(\n", indent)
+	out += fmt.Sprintf("%s\t\t\t\"%s\",\n", indent, field.FieldConfig.References.Resource)
+	out += fmt.Sprintf("%s\t\t\tnamespace, *arr.Name)\n", indent)
+	out += fmt.Sprintf("%s\t}\n", indent)
```
A-Hilaly, line 347: Ooof, i guess the only other way is using templates or multiline strings, but honestly, between all these not-so-good options i prefer what you used here...

pkg/generate/code/resource_reference.go
```diff
@@ -179,6 +185,187 @@ func ReferenceFieldsPresent(
...
+	out += fmt.Sprintf("%s\t\t}\n", indent)
+	out += fmt.Sprintf("%s\t}\n", indent)
+	out += fmt.Sprintf("%s\tif refResourceTerminal {\n", indent)
+	out += fmt.Sprintf("%s\t\treturn ackerr.ResourceReferenceTerminalFor(\n", indent)
+	out += fmt.Sprintf("%s\t\t\t\"%s\",\n", indent, field.FieldConfig.References.Resource)
+	out += fmt.Sprintf("%s\t\t\tnamespace, *arr.Name)\n", indent)
+	out += fmt.Sprintf("%s\t}\n", indent)
+	out += fmt.Sprintf("%s\tif !refResourceSynced {\n", indent)
+	out += fmt.Sprintf("%s\t\treturn ackerr.ResourceReferenceNotSyncedFor(\n", indent)
+	out += fmt.Sprintf("%s\t\t\t\"%s\",\n", indent, field.FieldConfig.References.Resource)
```
A-Hilaly, line 345: I see that you have mutiple instances of `\"%s\"`, you can also use `%q` to format a string + adding double quotes around it

A-Hilaly (commented):
Great one! Thank you for investing time in this. It's super useful, and testable...
Left a few comments and suggestions bellow

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.2.1
 	github.com/stretchr/testify v1.7.1
-	golang.org/x/mod v0.4.2
+	golang.org/x/mod v0.6.0
```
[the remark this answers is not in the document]
RedbackThomson: I honestly can't tell you where this came from. Go modules are a mystery to me

pkg/generate/code/resource_reference_test.go
```diff
@@ -198,3 +198,265 @@ if ko.Spec.Routes != nil {
...
+
+	crd := testutil.GetCRDByName(t, g, "Bucket")
+	require.NotNil(crd)
+
+	// the Go template has the appropriate nil checks to ensure the parent path exists
+	expected :=
+		`	if ko.Spec.Logging != nil {
+		if ko.Spec.Logging.LoggingEnabled != nil {
+			refVal := ""
+			if ko.Spec.Logging.LoggingEnabled.TargetBucketRef != nil && ko.Spec.Logging.LoggingEnabled.TargetBucketRef.From != nil {
```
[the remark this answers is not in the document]
RedbackThomson: I thought the same too except that the nil check we currently have only works for nested structs. As soon as you introduce a slice into the nesting it breaks.

pkg/generate/code/resource_reference.go
RedbackThomson: I tried using both. Multiline strings get ugly when you need to do the indentation because you then need to iterate over the string line-by-line and insert the indent. Templates might be the better option here but then you need to set up all that templating setup code. Not ideal and not super readable as it is :/

pkg/generate/code/resource_reference.go
```diff
@@ -179,6 +185,187 @@ func ReferenceFieldsPresent(
...
+//		if refResourceTerminal {
+//			return ackerr.ResourceReferenceTerminalFor(
+//				"Integration",
+//				namespace, *arr.Name)
+//		}
+//		if !refResourceSynced {
+//			return ackerr.ResourceReferenceNotSyncedFor(
+//				"Integration",
+//				namespace, *arr.Name)
+//		}
```
jaypipes, line 230: Seems to me that we could create a function in the `references.go.tpl` template file that handled this in a standard fashion:

```go
{{- objType := ( printf "%sapitypes.%s" .ReferencedServiceName .FieldConfig.References.Resource ) -}}
{{- if eq .FieldConfig.References.ServiceName "" -}}
{{- objType := ( printf "svcapitypes.%s" {{ .FieldConfig.References.Resource) -}}
{{ end -}}
// getReferencedResourceState_{{ .FieldConfig.References.Resource }} looks up whether a referenced resource
// exists and is in a ACK.ResourceSynced=True state. If the referenced resource does exist and is
// in a Synced state, returns nil, otherwise returns `ackerr.ResourceReferenceTerminalFor` or
// `ResourceReferenceNotSyncedFor` depending on if the resource is in a Terminal state.
func getReferencedResourceState_{{ .FieldConfig.References.Resource }}(
	obj *{{ objType }},
	name string, // the Kubernetes name of the referenced resource
) error {
	namespacedName := types.NamespacedName{
		Namespace: namespace,
		Name: name,
	}
	err := apiReader.Get(ctx, namespacedName, &obj)
	if err != nil {
		return err
	}
	var refResourceSynced, refResourceTerminal bool
	for _, cond := range obj.Status.Conditions {
		if cond.Type == ackv1alpha1.ConditionTypeResourceSynced &&
			cond.Status == corev1.ConditionTrue {
			refResourceSynced = true
		}
		if cond.Type == ackv1alpha1.ConditionTypeTerminal &&
			cond.Status == corev1.ConditionTrue {
			return ackerr.ResourceReferenceTerminalFor(
				"{{ .FieldConfig.References.Resource }}",
				namespace, name)
		}
	}
	if !refResourceSynced {
		return ackerr.ResourceReferenceNotSyncedFor(
			"{{ .FieldConfig.References.Resource }}",
			namespace, name)
	}
	return nil
}
```

We'd generate the above functions for each referenced field. So, imagine we had 2 reference fields called `SubnetID` and `VPCID`, we'd generate two functions: `getReferencedResourceState_SubnetID()` and `getReferencedResourceState_VPCID()`.

You would then be able to call those functions from your Go code-generating functions instead of cluttering the Go code-generating functions with heavy indentation. It would also make things easier to debug and troubleshoot since the very common "look this up and see if the resource is synced" logic is isolated into a single function, specialized for that reference field.

pkg/generate/code/resource_reference.go
jaypipes, line 347: to follow on from my comment above, all of this code could then be replaced with:

```go
out += fmt.Sprintf(
	"%sif err = getReferencedResourceState_%s(%s, *arr.Name); err != nil { return err }\n",
	index, field.Names.Camel, sourceVarName,
)
```

pkg/generate/code/resource_reference.go
```diff
@@ -179,6 +185,187 @@ func ReferenceFieldsPresent(
...
+			outPrefix += fmt.Sprintf("%s%s = &%s\n", indent, targetVarName, refTarget)
+		}
+	}
+
+	return outPrefix + outSuffix
+}
+
+func resolveSingleReference(field *model.Field, sourceVarName string, targetVarName string, shouldAppendTarget bool, indentLevel int) string {
+	out := ""
+	indent := strings.Repeat("\t", indentLevel)
```
jljaco, line 307: Can you just use https://pkg.go.dev/go/format#Source to deal with all of the formatting of the end result instead of bothering will all of these indentations?

pkg/generate/code/resource_reference_test.go
```diff
@@ -198,3 +198,265 @@ if ko.Spec.Routes != nil {
...
+	require := require.New(t)
+
+	g := testutil.NewModelForServiceWithOptions(t, "apigatewayv2",
+		&testutil.TestingModelOptions{
+			GeneratorConfigFile: "generator-with-reference.yaml",
+		})
+
+	crd := testutil.GetCRDByName(t, g, "Integration")
+	require.NotNil(crd)
+	expected :=
```
jljaco, line 213: Consider sticking this "expected" code into a separate file (perhaps under `testdata/`?) and compare its contents to the output below instead of embedding this value directly into the test

pkg/generate/code/resource_reference_test.go
```diff
@@ -198,3 +198,265 @@ if ko.Spec.Routes != nil {
...
+	require := require.New(t)
+
+	g := testutil.NewModelForServiceWithOptions(t, "apigatewayv2",
+		&testutil.TestingModelOptions{
+			GeneratorConfigFile: "generator-with-reference.yaml",
+		})
+
+	crd := testutil.GetCRDByName(t, g, "VpcLink")
+	require.NotNil(crd)
+	expected :=
```
jljaco, line 276: Ditto expected code in a separate file

pkg/generate/code/resource_reference_test.go
```diff
@@ -198,3 +198,265 @@ if ko.Spec.Routes != nil {
...
+	require := require.New(t)
+
+	g := testutil.NewModelForServiceWithOptions(t, "apigatewayv2",
+		&testutil.TestingModelOptions{
+			GeneratorConfigFile: "generator-with-nested-reference.yaml",
+		})
+
+	crd := testutil.GetCRDByName(t, g, "Authorizer")
+	require.NotNil(crd)
+	expected :=
```
jljaco, line 341: Ditto

pkg/generate/code/resource_reference_test.go
```diff
@@ -198,3 +198,265 @@ if ko.Spec.Routes != nil {
...
+	g := testutil.NewModelForServiceWithOptions(t, "s3",
+		&testutil.TestingModelOptions{
+			GeneratorConfigFile: "generator-with-nested-references.yaml",
+		})
+
+	crd := testutil.GetCRDByName(t, g, "Bucket")
+	require.NotNil(crd)
+
+	// the Go template has the appropriate nil checks to ensure the parent path exists
+	expected :=
```
jljaco, line 408: Ditto

pkg/generate/code/resource_reference.go
RedbackThomson: ++ Absolutely would love to do this programmatically. If we can do this as a full refactor for all methods in the code-generator.

pkg/generate/code/resource_reference_test.go
RedbackThomson: With my latest changes the expected code is vastly smaller. Probably small enough we can consider it staying inline.

jaypipes (changes requested):
I think you left one hard-coded string value in the code-generating function :) See comment inline..

pkg/generate/code/resource_reference_test.go
```diff
@@ -198,3 +198,127 @@ if ko.Spec.Routes != nil {
...
+	crd := testutil.GetCRDByName(t, g, "VpcLink")
+	require.NotNil(crd)
+	expected :=
+		`	ko.Spec.SecurityGroupIDs = []*string{}
+	for _, iter0 := range ko.Spec.SecurityGroupIDs {
+		arr := iter0.From
+		if arr == nil || arr.Name == nil || *arr.Name == "" {
+			return fmt.Errorf("provided resource reference is nil or empty: \"SecurityGroupRefs"\")
+		}
+		if err := getReferencedResourceState_VPCLink(ctx, apiReader, obj, *arr.Name, namespace); err != nil {
```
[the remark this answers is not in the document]
RedbackThomson: lmao oops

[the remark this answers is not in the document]
RedbackThomson: Don't you love test driven development

pkg/generate/code/resource_reference.go
```diff
@@ -185,6 +185,111 @@ func ReferenceFieldsPresent(
...
+			outPrefix += fmt.Sprintf("%s\tif arr == nil || arr.Name == nil || *arr.Name == \"\" {\n", indent)
+			outPrefix += fmt.Sprintf("%s\t\treturn fmt.Errorf(\"provided resource reference is nil or empty: \\%q\\\")\n", indent, field.ReferenceFieldPath())
+			outPrefix += fmt.Sprintf("%s\t}\n", indent)
+
+			outPrefix += fmt.Sprintf("%s\tif err := getReferencedResourceState_%s(ctx, apiReader, obj, *arr.Name, namespace); err != nil {\n", indent, field.FieldConfig.References.Resource)
+			outPrefix += fmt.Sprintf("%s\t\treturn err\n", indent)
+			outPrefix += fmt.Sprintf("%s\t}\n", indent)
+			outPrefix += fmt.Sprintf("%s\t%s = append(%s, obj.%s)\n", indent, targetVarName, targetVarName, field.FieldConfig.References.Path)
+			outPrefix += fmt.Sprintf("%s}\n", indent)
+		case ("map"):
```
A-Hilaly, line 265: Didn't know parenthesis here are possible

A-Hilaly (approved):
Neat! Good stuff :ok_hand:

PR #402, opened 2023-02-05
fix: Make CRD Synced Column type string

embano1:
Closes: https://github.com/aws-controllers-k8s/community/issues/1659

Description of changes:

Change CRD type from `boolean` to `string` for `SYNCED` column.


embano1:
@RedbackThomson the fix was easy, but not sure if we have E2E tests to assert it doesn't break existing controllers.

embano1:
I tried it with the `eventbridge` controller and the change fixes the issue.

PR #410, opened 2023-02-09
Fix default value of `--reconcile-resource` flag in deployment.yaml

A-Hilaly:
The runtime expects `--reconcile-resource-resync-seconds` to be an array
of string values. While in deployment.yaml we pass an empty string which
is invalid for `runtime/pkg/config`.

This patch fixes this issue by removing the
`--reconcile-resource-resync-seconds` value


RedbackThomson:
Thank you
/lgtm

jljaco:
This was to get tests passing in https://github.com/aws-controllers-k8s/code-generator/pull/409

PR #415, opened 2023-02-10
check for nil Name field in attr APIs

jaypipes:
The code which builds the input shapes for the GET_ATTRIBUTES and SET_ATTRIBUTES operations for attribute-based APIs was checking for a Name or ID field in order to set the value for ARN fields in the input shapes. However, some resources, like SNS' Subscription resources, do not have a Name or ID field.

This resulted in nil pointer dereferencing and panics when building the controller for SNS Subscriptions.

This patch simply double-checks that the `*string` returned from the `CRD.SpecIdentifierField()` method is not nil before adding in the Go code that constructs an ARN from a Name or ID field.

Related: aws-controllers-k8s/community#1491


A-Hilaly (approved):
Nice!

PR #417, opened 2023-02-13
Refactor reference resolution into Go code

RedbackThomson:
Continuation of https://github.com/aws-controllers-k8s/code-generator/pull/401

Refactors the `resolveReferenceFor<Field>` Go template code into direct Go code. This way we can be more intelligent about nested field references and support unit testing its output. 

Breaking change:
The previous Go template produced valid, but ultimately faulty, code for references within lists of structs. These are not supported (see https://github.com/aws-controllers-k8s/community/issues/1291). As such, this update produces a panic when it detects those types of references.


RedbackThomson:
I expect that `ec2-controller` fails the integration tests, since it unfortunately uses structs with references within lists - which is a known error condition for references. https://github.com/aws-controllers-k8s/community/issues/1291

This pull request puts a panic condition in, so that no new controllers can add fields before this condition is possible.

RedbackThomson:
All 3 test failures of the IAM controller are due to missing policy ARNs - unrelated to these changes but worthy of an investigation.

All the failures of the Lambda controller are within the `TestEventSourceMapping` class, none of which seem to be related to these changes either. There are other tests for references that are still passing.

A-Hilaly:
>All 3 test failures of the IAM controller are due to missing policy ARNs - unrelated to these changes but worthy of an investigation.
All the failures of the Lambda controller are within the TestEventSourceMapping class, none of which seem to be related to these changes either. There are other tests for references that are still passing.

Regarding IAM, tests are not failing in this PR @RedbackThomson maybe let's test with one commit earlier? https://github.com/aws-controllers-k8s/code-generator/pull/394

RedbackThomson:
> Regarding IAM, tests are not failing in this PR @RedbackThomson maybe let's test with one commit earlier? #394

You were right to suspect this. Thank you for pointing out that we were overwriting the `ko.Spec.Policies` list even if there were no references - always leading to an empty policy list. I have replaced it with a length check for the references slice, so that if there are no references, then nothing will change. I have also updated the code to put the resolved references into a new string slice and then only apply it to the spec if there are no errors.

pkg/generate/ack/controller.go
```diff
@@ -62,6 +62,16 @@ var (
...
+		"AddToMap": func(m map[string]interface{}, k string, v interface{}) map[string]interface{} {
+			if len(m) == 0 {
+				m = make(map[string]interface{})
+			}
+			m[k] = v
+			return m
+		},
+		"Nil": func() interface{} {
+			return nil
+		},
```
A-Hilaly, line 74: Interesting how this is returning an `interface{}` and plays well with `AddToMap`

A-Hilaly (approved):
Nice and neat!

PR #422, opened 2023-03-03
Allow manual setting of AWS client struct and interface names

A-Hilaly:
While developing the ACK pipes-controller, we encountered an issue where
the generated code didn't compile mainly because the code-generator
assumed that the AWS SDK client/interface/import-paths always followed
the same naming conventions. However, we discovered that Pipes was the
exception as they renamed the model, import path, and client
struct/interface.

This patch introduces a new feature that enables users to manually set
the client struct and interface names in the `generator.yaml` file.
This update allows for greater flexibility in the AWS client naming
within pipes-controller and possibly other controllers in the future.

In this patch, we've also introduced a breaking change by renaming
`model_name` to `sdk_names.model`. Mainly to keep all the SDK-related
naming conventions in the same place.

### DIFF
--- a/pkg/model/sdk_api.go
+++ b/pkg/model/sdk_api.go
@@ -327,9 +327,18 @@ func (a *SDKAPI) GetServiceFullName() string {
 	return a.API.Metadata.ServiceFullName
 }
 
-// APIInterfaceTypeName returns the name of the aws-sdk-go primary API
+// ClientInterfaceTypeName returns the name of the aws-sdk-go primary API
 // interface type name.
-func (a *SDKAPI) APIInterfaceTypeName() string {
+func (a *SDKAPI) ClientInterfaceTypeName() string {
+	if a == nil || a.API == nil {
+		return ""
+	}
+	return fmt.Sprintf("%s%s", a.API.StructName(), "API")
+}
+
+// ClientStructTypeName returns the name of the aws-sdk-go primary API
+// struct type name.
+func (a *SDKAPI) ClientStructTypeName() string {
 	if a == nil || a.API == nil {
 		return ""
 	}
--- a/cmd/ack-generate/command/common.go
+++ b/cmd/ack-generate/command/common.go
@@ -45,7 +45,7 @@ func loadModel(svcAlias string, apiVersion string, apiGroup string, defaultCfg a
 		return nil, err
 	}
 
-	modelName := strings.ToLower(cfg.ModelName)
+	modelName := strings.ToLower(cfg.SDKNames.Model)
 	if modelName == "" {
 		modelName = svcAlias
 	}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -39,12 +39,28 @@ type Config struct {
 	// SetManyOutput function fails with NotFound error.
 	// Default is "return nil, ackerr.NotFound"
 	SetManyOutputNotFoundErrReturn string `json:"set_many_output_notfound_err_return,omitempty"`
+	// SDKNames lets you specify SDK object names. This configuration field was
+	// introduces when we learned that the EventBridgePipes service renamed, not
+	// only the service model name, but also the iface name to `PipesAPI`. See
+	// https://github.com/aws/aws-sdk-go/blob/main/service/pipes/pipesiface/interface.go#L62
+	SDKNames SDKNames `json:"sdk_names"`
+}
+
+// SDKNames contains information on the SDK Client package. More precisely
+// it holds information on how to correctly import the {service}/iface main
+// interface and client structs.
+type SDKNames struct {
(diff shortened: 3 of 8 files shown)


jaypipes:
Pipes are always a problem.

jaypipes (changes requested):
@A-Hilaly @embano1 I think it would be better to just rely on the [`aws-sdk-go/private/model/api.API`](https://github.com/aws/aws-sdk-go/blob/5e5a69da19530ea083f5145638c52c893b091860/private/model/api/api.go#L120-L176) `InterfaceName` and `StructName` methods.

A-Hilaly:
>@A-Hilaly @embano1 I think it would be better to just rely on the [aws-sdk-go/private/model/api.API](https://github.com/aws/aws-sdk-go/blob/5e5a69da19530ea083f5145638c52c893b091860/private/model/api/api.go#L120-L176) InterfaceName and StructName methods.

Unfortunately, those don't work properly for us. The reason is that the `aws-sdk-go/pipes` package was generated using flags different than the default ones. And since we don't have access to those there is nothing we can do (i think)

jaypipes:
> > @A-Hilaly @embano1 I think it would be better to just rely on the [aws-sdk-go/private/model/api.API](https://github.com/aws/aws-sdk-go/blob/5e5a69da19530ea083f5145638c52c893b091860/private/model/api/api.go#L120-L176) InterfaceName and StructName methods.
> 
> @jaypipes Unfortunately, those don't work properly for us. The reason is that the `aws-sdk-go/pipes` package was generated using flags different than the default ones. And since we don't have access to those there is nothing we can do (i think)

@A-Hilaly can you prove this? I don't understand why those two methods won't work for this use case...

A-Hilaly:
@jaypipes [`api.InterfacePackageName`](https://github.com/aws/aws-sdk-go/blob/5e5a69da19530ea083f5145638c52c893b091860/private/model/api/api.go#L120-L123) returns `eventbridgepipesiface` (we need `pipesiface`) and `api.StructName` returns `EventBridgePipes` (we need `Pipes`/`PipesAPI`). Note that the values returned by `&api{}` are correct theoritically speaking, but pipes teams generated their controller using different struct/interface/package names: probably with something like
```
go run -tags codegen ./private/model/cli/gen-api/main.go -use-service-id -path pipes ./models/apis/pipes/2015-10-07/api-2.json
```

pkg/model/sdk_api.go
```diff
@@ -327,9 +327,18 @@ func (a *SDKAPI) GetServiceFullName() string {
...
 
-// APIInterfaceTypeName returns the name of the aws-sdk-go primary API
+// ClientInterfaceTypeName returns the name of the aws-sdk-go primary API
 // interface type name.
-func (a *SDKAPI) APIInterfaceTypeName() string {
+func (a *SDKAPI) ClientInterfaceTypeName() string {
+	if a == nil || a.API == nil {
+		return ""
+	}
+	return fmt.Sprintf("%s%s", a.API.StructName(), "API")
```
jaypipes, line 336: Should this be `a.API.InterfaceName()` by default?

A-Hilaly: There is no `a.API.InterfaceName` :/ . However there a `a.API.InterfacePackageName` but it return the package name that defines the interface.. which is not what we need here.

RedbackThomson (approved):
Happy to approve this. I'm a bit tentative about the naming, only because I don't fully understand the difference between these. Would you mind adding examples for each of these to the PR description for our reference? (examples of the normal default, and then what `pipes` does differently?)

pkg/model/sdk_api.go
jaypipes: Yeah, looking in the upstream code they indeed use `StructName` ... :/

jaypipes (approved):
:+1:

bobh66:
@jaypipes After this commit I'm seeing errors when running ack-generate crossplane for the prometheusservice in the crossplane-contrib/provider-aws project:

```
./ack-generate crossplane --aws-sdk-go-version v1.44.174 prometheusservice --output ../../
Error: service prometheusservice not found
+ exit 1
```
The Makefile is generating a list of services from the directories in https://github.com/crossplane-contrib/provider-aws/tree/master/apis that contain generator_config.yaml files.  This logic worked until these changes were applied.

I can dig into it more tomorrow but I thought I would check to see if it's something obvious that we can change on the provider-aws side.

Thanks

haarchri:
@bobh66 you need to switch 

model_name to sdk_names.model

https://github.com/crossplane-contrib/provider-aws/blob/master/apis/prometheusservice/generator-config.yaml#L1

A-Hilaly:
modified PR description to include the example you provided @bobh66 . Thank you folks!

PR #424, opened 2023-03-06
use `Operation.ExportedName` not `Operation.Name`

jaypipes:
This was a nasty issue to troubleshoot. :( Turns out that for REST+XML-based APIs (such as CloudFront), the aws-sdk-go `private/model/api.Operation` objects have a `Name` field that contains crap like this: "CreateCachePolicy2020_05_31". What this means is that when looking up things like output wrapper field paths by the `Operation.Name` of "CreateCachePolicy" (from the `generator.yaml` file's `operations` map), things were failing because the `private/model/api.API:Operations` map is actually keyed by `Operation.ExportedName`, not `Operation.Name`. `Operation.ExportedName` for REST+XML APIs like CloudFront contain the proper string "CreateCachePolicy" without the date suffix.

### DIFF
--- a/pkg/config/operation.go
+++ b/pkg/config/operation.go
@@ -56,7 +56,7 @@ func (c *Config) OperationIsIgnored(operation *awssdkmodel.Operation) bool {
 	if operation == nil {
 		return true
 	}
-	return util.InStrings(operation.Name, c.Ignore.Operations)
+	return util.InStrings(operation.ExportedName, c.Ignore.Operations)
 }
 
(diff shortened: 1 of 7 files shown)


A-Hilaly (approved):
+++ Thank you sir!

jaypipes:
@A-Hilaly @RedbackThomson found a few more uses of `op.Name` in the code gen and templates... please re-review.

A-Hilaly (approved):
👍

PR #430, opened 2023-03-20
Remove optGenVersion from api path for goimports

bobh66:
Description of changes:

`ack-generate crossplane` fails if there is no `apis/_service_/v1alpha1 directory`. The code includes optGenVersion in the apiPath for goimports which causes an error when the directory doesn't exist.

Removing optGenVersion from the apiPath causes `goimports` to run on all files/directorys under` apis/<service>`

### DIFF
--- a/cmd/ack-generate/command/crossplane.go
+++ b/cmd/ack-generate/command/crossplane.go
@@ -87,7 +87,7 @@ func generateCrossplane(_ *cobra.Command, args []string) error {
 			return err
 		}
 	}
-	apiPath := filepath.Join(optOutputPath, "apis", svcAlias, optGenVersion)
+	apiPath := filepath.Join(optOutputPath, "apis", svcAlias)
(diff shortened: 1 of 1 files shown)


haarchri:
@A-Hilaly can you have a look ? unblocks us for an PR https://github.com/crossplane-contrib/provider-aws/pull/1704

A-Hilaly:
ec2-controller failures are unrelated. Please ignore :)

bobh66:
@A-Hilaly the crossplane sub-command does not accept the --version flag so it fails with:
```
Error: unknown flag: --version
```
I would not expect any impact on previously generated CRDs as this is only affecting the execution of `goimports` to validate the generated API and controller files.

A-Hilaly:
Gotcha, thank you! merging this soon.
/lgtm

A-Hilaly (approved):

PR #435, opened 2023-04-06
Support lists of structs containing refs and `ClearResolvedReferences`

RedbackThomson:
Implements https://github.com/aws-controllers-k8s/runtime/pull/121

Description of changes:
This PR abstracts out the process of iterating through all references into a separate method called `iterReferenceValues`. `iterReferenceValues` produces Go code that drills down into the spec for every ref, iterating through any slices as necessary. Once it reaches the ref object, it calls a callback which can be used by other methods for logic pertaining to accessing the refs. 

This change allows `ResolveReferencesForField` to access all refs within lists of structs (or even within lists of lists of structs, etc.). `iterReferenceValues` is also used to implement `ClearResolvedReferencesForField`, which simply sets the concrete value to `nil` if it detects a non-nil value in the ref field (or `len > 0` for lists of refs). 

This PR also removes `hasNonNilReferences`. `hasNonNilReferences` was being used to indicate whether there were any references inside the resource, which required another set of iterating through all ref fields. Instead, each `resolveReferencesFor*` returns a boolean which indicates whether it found a reference during its iteration.

Below is a Gist that shows the output of running the code-generator on the current EC2 `generator.yaml`:
https://gist.github.com/RedbackThomson/8e11cbbe96065a4eb812c387665c747d

For testing:
`ec2` and `lambda` already have a series of tests for references. Both of these tests are passing, but they were also passing before this PR. I have a branch of the `ec2-controller` with additional tests using `RouteTables`, which have lists of structs containing refs, and all of those are passing as well. 


pkg/generate/code/resource_reference_test.go
```diff
@@ -72,18 +82,17 @@ func Test_ReferenceFieldsValidation_SliceOfReferences(t *testing.T) {
...
 	crd := testutil.GetCRDByName(t, g, "VpcLink")
 	require.NotNil(crd)
+
+	field := crd.Fields["SubnetIDs"]
 	expected :=
-		`	if ko.Spec.SecurityGroupRefs != nil && ko.Spec.SecurityGroupIDs != nil {
-		return ackerr.ResourceReferenceAndIDNotSupportedFor("SecurityGroupIDs", "SecurityGroupRefs")
-	}
-	if ko.Spec.SubnetRefs != nil && ko.Spec.SubnetIDs != nil {
+		`	if ko.Spec.SubnetRefs != nil && ko.Spec.SubnetIDs != nil {
```
A-Hilaly, line 88: When dealing with arrays best is to check the `len` - since arrays can be non nil still have 0 eleements e.g `subnetRefs: []`

pkg/generate/code/resource_reference_test.go
```diff
@@ -146,15 +187,33 @@ func Test_ReferenceFieldsPresent_SliceOfReferences(t *testing.T) {
...
+				return hasReferences, fmt.Errorf("provided resource reference is nil or empty: SecurityGroupRefs")
+			}
+			obj := &ec2apitypes.SecurityGroup{}
+			if err := getReferencedResourceState_SecurityGroup(ctx, apiReader, obj, *arr.Name, namespace); err != nil {
+				return hasReferences, err
+			}
+			if ko.Spec.SecurityGroupIDs == nil {
+				ko.Spec.SecurityGroupIDs = make([]*string, 0, 1)
+			}
+			ko.Spec.SecurityGroupIDs = append(ko.Spec.SecurityGroupIDs, (*string)(obj.Status.ID))
```
A-Hilaly, line 207: The more I think about this the more i believe that maybe we should not allow "normal" and "referenced" fields at the same time (`SecurityGroupIDs` and `SecurityGroupRefs`). Using both can lead us into an area of problems that is very hard to solve (unless if we want to leverage annotations, to keep track of what was referenced and was was not). Or even introduce a new Status field to help doing so (like Conditions and ACKResourceMetadata)

pkg/generate/code/resource_reference_test.go
```diff
@@ -31,13 +31,22 @@ func Test_ReferenceFieldsValidation_NoReferenceConfig(t *testing.T) {
...
-	assert.Equal(expected, code.ReferenceFieldsValidation(crd, "ko", 1))
+
+	field := crd.Fields["APIID"]
+	expected :=
+		`	if ko.Spec.APIRef != nil && ko.Spec.APIID != nil {
+		return ackerr.ResourceReferenceAndIDNotSupportedFor("APIID", "APIRef")
+	}
+	if ko.Spec.APIRef == nil && ko.Spec.APIID == nil {
+		return ackerr.ResourceReferenceOrIDRequiredFor("APIID", "APIRef")
+	}
```
A-Hilaly, line 44: One of the good use case of https://github.com/aws-controllers-k8s/pkg/pull/14/files#diff-4b1976072e607d66abb6e181c6699c9be57d46ada282e77ee704c6cdbebb17e7R36-R48 i believe. I know it's a pain to check for the zero values as well. But users could provide `apiID: ""` ...

A-Hilaly (commented):
Awesome, nice work on this mate 👍
 Left few comments. I will get back with a more constructive review once i test it this patch with one of the controllers!

pkg/generate/code/resource_reference_test.go
RedbackThomson: True, we do this in other places as well. I copied the old code for this one part, so I'll change it over to the new method.

pkg/generate/code/resource_reference_test.go
RedbackThomson: We don't allow a mix of concrete and refs. That's the point of the [`validateReferenceFields` method](https://gist.github.com/RedbackThomson/8e11cbbe96065a4eb812c387665c747d#file-references-go-L116).

pkg/generate/code/resource_reference_test.go
RedbackThomson: `lo` also supports [`IsEmpty`](https://github.com/samber/lo#isempty)

RedbackThomson:
The caveat to these changes that is still not fixed:

These changes allow for resolving references within lists of structs, but reference fields are removed (replaced with their concrete values) from the spec of the resources at the end of the create resource reconciliation. This is because `sdkFind` inevitably calls `Describe*` and the response of this call is to place the values into `spec`. Any part of the describe response that contains a list, `sdkFind` creates an entirely new list from the output of the describe and replaces the field in the spec with this. For example, in `ec2-controller` `RouteTable`, `Spec.Routes` comes from the `elem.Routes` of the `DescribeRouteTable` call - https://github.com/aws-controllers-k8s/ec2-controller/blob/main/pkg/resource/route_table/sdk.go#L149. Any references that may have been in `Spec.Routes` are now overwritten and cannot contain the original references.

A-Hilaly (approved):
👍

jljaco (approved):
/lgtm

PR #440, opened 2023-05-01
Update template to support watch namespace

rahtr:
As the controller runtime supports watching namespaces, we would need to add a support to watch namespaces other than where the controller is deployed. This is a common use case for multi-tenant K8s environments. This PR introduces a `watchNamespace` flag which lets user define the namespaces to be watched by the controller. If left blank, it would take the release namespace.

### DIFF
--- a/templates/helm/values.yaml.tpl
+++ b/templates/helm/values.yaml.tpl
@@ -72,6 +72,10 @@ log:
 # cluster wide.
 installScope: cluster
 
+# Set the value of the "namespace" to be watched by the controller
+# This would only work if installScope is set to "namespace". If left empty it would watch the release namespace by default.
+watchNamespace: ""
+
 resourceTags:
   # Configures the ACK service controller to always set key/value pairs tags on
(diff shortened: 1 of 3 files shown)


jaypipes (approved):
Sure, fine with me. Thanks @rahtr!

templates/helm/values.yaml.tpl
```diff
@@ -72,6 +72,10 @@ log:
 # cluster wide.
 installScope: cluster
 
+# Set the value of the "namespace" to be watched by the controller
+# This would only work if installScope is set to "namespace". If left empty it would watch the release namespace by default.
```
RedbackThomson, line 76: nit: Might just clarify what is meant here:
```suggestion
# This value is only used when the `installScope` is set to "namespace". If left empty, the default value is the release namespace for the chart.
```

A-Hilaly (approved):
nice addition, thank you @rahtr!
/lgtm

PR #441, opened 2023-05-02
Add new `$field.set[].To` configuration

A-Hilaly:
Issue https://github.com/aws-controllers-k8s/community/issues/1775

- Add new generator config to allow setting an `sdkField` from a specific CR field

### DIFF
--- a/pkg/config/field.go
+++ b/pkg/config/field.go
@@ -191,6 +191,19 @@ type SetFieldConfig struct {
 	// f17elem = *f17iter.DBSecurityGroupName
 	// ```
 	From *string `json:"from,omitempty"`
+	// To instructs the code generator to output Go code that sets the value of
+	// an Input sdkField with the content of a CR field.
+	//
+	// ```yaml
+	// resources:
+	//   User:
+	//     fields:
+	//       URL:
+	//         set:
+	//           - method: Update
+	//             to: NewURL
+	// ```
+	To *string `json:"to,omitempty"`
 	// Ignore instructs the code generator to ignore this field in the Output
 	// shape when setting the value of the resource's field in the Spec. This
 	// is useful when we know that, for example, the returned value of field in
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -415,6 +415,36 @@ func (r *CRD) GetPrimaryKeyField() (*Field, error) {
 	return primaryField, nil
 }
 
+// GetMatchingInputShapeFieldName returns the name of the field in the Input shape.
+// For simplicity, we assume that there will be only one setConfig for the
+// any unique sdkField, per operation. Which means that we will never set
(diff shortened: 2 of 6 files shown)


pkg/config/field.go
```diff
@@ -191,6 +191,14 @@ type SetFieldConfig struct {
 	// f17elem = *f17iter.DBSecurityGroupName
 	// ```
 	From *string `json:"from,omitempty"`
+	// resources:
+	//   User:
+	//     fields:
+	//       URL:
+	//         set:
+	//           - method: Update
+	//			   to: NewURL
```
jaypipes, line 200:
```suggestion
	//	     to: NewURL
```

pkg/config/field.go
```diff
@@ -191,6 +191,14 @@ type SetFieldConfig struct {
 	// f17elem = *f17iter.DBSecurityGroupName
 	// ```
 	From *string `json:"from,omitempty"`
+	// resources:
```
jaypipes, line 194: Can you include a description of the `To` field above this example YAML please? Also, I recommend encapsulating  the YAML with: ` ```yaml``` ` like the above `From` docstring.

pkg/model/crd.go
```diff
@@ -415,6 +415,35 @@ func (r *CRD) GetPrimaryKeyField() (*Field, error) {
 	return primaryField, nil
 }
 
+// For simplicity, we assume that there will be only one setConfig for the
```
jaypipes, line 418:
```suggestion
// GetMatchingInputShapeFieldName returns the name of the field in the Input shape
// in the SDK that we should set to the value of the supplied CRD field.
// For simplicity, we assume that there will be only one setConfig for the...
```

pkg/model/crd.go
```diff
@@ -415,6 +415,35 @@ func (r *CRD) GetPrimaryKeyField() (*Field, error) {
...
+// any unique sdkField, per operation. Which means that we will never set
+// two different sdk field from the same
+func (r *CRD) GetSDKFieldSetter(opType OpType, sdkField string) string {
```
jaypipes, line 421: Recommend renaming this to `GetMatchingInputShapeFieldName()`

pkg/config/field.go
A-Hilaly: I think it's just Github preview.. i see proper indentation from here:
Screenshot 2023-05-02 at 11 22 51

jaypipes (changes requested):
Good stuff. Some minor nits inline but nothing major. Thanks @A-Hilaly!

pkg/config/field.go
A-Hilaly: Sniper eyes!

jaypipes (approved):
muy bueno.

PR #442, opened 2023-05-02
only call SetAttributes() when attrMap not empty

jaypipes:
For SQS's `CreateQueue` API, if you pass an empty map to the `CreateQueueInput.SetAttributes()` method, everything succeeds however the service will return the following cryptic error message:

`MalformedInput: End of list found where not expected`

This patch updates the generated code for attribute-based APIs to only call `SetAttributes()` on the input shape when there are attributes to set.

Issue aws-controllers-k8s/community#1695


A-Hilaly (approved):
Looks good!

A-Hilaly:
Looks like `s3-controller-test` is flaky 
/retest

PR #454, opened 2023-06-22
Add `custom find_operation` support

nnbu:
Issue: aws-controllers-k8s/community#1827

Description of changes:
This change allow to implement custom find function. The name of the function could be written in generator.yaml as 'find_operation: <name-of-function>' similar to update_operation. And generated code for sdkFind() would call this custom find function. The find function itself could be written in hooks.go.

### DIFF
--- a/pkg/config/resource.go
+++ b/pkg/config/resource.go
@@ -63,6 +63,18 @@ type ResourceConfig struct {
 	// very little consistency to the APIs that we can use to instruct the code
 	// generator :(
 	UpdateOperation *UpdateOperationConfig `json:"update_operation,omitempty"`
+	// ReadOperation contains instructions for the code generator to generate
+	// Go code for the read operation for the resource. For some resources,
+	// there is no describe/find/list apis. However, it is possible to write
+	// `read` function for such resources using another resource's (say,
+	// resource B) `read` function. Then, one could write custom code to find
+	// the relevant resource from the response of resource B's read function.
+	//
+	// Additionally, if the `read` function for a resource is not available
+	// then reconciler's progression is stalled with the error `read`
+	// operation `notImplemented`. Custom read function allows the
+	// reconciler's progression.
+	ReadOperation *ReadOperationsConfig `json:"find_operation,omitempty"`
 	// Reconcile describes options for controlling the reconciliation
 	// logic for a particular resource.
 	Reconcile *ReconcileConfig `json:"reconcile,omitempty"`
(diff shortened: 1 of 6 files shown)


pkg/config/resource.go
```diff
@@ -62,7 +62,19 @@ type ResourceConfig struct {
...
+	// there is no describe/find/list apis. However, it is possible to write
+	// `find` function for such resources using another resource's (say,
+	// resource B) `find` function. Then, one could write custom code to find
+	// the relevant resource from the response of resource B's find function.
+	//
+	// Additionally, if there the `find` function for a resource is not
+	// available then reconciler's progression is stalled with the error
+	// `find` operation `notImplemented`. Custom find function allows the as
+	// reconciler's progression.
+	FindOperation *CustomOperationsConfig `json:"find_operation,omitempty"`
```
A-Hilaly, line 77: To keep consistent naming across all the packages, let's name this `ReadOperation` or `ReadOneOperation` ?

pkg/config/resource.go
```diff
@@ -312,12 +324,12 @@ type ListOperationConfig struct {
 	MatchFields []string `json:"match_fields"`
 }
 
-// UpdateOperationConfig contains instructions for the code generator to handle
-// Update operations for service APIs that have resources that have
-// difficult-to-standardize update operations.
-type UpdateOperationConfig struct {
```
A-Hilaly, line 318: Since `omitUnchangedFields` doesn't apply to read operations, we can create a new struct called `ReadOperationConfig`

pkg/generate/code/set_resource.go
```diff
@@ -962,6 +962,8 @@ func SetResourceIdentifiers(
 		case r.Ops.ReadMany != nil:
 			// If single lookups can only be done using ReadMany
 			op = r.Ops.ReadMany
+		default:
+			return ""
```
A-Hilaly, line 966: Not sure why is this needed, it could also break other functionalities

pkg/config/resource.go
nnbu: done

pkg/config/resource.go
nnbu: done

pkg/generate/code/set_resource.go
nnbu: `If op == nil` on line 957 is true, then we get into this `switch` statement. However, if both cases `r.Ops.GetAttributes` and `r.Ops.ReadMany` are `nil`, then `op` remains null. and then the code generator crashes on line 969 when it dereferences `op` as `op.InputRef.Shape`.

This crash happened when I was testing `networkAclEntry` which does not have `read` operation and `op` remained `nil`. 

So fixing this already existing bug.

a-hilaly (approved):
Thank you @nnbu !
/lgtm
/test all

PR #455, opened 2023-07-17
Add leader elections flags/values to the deployment manifests

a-hilaly:
Issue: https://github.com/aws-controllers-k8s/community/issues/1753#issuecomment-1635030176

ACK controllers use k8s-sigs/controller-runtime behind the scenes, which
support leader election. This feature is not properly working due to a
missing configuration `LeaderElectionNamespace` which is used by the
manager to create `k8s.io/coordination` Lease objects.

This patch sets the default `LeaderElectionNamespace` to `ack-system`
and adds the capability of enabling leader election using helm values.

Co-authored-by: Adam Cornett <adc@redhat.com>


templates/cmd/controller/main.go.tpl
```diff
@@ -87,13 +91,14 @@ func main() {
...
-		LeaderElectionID:   awsServiceAPIGroup,
-		Namespace:		    ackCfg.WatchNamespace,
+		Scheme:                  scheme,
+		Port:                    port,
+		Host:                    host,
+		MetricsBindAddress:      ackCfg.MetricsAddr,
+		LeaderElection:          ackCfg.EnableLeaderElection,
+		LeaderElectionID:        awsServiceAPIGroup,
+		Namespace:               ackCfg.WatchNamespace,
+		LeaderElectionNamespace: ackSystemNamespace,
```
A-Hilaly, line 101: I was thinking about adding a `--leader-election-namespace` flag to allow users to set their own.. thoughts?

A-Hilaly:
Example logs of a controller acquiring leader lease:
```
I0717 22:07:29.267656  313484 leaderelection.go:248] attempting to acquire leader lease ack-system/lambda.services.k8s.aws...
I0717 22:07:29.282064  313484 leaderelection.go:258] successfully acquired lease ack-system/lambda.services.k8s.aws
2023-07-17T22:07:29.282+0100	DEBUG	events	ma_ca5abb0a-74fe-45a9-85e1-3b0c3a95e9ac became leader	{"type": "Normal", "object": {"kind":"Lease","namespace":"ack-system","name":"lambda.services.k8s.aws","uid":"a68f0032-2021-436a-b4a2-91fb44f297e9","apiVersion":"coordination.k8s.io/v1","resourceVersion":"383697"}, "reason": "LeaderElection"}
```

Example logs of a controller waiting for leader lease to be freed:
```
I0717 22:10:32.402053  314991 leaderelection.go:248] attempting to acquire leader lease ack-system/lambda.services.k8s.aws...
```

acornett21 (changes requested):
I feel like this change should be duplicated to the kuztomize templates as well.

templates/cmd/controller/main.go.tpl
acornett21: I'm not the sure what the use case for this flag is, or what it actually does under the covers. `operator-sdk` only scaffolds with the below options:

```
        Scheme:                 scheme,
        MetricsBindAddress:     metricsAddr,
        Port:                   9443,
        HealthProbeBindAddress: probeAddr,
        LeaderElection:         enableLeaderElection,
        LeaderElectionID:       leaderElectionID,
```

A-Hilaly: `LeaderElectionNamespace` is the namespace where the `Lease` object is created

A-Hilaly: Without namespace, i'm running into `unable to create controller manager	{"aws.service": "lambda", "error": "unable to find leader election namespace: not running in-cluster, please specify LeaderElectionNamespace"}` - does operator-sdk use leases as well? or maybe configmaps (that's what controller-runtime used before leases introduction)?

templates/config/controller/deployment.yaml.tpl
```diff
@@ -37,6 +37,10 @@ spec:
         - "$(ACK_RESOURCE_TAGS)"
         - --watch-namespace
         - "$(ACK_WATCH_NAMESPACE)"
+        - --enable-leader-election
+        - "$(ENABLE_LEADER_ELECTION)"
```
acornett21, line 41: These 2 lines need to be combined, `booleans` need to be on the same line, since the presence of the flag is a truthy statement, and nothing else gets parsed after that. This is coming from: 

```
Package pflag is a drop-in replacement for Go's flag package, implementing
POSIX/GNU-style --flags.

pflag is compatible with the GNU extensions to the POSIX recommendations
for command-line options. See
http://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html
```

Our existing boolean flags are also incorrect, I'll raise a PR.
```suggestion
        - --enable-leader-election=$(ENABLE_LEADER_ELECTION)
```

a-hilaly: Addressed

templates/config/controller/deployment.yaml.tpl
```diff
@@ -66,6 +70,10 @@ spec:
           value: "info"
         - name: ACK_RESOURCE_TAGS
           value: "services.k8s.aws/controller-version=%CONTROLLER_SERVICE%-%CONTROLLER_VERSION%,services.k8s.aws/namespace=%K8S_NAMESPACE%"
+        - name: ENABLED_LEADER_ELECTION
```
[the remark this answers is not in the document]
a-hilaly: Addressed

templates/config/rbac/leader_election_role.yaml.tpl
```diff
@@ -0,0 +1,37 @@
...
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
+  - delete
```
acornett21, line 18: I have confirmed that ACK does not need the `configmap`. This is sort of an either/or thing, where `configmap` or `leases` can be used, and the operator developer can choose which to use. For `legacy` purposes and for operators that might have been developed before `leases` were created, this is kept in the scaffolding.

Since ACK isn't exposing this [flag](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.0/pkg/manager/manager.go#L144-L175) which allows the choice, we are safe to remove the `confimap` section from the RBAC. 

@a-hilaly I hope all the above makes sense, let me know if there are questions...After this is removed I think we are good to go.

a-hilaly:
/hold

need to pin runtime `v0.27.1`

acornett21:
@a-hilaly Why did you revert the RBAC changes?

a-hilaly:
>@a-hilaly Why did you revert the RBAC changes?

Looks like it's gonna be a bigger change than we expected.. since the feature is disabled by default, it's safe to go with this now and we will address the RBAC matters in a seperate PR

acornett21:
> since the feature is disabled by default, it's safe to go with this now and we will address the RBAC matters in a seperate PR

@a-hilaly Can you elaborate on this? If someone goes to enable this, the controller will fail to function and throw errors without RBAC changes present.

a-hilaly:
@acornett21 good points, and thank you for bringing this! I just added few configurations for kustomize and helm chart templates.

templates/cmd/controller/main.go.tpl
```diff
@@ -87,13 +87,14 @@ func main() {
...
-		MetricsBindAddress: ackCfg.MetricsAddr,
-		LeaderElection:		ackCfg.EnableLeaderElection,
-		LeaderElectionID:   awsServiceAPIGroup,
-		Namespace:		    ackCfg.WatchNamespace,
+		Scheme:                  scheme,
+		Port:                    port,
+		Host:                    host,
+		MetricsBindAddress:      ackCfg.MetricsAddr,
+		LeaderElection:          ackCfg.EnableLeaderElection,
+		LeaderElectionID:        awsServiceAPIGroup,
```
RedbackThomson, line 95: Do we want to prefix this with `ack-` just in case?

templates/config/rbac/leader-election-role-binding.yaml.tpl
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  namespace: ack-system
```
RedbackThomson, line 5: Use `namespace: {{ .Release.Namespace }}`. We don't know that the controller will be installed into `ack-system`.

templates/config/rbac/leader-election-role.yaml.tpl
```diff
@@ -0,0 +1,26 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{.ServicePackageName}}-leader-election-role
+  namespace: ack-system
```
RedbackThomson, line 6: Use `namespace: {{ .Release.Namespace }}`

templates/config/rbac/leader-election-role-binding.yaml.tpl
```diff
@@ -0,0 +1,14 @@
...
+  name: {{.ServicePackageName}}-leader-election-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{.ServicePackageName}}-leader-election-role
+subjects:
+- kind: ServiceAccount
+  name: {{.ServiceAccountName}}
+  namespace: ack-system
```
RedbackThomson, line 14: Use `namespace: {{ .Release.Namespace }}`

templates/helm/values.yaml.tpl
```diff
@@ -121,3 +121,11 @@ serviceAccount:
   name: {{ .ServiceAccountName }}
   annotations: {}
     # eks.amazonaws.com/role-arn: arn:aws:iam::AWS_ACCOUNT_ID:role/IAM_ROLE_NAME
+
+# leader election configurations
```
RedbackThomson, line 125: Please use proper casing for this - this is documentation. Maybe you could link to the K8s documentation about leader election, in case people wonder what it is

RedbackThomson (changes requested):

templates/config/rbac/leader-election-role-binding.yaml.tpl
a-hilaly: This is not a helm configuration, it's the plain manifests

templates/helm/values.yaml.tpl
a-hilaly: There is no official documentation on this topic.. it's more a controller-runtime thing. Maybe the Go docs?

templates/config/rbac/leader-election-role-binding.yaml.tpl
RedbackThomson: Ah I missed that. Nvm

templates/helm/values.yaml.tpl
RedbackThomson: I did find this section about leader election on the K8s page for leases - https://kubernetes.io/docs/concepts/architecture/leases/#leader-election

But it's vague.

RedbackThomson (approved):
/lgtm

RedbackThomson (commented):
/lgtm

PR #88, opened 2021-06-10
Remove unnecessary homedir module

RedbackThomson:
`os.UserHomeDir` is supported in go1.12+. We no longer need a third party module to load the home directory.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #462, opened 2023-08-23
Support of injecting fields in existing structures: CustomNestedFields

Vandita2020:
**Issue #**
This issue arises from the code changes in lambda-controller for[ PR #88](https://github.com/aws-controllers-k8s/lambda-controller/pull/88)

**Summary**
Support of injecting fields in existing structures: CustomNestedFields

**Description**
The code in this PR helps in accepting a custom nestedField defined using `type` as a part of its parentField. In the issue mentioned above when we inject a new field `S3SHA256` into `Code` struct, instead of nesting it under `Code` the generator created a separate field named `CodeS3SHA256`. 

The new field `S3SHA256` is defined like this in `generator.yaml` file:
```
Function:
    fields:
      Code.S3SHA256:
        type: string
```

After applying the changes in this PR, the field `S3SHA256` is properly injected as a part of `Code` struct.

```
type FunctionCode struct {
	ImageURI *string `json:"imageURI,omitempty"`
	S3Bucket *string `json:"s3Bucket,omitempty"`
	S3BucketRef     *ackv1alpha1.AWSResourceReferenceWrapper `json:"s3BucketRef,omitempty"`
	S3Key           *string                                  `json:"s3Key,omitempty"`
	S3ObjectVersion *string                                  `json:"s3ObjectVersion,omitempty"`
	S3SHA256        *string                                  `json:"s3SHA256,omitempty"`
	ZipFile         []byte                                   `json:"zipFile,omitempty"`
} 
```

This PR contains code implementation to inject Nested fields into the Spec or Status struct and has respective model-test to validate the changes.

**Acknowledgment**


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "go.buildTags": "codegen",
+}
```
a-hilaly, line 3: Do not commit this please. Maybe let's add `.vscode/settings.json` to `.gitignore` file ?

pkg/model/model.go
```diff
@@ -76,6 +81,32 @@ func (m *Model) crdNames() []string {
...
+func CheckIfFieldExists(
+	field map[string]*Field,
+	currValue string,
+) bool {
+	_, ok := field[currValue]
+	if ok {
+		return true
+	}
+	return false
+}
```
a-hilaly, line 93: Mmmmm i think this is unecessary.. it will always fall down to:
```go
                  if _, ok := fieldsInParent[parent]; ok {
						// Check if the field is of type structure
						previous, res = CheckType(value, fieldsInParent)
						if res != true {
							fmt.Printf("Cannot add new field here, %s is not of type Structure", parentName)
							break
						}
					}
```

pkg/model/model.go
```diff
@@ -76,6 +81,32 @@ func (m *Model) crdNames() []string {
...
+
+func CheckType(
```
a-hilaly, line 95: Same comment here, i don't think we need this function.. or the least it should be named `isShapeStructure` .. and still not very convinced of moving this logic to a function

pkg/model/model.go
```diff
@@ -49,6 +49,11 @@ type Model struct {
 	docCfg *ackgenconfig.DocumentationConfig
 }
 
+type Nested struct {
+	arr    []string
+	Config *ackgenconfig.FieldConfig
+}
```
a-hilaly, line 55: i know this is working. Pretty sure we can find a better way to represent this data.. (i think we are copying a lot `arr`)

pkg/model/model.go
```diff
@@ -296,8 +344,61 @@ func (m *Model) GetCRDs() ([]*CRD, error) {
 		// Now add the additional printer columns that have been defined explicitly
 		// in additional_columns
 		crd.addAdditionalPrinterColumns(m.cfg.GetAdditionalColumns(crdName))
-
 		crds = append(crds, crd)
+
+		for _, toInjectNestedFields := range alltoInjectNestedFields {
+
+			addThis := ""
```
a-hilaly, line 351: need a better name than `addThis`

pkg/model/model.go
```diff
@@ -296,8 +344,61 @@ func (m *Model) GetCRDs() ([]*CRD, error) {
...
+			var previous *Field
+			var res bool
+
+			len := len(toInjectNestedFields.arr)
```
a-hilaly, line 355: `len` is a Go keyword, i suggest just to use `len(toInjectNestedFields.arr)` whenever you need to.

pkg/model/model_lambda_test.go
```diff
@@ -103,3 +103,29 @@ func TestLambda_Function(t *testing.T) {
...
+
+	crd := getCRDByName("Function", crds)
+	require.NotNil(crd)
+
+	assert.Contains(crd.SpecFields, "Code")
+	codeField := crd.SpecFields["Code"]
+
+	// Check if Nested Field is inside its Parent Field
+	assert.Contains(codeField.MemberFields, "S3SHA256")
+	assert.Contains(codeField.ShapeRef.Shape.MemberRefs, "S3SHA256")
```
a-hilaly, line 130: Good testing!

pkg/model/model.go
```diff
@@ -296,8 +344,61 @@ func (m *Model) GetCRDs() ([]*CRD, error) {
...
+			// To add newField in its parentField's MemeberFields
+			fPath := addThis
+			memberNames := names.New(addThis)
+			typeOverride := *toInjectNestedFields.Config.Type
+			shapeRef := m.SDKAPI.GetShapeRefFromType(typeOverride)
+			fConfig := crd.cfg.GetFieldConfigByPath(crd.Names.Original, fPath)
+			f := NewField(crd, fPath, memberNames, shapeRef, fConfig)
+			crd.Fields[fPath] = f
+
+			previous.MemberFields[addThis] = f
```
a-hilaly, line 396: :+1:

pkg/model/model.go
```diff
@@ -296,8 +344,61 @@ func (m *Model) GetCRDs() ([]*CRD, error) {
...
+			addThis := ""
+			var previous *Field
+			var res bool
+
+			len := len(toInjectNestedFields.arr)
+
+			for n, value := range toInjectNestedFields.arr {
+				if n == 0 {
+					// first element is a field of CRD
+					fieldsInCRD := crd.SpecFields
```
a-hilaly, line 360: This could have been declared outside, it's a non necessary operation to do at every loop (same for few variables below)

pkg/model/model.go
```diff
@@ -296,8 +344,61 @@ func (m *Model) GetCRDs() ([]*CRD, error) {
...
+					parentName := crd.Names.Camel
+					if CheckIfFieldExists(fieldsInCRD, value) {
+						// Check if the field is of type structure
+						previous, res = CheckType(value, fieldsInCRD)
+						if res != true {
```
a-hilaly, line 365:
```suggestion
						if !res {
```

a-hilaly (commented):
@Vandita2020 This is great work, well done!
I left a few comments to improve the code and make it more similar to what we have in the other packages.

pkg/model/model.go
Vandita2020: Have removed the function and added the logic to the calling function

pkg/model/model.go
Vandita2020: I didn't actually get what you mean by copying `arr` data a lot. It is just defined once to store the names of nested fields and is then iterated over to check the nested fields type and whether they are a part of their parent field. But yeah I'm open to looking for a better way to represent the data.

pkg/model/model.go
Vandita2020: Changed the name to `newCustomField`. Let me know if that sounds okay.

pkg/model/model.go
Vandita2020: This whole part is removed now

pkg/model/model.go
Vandita2020: There is a relation between `fieldsInParent` and `previous` variable, which changes with every loop. I can declare the `fieldsInCRD` outside the loop, but that is the only variable that can be defined outside the loop. To maintain the code readability I declared it the same way as `fieldsInParent`.

.vscode/settings.json
RedbackThomson: I'd even recommend adding `.vscode` to a global gitignore on your system. Otherwise every repo needs to add every IDE's config paths (`.vscode`, `.idea`, etc.) individually.

Vandita2020: Done, thanks!

pkg/model/crd.go
```diff
@@ -626,6 +626,94 @@ func (r *CRD) addAdditionalPrinterColumns(additionalColumns []*ackgenconfig.Addi
...
+}
+
+// addCustomNestedFields injects the provided customNestedFields into the
+// Spec or Status struct as a nested field. The customNestedFields are
+// identified by the field path. The field path is a dot separated string
+// that represents the path to the nested field. For example, if we want to
+// inject a field called "Password" into the "User" struct, the field path
+// would be "User.Password". The field path can be as deep as needed.
+func (crd *CRD) addCustomNestedFields(customNestedFields map[string]*ackgenconfig.FieldConfig) {
+	// Now that we collected all the nested fields, we can process, validate
```
valerena, line 652: What does "now" mean here? The message doesn't seem relevant in the beginning of a function.

pkg/model/crd.go
```diff
@@ -626,6 +626,94 @@ func (r *CRD) addAdditionalPrinterColumns(additionalColumns []*ackgenconfig.Addi
...
+	// that they indeed are inject-able (i.e. the parent field is of type struct)
+	// and inject them into the Spec or Status struct.
+	for customNestedField, customNestedFieldConfig := range customNestedFields {
+		fieldParts := strings.Split(customNestedField, ".")
+		// we now than the length of fieldParts is at least 2
+		// it is safe to access the firs element.
+		topLevelField := fieldParts[0]
+
+		f, ok := crd.checkSpecOrStatus(topLevelField)
+		// f, ok := crd.SpecFields[topLevelField]
```
valerena: Remove commented out code

pkg/model/crd.go
```diff
@@ -626,6 +626,94 @@ func (r *CRD) addAdditionalPrinterColumns(additionalColumns []*ackgenconfig.Addi
...
+// that represents the path to the nested field. For example, if we want to
+// inject a field called "Password" into the "User" struct, the field path
+// would be "User.Password". The field path can be as deep as needed.
+func (crd *CRD) addCustomNestedFields(customNestedFields map[string]*ackgenconfig.FieldConfig) {
+	// Now that we collected all the nested fields, we can process, validate
+	// that they indeed are inject-able (i.e. the parent field is of type struct)
+	// and inject them into the Spec or Status struct.
+	for customNestedField, customNestedFieldConfig := range customNestedFields {
+		fieldParts := strings.Split(customNestedField, ".")
+		// we now than the length of fieldParts is at least 2
```
valerena, line 657: we *know that* the length

pkg/model/crd.go
```diff
@@ -626,6 +626,94 @@ func (r *CRD) addAdditionalPrinterColumns(additionalColumns []*ackgenconfig.Addi
...
+			panic(msg)
+		}
+		// else {
+		// 	f, ok = crd.SpecFields[topLevelField]
+		// 	if ok && f.ShapeRef.Shape.Type != "structure" {
+		// 		// We need to panic here because the user is providing wrong configuration.
+		// 		msg := fmt.Sprintf("Expected parent field to be of type structure, but found %s", f.ShapeRef.Shape.Type)
+		// 		panic(msg)
+		// 	}
+		// }
```
valerena, line 675: Remove commented code

pkg/model/model.go
```diff
@@ -136,8 +136,16 @@ func (m *Model) GetCRDs() ([]*CRD, error) {
 			crd.AddSpecField(memberNames, memberShapeRef)
 		}
 
-		// Now any additional Spec fields that are required from other API
-		// operations.
+		// A list of fields that should be processed after building the gathering
+		// the Spec and Status top level fields. The customNestedFields will be
```
valerena, line 140: the phrase "after building the gathering the Spec and Status" feels wrong. I think there's an extra word that shouldn't be there.

a-hilaly (approved):
Very nice! Thanks a lot for your contribution @Vandita2020 . This will be beneficial to all the ACK controllers and it's community.
/approve
/lgtm

PR #463, opened 2023-08-31
Upgrade ACK controllers runtime image

a-hilaly:
Upgrade `eks-distro-build-tooling/eks-distro-minimal-base-nonroot` image to tag `2023-08-24-1692903666.2023`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Base image to use at runtime
-ARG base_image=public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-nonroot:2021-12-01-1638322424
+ARG base_image=public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-nonroot:2023-08-24-1692903666.2023
 
 # Golang image to use for compiling the manager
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Base image to use at runtime
-ARG base_image=public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-nonroot:2021-12-01-1638322424
+ARG base_image=public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-nonroot:2023-08-24-1692903666.2023
```
a-hilaly, line 2: I'm thinking that maybe we should ping this to `public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-nonroot:latest-al23` to make sure that we always build with the latest image? Thoughts?

jljaco: My $0.02: I don't think it's wise to pin this to a moving target like `latest-al23`.  Too much risk of introducing breaking changes from upstream.

jljaco (approved):

PR #464, opened 2023-09-05
Functionality to make `ignore` work

Vandita2020:
**Issue** 
This PR is related to fix the issue generating after executing PR #462 

**Summary**
To disregard fields marked with `set: ignore` for nested fields. This PR also enhances the functionality of `set:ignore` to support the exclusion of fields from `set_sdk`.

**Description**

The PR  fixes the issue coming after a new custom field is created and has `set.ignore` set to `true` for it. Even after setting the field to ignore, the field is getting used to set the SDK fields. In this case particularly, the new field was defined as shown below:

```
Code.S3SHA256:
        type: string
        set:  
          - method: Create
            ignore: true
```
And the issue it is causing is under `sdk.go/newCreateRequestPayload` function, where it is being set, as shown below:
```
if r.ko.Spec.Code.S3SHA256 != nil {
	f1.SetS3SHA256(*r.ko.Spec.Code.S3SHA256)
}
```
The goal is to ignore this particular line of code. And also to improve the basic functionality of `set.ignore`. 

The `set.ignore` functionality was till now supporting exclusion of fields only when the field was used to set for the resource. 

This PR checks the `set.ignore` for field and ignores the following code that creates request payload if it is set to `true`. 

**Limitations**
Currently this PR checks the `set.ignore` field only for members of the field whose type is `structure`. 

**Acknowledgment**

### DIFF
--- a/pkg/config/field.go
+++ b/pkg/config/field.go
@@ -14,6 +14,7 @@
 package config
 
 import (
+	"encoding/json"
 	"strings"
 )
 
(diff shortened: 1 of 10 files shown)


a-hilaly (commented):
Thank you @Vandita2020 - looks like this PR have some changes that we reviewd in #462 . Can you please delete the first 4 commits in this branch?
/hold

a-hilaly (approved):
Great stuff~
/lgtm

PR #468, opened 2023-09-14
(fix) enable custom names for service accounts

grandwizard28:
Issue N/A

Allow leaderElection role to be bound to a custom `ServiceAccount` name, similarily to the main cluster role.

### DIFF
--- a/templates/helm/templates/leader-election-role-binding.yaml.tpl
+++ b/templates/helm/templates/leader-election-role-binding.yaml.tpl
@@ -14,6 +14,6 @@ roleRef:
   name: {{.ServicePackageName}}-leader-election-role
 subjects:
 - kind: ServiceAccount
-  name: {{ include "service-account.name" . }}
+  name: {{ "{{ include "service-account.name" . }}" }}
   namespace: {{ "{{ .Release.Namespace }}" }}
 {{- "{{- end }}" }}
--- a/templates/helm/templates/leader-election-role-binding.yaml.tpl
+++ b/templates/helm/templates/leader-election-role-binding.yaml.tpl
@@ -14,6 +14,6 @@ roleRef:
   name: {{.ServicePackageName}}-leader-election-role
 subjects:
 - kind: ServiceAccount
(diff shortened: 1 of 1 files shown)


a-hilaly (approved):
Thank you!
/lgtm

templates/helm/templates/leader-election-role-binding.yaml.tpl
```diff
@@ -14,6 +14,6 @@ roleRef:
   name: {{.ServicePackageName}}-leader-election-role
 subjects:
 - kind: ServiceAccount
-  name: {{.ServiceAccountName}}
+  name: {{ include "service-account.name" . }}
```
a-hilaly, line 17: @grandwizard28 since this is template generating helm templates we'll have to do something like
```
{{ "{{ include "service-account.name" . }}" }}
```

grandwizard28: Oops, my bad!

templates/helm/templates/leader-election-role-binding.yaml.tpl
```diff
@@ -14,6 +14,6 @@ roleRef:
   name: {{.ServicePackageName}}-leader-election-role
 subjects:
 - kind: ServiceAccount
-  name: {{.ServiceAccountName}}
+  name: {{ "{{ include "service-account.name" . }}" }}
```
a-hilaly, line 17: Looks like we need to prepend an `\` for the helm brackets. I still get confused with this templinception..
```suggestion
  name: {{ "{{ include \"service-account.name\" . }}" }}
```

vibhu-mudrex: Hahaha, what a coincidence, I just pushed the same patch!

a-hilaly:
tests decided to be flaky today...
 /test s3-controller-test

a-hilaly (approved):
Thank you @grandwizard28 @vibhu-mudrex ! 👍 
/test all
/lgtm

PR #472, opened 2023-10-04
Adding functionality to assign one API to multiple resources

Vandita2020:
Fixes [#1917](https://github.com/aws-controllers-k8s/community/issues/1917)

**Description:**
The PR adds functionality to use same API for an operation on multiple resources. This allows `ResourceName` to be a list so that same API for an `OperationType` can map to multiple resources. This is required in certain cases where the same API, for same operation is used for multiple resources.

The solution this PR applies is to pass multiple resources for same API.

This supports following cases cases:

1. One API -> One Resource
```
PublishVersion:
    operation_type:
      - Create
    resource_name: 
      - Version
```

2. One API -> Multiple Resource
```
DeleteFunction:
    operation_type:
      - Delete
    resource_name:
      - Function
      - Version
```      

### DIFF
--- a/pkg/model/sdk_api.go
+++ b/pkg/model/sdk_api.go
@@ -78,12 +78,14 @@ func (a *SDKAPI) GetOperationMap(cfg *ackgenconfig.Config) *OperationMap {
 	// create an index of Operations by operation types and resource name
 	opMap := OperationMap{}
 	for opID, op := range a.API.Operations {
-		opTypeArray, resName := getOpTypeAndResourceName(opID, cfg)
-		for _, opType := range opTypeArray {
+		opTypes, opResourceNames := getOpTypesAndResourcesMapping(opID, cfg)
+		for _, opType := range opTypes {
 			if _, found := opMap[opType]; !found {
 				opMap[opType] = map[string]*awssdkmodel.Operation{}
 			}
-			opMap[opType][resName] = op
+			for _, resName := range opResourceNames {
+				opMap[opType][resName] = op
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


Vandita2020:
do-not-merge/work-in-progress

pkg/model/sdk_api.go
```diff
@@ -98,16 +100,18 @@ func (a *SDKAPI) GetOperationMap(cfg *ackgenconfig.Config) *OperationMap {
 	//
 	// see: https://github.com/aws-controllers-k8s/community/issues/1555
 	for opID, opCfg := range cfg.Operations {
-		if opCfg.ResourceName == "" {
+		if opCfg.ResourceName == nil {
```
a-hilaly, line 103: `opCfg.Resource` is of type `string`, you can't compare to a nil pointer

pkg/model/sdk_api.go
```diff
@@ -360,20 +364,23 @@ func NewSDKAPI(api *awssdkmodel.API, apiGroupSuffix string) *SDKAPI {
...
-		for _, operationType := range operationConfig.OperationType {
-			opType = OpTypeFromString(operationType)
-			opTypes = append(opTypes, opType)
-		}
+	// The existance of the operation in the config file is not enough to
+	// override the operation type and/or resource name. The operation type
+	// and/or resource name must be specified in the config file.
+	if !exists || len(opConfig.ResourceName) == 0 || len(opConfig.OperationType) == 0 {
+		return []OpType{opType}, []string{resName}
+	}
```
a-hilaly, line 376: @Vandita2020 You approach is correct, we only missed this last bits of code ^. When a user provides an Operation configuration without `ResourceName` nor `OperationType` directives - this block doesn't get called. Causing the opTypes list to reset (L378)

PR #483, opened 2023-12-31
Add support for Multi Namespace watch mode

a-hilaly:
This commit leverages new functionalities recently introduces in the k8s
upstream `controller-runtime`, enabling users to configure the controller
for multi-namespace watch mode.

Now users can specify multiple namespaces using a comma-separated list,
such as `default,namespace1,namespace2` (either using helm values or
binary flags (for the bravest, running ACK controllers using k8s
Deployments)


templates/cmd/controller/main.go.tpl
```diff
@@ -86,14 +92,30 @@ func main() {
...
 	mgr, err := ctrlrt.NewManager(ctrlrt.GetConfigOrDie(), ctrlrt.Options{
-		Scheme:                  scheme,
-		Port:                    port,
-		Host:                    host,
-		MetricsBindAddress:      ackCfg.MetricsAddr,
+		Scheme: scheme,
+		Cache: ctrlrtcache.Options{
+			Scheme:            scheme,
+			DefaultNamespaces: defaultNamesSpaces,
+		},
```
acornett21, line 109: I'm not sure if this is the ideal approach. I'm not certain but I think we'd want to use `cache.MultiNamespacedCacheBuilder` in a more dynamic way. See example [here](https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#configuring-cluster-scoped-operators-with-multinamespacedcachebuilder)

acornett21:
If we start supporting multi namespace we should also add in `OwnNamespace` `SingleNamespace` and `MultiNamespace` modes in the CSV. Code that would need to be updated is [here](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/olm/olm.go#L137-L142)

templates/cmd/controller/main.go.tpl
a-hilaly: That was possible in `0.15.3` which is the version sdk-operator still using, however it was deprecated in 0.16.0 (https://github.com/kubernetes-sigs/controller-runtime/pull/2423) and a newer granular configurations were possible with:
- https://github.com/kubernetes-sigs/controller-runtime/pull/2421/files
- https://github.com/kubernetes-sigs/controller-runtime/pull/2539/files

acornett21: AHH okay cool. I'm still not up to speed on `0.16.x`

acornett21 (approved):

PR #485, opened 2024-01-10
Introducing a generator configuration for GoTag customization

a-hilaly:
This patch introduces a new configuration, `go_tag`, to allow ACK
developers to customize the "Go tags" in the generated structures.
With this configuration users can easily override the default Go tags,
making adjustements like changing the tag name, adding omitempty
directives or even introducing new ones (yaml, gorm, protobuf etc).

This change helps ACK users overcome the challenge imposed by our names
utility package (`aws-controllers-k8s/pkg/names/`) which adds an
underscore suffix to field names that matches Golang keywords. Resulting
in an odd UX when consuming the generated ACK  CRDs (e.g using
`type_: <...>` instead of `type: <...>`)

We purposefully avoided directly modifying the `names` utility package
(trimming the underscores when generating the go tags) as it will break
compatibility with the already existing CRDs (yep we missed this).
Without some sort of mutating webhooks that will handle field renames,
we'll avoid going that route, and instead start using `go_tag` configs
for future CRDs/fields.

This patch will help in tweaking the generated code for
https://github.com/aws-controllers-k8s/eks-controller/pull/84

### DIFF
--- a/pkg/config/field.go
+++ b/pkg/config/field.go
@@ -405,6 +405,30 @@ type FieldConfig struct {
 	// TODO(jaypipes,crtbry): Figure out if we can roll the CustomShape stuff
 	// into this type override...
 	Type *string `json:"type,omitempty"`
+	// GoTag is used to override the default Go tag injected into the fields of
+	// a generated go structure. This is useful if we want to override the json
+	// tag name or add an omitempty directive to the tag. If not specified,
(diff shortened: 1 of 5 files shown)


jonathan-innis (approved):
/lgtm

a-hilaly:
experiencing some etcd timeouts with the ec2-controller, restarting to see if it's consistent
/test ec2-controller-test

PR #494, opened 2024-01-20
Experimental configurations to override controller names

a-hilaly:
Since the early days of ACK, we've always relied on the `serviceID` to
name the controller repositories. The way names, go modules, import
paths, helm charts are named relied on this `serviceID`. Which was the
main reason why some controllers had a bit of odd names, like
`prometheusservice-controller` instead of `amp-controller`,
`acmpca-controller` instead of `pca-controller`. This patch introduces
experimental configurations that instructs the code generator to
override the serviceID/servicePackageName when it comes to import
paths/helm charts names etc...

Related issue: https://github.com/aws-controllers-k8s/community/issues/1411

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -44,6 +44,14 @@ type Config struct {
 	// only the service model name, but also the iface name to `PipesAPI`. See
 	// https://github.com/aws/aws-sdk-go/blob/main/service/pipes/pipesiface/interface.go#L62
 	SDKNames SDKNames `json:"sdk_names"`
+	// ControllerName lets you specify a service alias override. This can be
+	// used to only override the controller name exposed to the user. Meaning that
+	// it will only change the module name, the import paths, and the controller
+	// name. This is useful when the service identifier is not very user friendly
+	// and you want to expose a more user friendly name to the user. e.g docdb ->
+	// documentdb.
+	// This will also change the helm chart and image names.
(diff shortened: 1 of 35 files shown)


pkg/config/config.go
```diff
@@ -44,6 +44,14 @@ type Config struct {
...
 	// https://github.com/aws/aws-sdk-go/blob/main/service/pipes/pipesiface/interface.go#L62
 	SDKNames SDKNames `json:"sdk_names"`
+	// ControllerName lets you specify a service alias override. This can be
+	// used to only override the controller name exposed to the user. Meaning that
+	// it will only change the module name, the import paths, and the controller
+	// name. This is useful when the service identifier is not very user friendly
+	// and you want to expose a more user friendly name to the user. e.g docdb ->
+	// documentdb.
+	// This will also change the helm chart and image names.
+	ControllerName string `json:"controller_name,omitempty"`
```
a-hilaly, line 54: So far this is working very well with `efs-controller` and `documentdb-controller`

a-hilaly:
Looks like https://github.com/aws-controllers-k8s/test-infra/pull/422 caught a bug with Prometheus controller :)

a-hilaly:
Tracking documentdb changes in https://github.com/aws-controllers-k8s/documentdb-controller/pull/4

PR #495, opened 2024-01-29
Enhance controller RBAC and helm charts generation

a-hilaly:
This patch enhances the permissions for controllers with the addiotion
of multi-namespace watch mode. The controlelrs now have more
fine-grained permissions, enabling them to watch specific namespaces
exclusively.

Key changes:
- Refactored helm chart generation tooling, especially Roles and
  ClusterRoles
- Addressed a bug where the RoleBinding was incorrectly assigning the
  roles into a different namespace when the controller was installed ina
  namespace that is different fro the one it monitored.

Notable features:
- We start generating one Role/RoleBinding for each monitored namespace.
- We introduced seperate Role and ClusterRoles for the CARM feature
  (prefixed with ack-*-cache)

Less importantly we are replacing some bash-fu that used to inject RBAC
rules into Roles and ClusterRoles, with a different bash-fu that allows
reusing rules for different Roles.


a-hilaly:
Too many hiccups from DynamoDB API... Merging manually as it's not related to this change.

PR #535, opened 2024-08-07
wrong chart name

avielb-navina:
### Issue 
the generator name creates a charge that is called for example "iam-controller" yet the chart name at `Chart.yaml` is iam-chart which is incorrent. 

### Description of changes
the suggested change is aligning this to to match the generated controller name correctly. 

### DIFF
--- a/templates/helm/Chart.yaml.tpl
+++ b/templates/helm/Chart.yaml.tpl
@@ -1,5 +1,5 @@
 apiVersion: v1
-name: {{ .ControllerName }}-chart
+name: {{ .ControllerName }}-controller
 description: A Helm chart for the ACK service controller for {{ .Metadata.Service.FullName }} ({{ .Metadata.Service.ShortName }})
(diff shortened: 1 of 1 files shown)


avielb-navina:
anyone here?

a-hilaly:
thank you for your contribution @avielb-navina - I have a quick question, will this break the experience for users that run a `helm upgrade` for existing charts that they installed before we incorporate this change?

templates/helm/Chart.yaml.tpl
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-name: {{ .ControllerName }}-chart
+name: {{ .ControllerName }}-controller
```
a-hilaly, line 2: i'm wondering if this will break existing controller upgrade process

avielb-navina: can we test it? 
it will change the name of the created chart so i believe that it will. when is the upgrade process described?

PR #538, opened 2024-08-09
Support custom nested fields that have path that contains types other than struct

TiberiuGC:
Issue #, if available:

Description of changes:

Say we want to overwrite a nested field type e.g. for `WAFv2` controller (this is needed due to a cyclic dependency):
```
resources:
  RuleGroup:
      Rules.Statement.AndStatement:
        type: string
```

Currently, the [overwrite code logic](https://github.com/aws-controllers-k8s/code-generator/blob/62466746500d0c01fcfa8e8e3ca6abd9386e4687/pkg/model/crd.go#L660-L687) does not work for the above scenario and it returns the following error:
```
Expected parent field to be of type structure, but found list
```

This happens because `Rules` is not of type `struct`, but `[]struct`, and the referenced code logic requires that all fields within the path (i.e. `Rules` and `Statement` to be of type `struct`).

This PR relaxes the logic to allow fields within the path to also be of types `[]struct` and `map[]struct`.

### DIFF
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -656,17 +656,23 @@ func (crd *CRD) addCustomNestedFields(customNestedFields map[string]*ackgenconfi
 		topLevelField := fieldParts[0]
 
 		f, ok := crd.checkSpecOrStatus(topLevelField)
-
-		if ok && f.ShapeRef.Shape.Type != "structure" {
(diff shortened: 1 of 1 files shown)


a-hilaly (approved):
👍 
/lgtm

PR #549, opened 2024-10-02
Make controller base_image version an argument

michaelhtm:
Description of changes:
This change allows us to pass the base image version as 
an argument allowing us to pass the latest one after each release

### DIFF
--- a/scripts/build-controller-image.sh
+++ b/scripts/build-controller-image.sh
@@ -91,6 +91,7 @@ fi
 pushd "$ROOT_DIR" 1>/dev/null
   # Get the golang version from the code-generator
   GOLANG_VERSION=${GOLANG_VERSION:-"$(go list -f {{.GoVersion}} -m)"}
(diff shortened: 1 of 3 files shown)


a-hilaly (approved):
Nice!!
/lgtm
/retest

scripts/build-controller-image.sh
```diff
@@ -91,6 +91,7 @@ fi
 pushd "$ROOT_DIR" 1>/dev/null
   # Get the golang version from the code-generator
   GOLANG_VERSION=${GOLANG_VERSION:-"$(go list -f {{.GoVersion}} -m)"}
+  BASE_IMAGE_VERSION=${BASE_IMAGE_VERSION:-"$(echo 2024-08-13-1723575672.2)"}
```
a-hilaly, line 94: nit: I think it's preferable to fail if we don't have a BASE_IMAGE_VERSION here. It's better for us to investigate what's wrong than silently building with an old version.

PR #551, opened 2024-10-04
Make controller base_image version an argument

michaelhtm:
Description of changes:
This change allows us to pass the base image version 
as an argument allowing us to pass the latest one after each release

### DIFF
--- a/scripts/build-controller-image.sh
+++ b/scripts/build-controller-image.sh
@@ -5,6 +5,7 @@ set -eo pipefail
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 SCRIPTS_DIR=$DIR
 ROOT_DIR=$DIR/..
+TEST_INFRA_DIR=$ROOT_DIR/../test-infra
 DOCKERFILE_PATH=$ROOT_DIR/Dockerfile
 ACK_DIR=$ROOT_DIR/..
 DOCKERFILE=${DOCKERFILE:-"$DOCKERFILE_PATH"}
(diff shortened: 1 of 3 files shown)


scripts/build-controller-image.sh
```diff
@@ -88,9 +89,17 @@ if ! is_public_ecr_logged_in; then
...
+  GOLANG_VERSION=$(cat build_config.yaml | yq .go_version)
+  BASE_IMAGE_VERSION=$(cat build_config.yaml | yq .eks_distro_version) 
+popd 1>/dev/null
+
+
 pushd "$ROOT_DIR" 1>/dev/null
   # Get the golang version from the code-generator
   GOLANG_VERSION=${GOLANG_VERSION:-"$(go list -f {{.GoVersion}} -m)"}
+  BASE_IMAGE_VERSION=${BASE_IMAGE_VERSION:-"$(echo 2024-08-13-1723575672.2)"}
 popd 1>/dev/null
```
a-hilaly, line 103: We're doing this twice should we get rid of the ROOT_DIR one?

scripts/build-controller-image.sh
```diff
@@ -88,9 +89,17 @@ if ! is_public_ecr_logged_in; then
...
+  # Get the golang version from build_config.yaml
+  GOLANG_VERSION=$(cat build_config.yaml | yq .go_version)
+  BASE_IMAGE_VERSION=$(cat build_config.yaml | yq .eks_distro_version) 
+popd 1>/dev/null
+
+
 pushd "$ROOT_DIR" 1>/dev/null
   # Get the golang version from the code-generator
   GOLANG_VERSION=${GOLANG_VERSION:-"$(go list -f {{.GoVersion}} -m)"}
+  BASE_IMAGE_VERSION=${BASE_IMAGE_VERSION:-"$(echo 2024-08-13-1723575672.2)"}
```
a-hilaly, line 102: I think we should fail if we're missing a GOLANG_VERSION/BASE_IMAGE_VERSION

scripts/build-controller-image.sh
michaelhtm: this is a failsafe, just in case the image is not being built with test-infra in the expected path

a-hilaly (approved):
👍 
/lgtm

PR #557, opened 2024-11-01
fix(helm): Ensure non-default service account name works with namespaces-cache and configmaps-cache role bindings

gunzy83:
Issue #, if available: N/A

Relates to https://github.com/aws-controllers-k8s/kms-controller/pull/83

Description of changes: This change ensures that the required `ClusterRoleBinding` and `RoleBinding` objects apply to a `ServiceAccount` with a custom name.

Error seen with custom service account name:

```
pkg/mod/k8s.io/client-go@v0.30.1/tools/cache/reflector.go:232: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: configmaps is forbidden: User "system:serviceaccount:kms-controller:kms-controller" cannot list resource "configmaps" in API group "" in the namespace "kms-controller"
```

### DIFF
--- a/templates/helm/templates/caches-role-binding.yaml.tpl
+++ b/templates/helm/templates/caches-role-binding.yaml.tpl
@@ -8,7 +8,7 @@ roleRef:
   name: ack-namespaces-cache-{{ .ControllerName }}-controller
 subjects:
 - kind: ServiceAccount
-  name: ack-{{ .ControllerName }}-controller
+  name: {{ IncludeTemplate "service-account.name" }}
   namespace: {{ "{{ .Release.Namespace }}" }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
@@ -22,5 +22,5 @@ roleRef:
   name: ack-configmaps-cache-{{ .ControllerName }}-controller
 subjects:
 - kind: ServiceAccount
-  name: ack-{{ .ControllerName }}-controller
+  name: {{ IncludeTemplate "service-account.name" }}
   namespace: {{ "{{ .Release.Namespace }}" }}


a-hilaly (approved):
Thank you @gunzy83!!
/approve
/assign @michaelhtm

mtougeron:
> @mtougeron @eadasiak any thoughts on this change? i'm wondering how did this work with all the CARM installations we've done before?

I'm not sure about the code that generates these templates, but I'm assuming it's probably that whatever populates `name: {{ .ServiceAccountName }}` here https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/helm/values.yaml.tpl#L141 sets it essentially to the same value by default. So it you didn't change from the ServiceAccount name the chart uses by default it would work fine. But if you did try to change it, it wouldn't work properly.

For example, the [rendered Deployment manifest](https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/helm/templates/deployment.yaml.tpl#L36) uses `serviceAccountName: {{ IncludeTemplate "service-account.name" }}` so if someone sets

```yaml
serviceAccount:
  name: foo
```

in their values.yaml (or via `--set serviceAccount.name=foo`) that field would be properly updated. But the ClusterRoleBinding & RoleBinding would still have `ack-ec2-controller`

gunzy83:
> > @mtougeron @eadasiak any thoughts on this change? i'm wondering how did this work with all the CARM installations we've done before?
> 
> I'm not sure about the code that generates these templates, but I'm assuming it's probably that whatever populates `name: {{ .ServiceAccountName }}` here https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/helm/values.yaml.tpl#L141 sets it essentially to the same value by default. So it you didn't change from the ServiceAccount name the chart uses by default it would work fine. But if you did try to change it, it wouldn't work properly.
> 
> For example, the [rendered Deployment manifest](https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/helm/templates/deployment.yaml.tpl#L36) uses `serviceAccountName: {{ IncludeTemplate "service-account.name" }}` so if someone sets
> 
> ```yaml
> serviceAccount:
>   name: foo
> ```
> 
> in their values.yaml (or via `--set serviceAccount.name=foo`) that field would be properly updated. But the ClusterRoleBinding & RoleBinding would still have `ack-ec2-controller`

I think this is only a problem when `installScope` is `namespace`. When `installScope` is `cluster` the cluster role provides enough permissions and the binding for that is updated with a custom service account name. Perhaps the subset of users who both run with `installScope=namespace` and a custom service account name is small.

a-hilaly:
Thank you @gunzy83 and @mtougeron !! Yeah looks like there is an edge case [where there is a namespace mismatch between the service account name and the binding](https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/helm/templates/caches-role-binding.yaml.tpl#L11). I just tested this patch with new controllers and it's indeed generating the expected namespace value.

gunzy83:
Thanks everyone! :tada:

PR #564, opened 2025-01-11
Migrate ACK to AWS SDK GO V2

michaelhtm:
Description of changes:
Since `aws-sdk-go` will no longer be maintained after 07/31/2024.
With this PR, we will be using the aws-sdk-go-v2 API models to generate
our controllers.
The changes are:
* Introduced an apiv2 package that loads the V2 models into the API structure
* Change how we set the resource and sdk, since most fields in V2 are now non-pointers
* Add V2 model files for unit tests
* Modify Unit test against these changes


pkg/apiv2/converter.go
cheeseandcereal, line 1: One thing that would be good to get confidence that we covered all edge cases of conversations to/from the model/api sdk go v2 stuff, is if you could create some sort of script that generates a bunch of controllers, without any custom code/gen/anything, and confirm they compile.

This will help tell us if the assumptions that the code here is making about the model conversion are correct or not by simply trying to use it on a bunch of controllers.

cheeseandcereal (changes requested):

rushmash91:
Nice!
/lgtm

PR #565, opened 2025-01-22
Switch ACK to CEL-Based Immutability and Remove Runtime Checks

rushmash91:
Issue #2202

Description of changes:

Updates the ACK code generation process to enforce immutability using CEL validation rules instead of runtime comparisons. By adding `+kubebuilder:validation:XValidation` annotations to fields marked `is_immutable: true`, Kubernetes `1.25+` now blocks changes at admission time, eliminating the need for separate runtime checks or conditions.

##### Why?
1. More immediate feedback: Users get a direct error when attempting to mutate an immutable field.
2. Less runtime complexity: We no longer need to detect or conditionally reject these changes in the controller code.
3. Cleaner codegen: The old `HasImmutableFieldChanges()` and similar logic can be removed.

##### Caveats
Clusters must be on Kubernetes 1.25+, which supports CEL.
[Blog](https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/)

##### Changes
•Add CEL rules for `is_immutable: true` fields in the CRD template (`+kubebuilder:validation:XValidation`).
•Remove runtime immutability checks (`HasImmutableFieldChanges`, etc.) in the generated `sdk.go` files.
•Clean up references to "immutable fields” in `crd.go` to avoid duplicating logic.
•Update inline comments and docstrings to clarify that immutability is now admission-based.

### DIFF
--- a/pkg/config/field.go
+++ b/pkg/config/field.go
@@ -408,8 +408,9 @@ type FieldConfig struct {
 	// IsSecret instructs the code generator that this field should be a
 	// SecretKeyReference.
 	IsSecret bool `json:"is_secret"`
-	// IsImmutable instructs the code generator to add advisory conditions
-	// if user modifies the spec field after resource was created.
+	// IsImmutable indicates that the field is enforced as immutable at the
+	// admission layer. The code generator will add kubebuilder:validation:XValidation
+	// lines to the CRD, preventing changes to this field after it’s set.
 	IsImmutable bool `json:"is_immutable"`
 	// From instructs the code generator that the value of the field should
 	// be retrieved from the specified operation and member path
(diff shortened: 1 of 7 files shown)


rushmash91:
Controllers with immutable code in hooks {"type *resourceManager has no field or method getImmutableFieldChanges"}: 

- efs-controller
- dynamodb-controller
- eks-controller
- eventbridge-controller
- iam-controller

a-hilaly (commented):
Neat! Good job on this @rushmash91 🚀 

As you stated, we'll wait until EKS 1.24 reaches end of support ship this to all controllers

a-hilaly (approved):
Awesome! Really excited for this to go through
/unhold

PR #568, opened 2025-02-10
Generate FilterSystemTags and mirrorAWSTags in controllers

michaelhtm:
Issue [#2248](https://github.com/aws-controllers-k8s/community/issues/2248)

Description of changes:
By generating these two functions, we ensure that AWS tags will not 
be present in resource Spec during adoption/regular reconciliations.

FilterSystemTags will be called right after the ReadOne operation during
adoption, and it will ensure we don't patch AWS (aws:cloudformation..)
or Controller tags (services.aws.k8s/...) into the resource.

MirrowAWSTags instead, is called during the regular reconciliation loop,
after each ReadOne operation. It ensures that if the resource in AWS has
AWS tags, they need to be reflected in the desired Spec. This will ensure
that those tags will not be detected in the Delta and trigger an update. 
Since the controller only patches the difference between the latest and 
desired resource, mirroring the AWS tags ensures we don't include them
in our Spec.

### DIFF
--- a/templates/pkg/resource/manager.go.tpl
+++ b/templates/pkg/resource/manager.go.tpl
@@ -89,6 +89,7 @@ func (rm *resourceManager) ReadOne(
 		panic("resource manager's ReadOne() method received resource with nil CR object")
 	}
 	observed, err := rm.sdkFind(ctx, r)
+	mirrorAWSTags(r, observed)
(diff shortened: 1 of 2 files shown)


a-hilaly (approved):
:+1:
/lgtm

PR #571, opened 2025-03-06
Expand the search space for shapes in `from.operation`

michaelhtm:
Description of changes:
So currently, when you define a `from` `operation` `path`
in generator.yaml fieldConfig, the code-generator checks for
the read_only configuration to decide whether to get the field
from the input shape or the output shape.

The change introduced here tries to retrieve spec fields from input
shapes first, if not found it checks output shape.
If read_only is configured to true, it does so in reverse order
(checks output shape first, if not found, checks input shape)

### DIFF
--- a/pkg/testdata/models/apis/route53/0000-00-00/generator.yaml
+++ b/pkg/testdata/models/apis/route53/0000-00-00/generator.yaml
@@ -1,0 +1,97 @@
+ignore:
+  field_paths:
+  - ChangeResourceRecordSetsOutput.ChangeInfo.Comment
+  - ResourceRecordSet.GeoProximityLocation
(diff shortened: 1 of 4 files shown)


pkg/testdata/models/apis/route53/0000-00-00/generator.yaml
a-hilaly, line 1: Do we need to include all the info in here? can we use only the directives that relevant to this test?

michaelhtm: good catch..removing them

a-hilaly (approved):
Neat, thanks @michaelhtm !
/lgtm
/retest

michaelhtm:
<h1>Nice</h1>
/test ec2-controller-test

PR #572, opened 2025-03-06
Add new tag funcitons that preserve Tag order for lists

michaelhtm:
Description of changes:
To be continued TODO


templates/pkg/resource/tags.go.tpl
```diff
@@ -86,6 +125,45 @@ func FromACKTags(tags acktags.Tags) {{ $tagFieldGoType }} {
...
+func fromACKTagsWithKeyOrder(tags acktags.Tags, keyOrder []string) {{ $tagFieldGoType }} {
+    result := {{ $tagFieldGoType }}{}
+{{- if $hookCode := Hook .CRD "pre_convert_from_ack_tags" }}
+{{ $hookCode }}
+{{ end }}
+
+{{- if eq "list" $tagFieldShapeType }}
+    for _, k := range keyOrder {
+		kCopy := k
+		vCopy := tags[k]
```
a-hilaly, line 142: Do we really need to copy here?

templates/pkg/resource/tags.go.tpl
a-hilaly, line 1: It's nearly impossible to test Go tempalted based implementations, can we switch to pure Go implementation so that we can unit test it?

templates/pkg/resource/tags.go.tpl
```diff
@@ -62,6 +62,45 @@ func ToACKTags(tags {{ $tagFieldGoType }}) acktags.Tags {
...
+                result[*t.{{ $keyMemberName }}] = *t.{{ $valueMemberName }}
+            }
+        }
+    }
+{{ end }}
+{{- if $hookCode := Hook .CRD "post_convert_to_ack_tags" }}
+{{ $hookCode }}
+{{ end }}
+    return result, keyOrder
+}
```
a-hilaly, line 102: I'm not a big fan of add more functions to this file, can just update the existing ones?

a-hilaly (commented):
Thanks Michael! a few comments in line

templates/pkg/resource/tags.go.tpl
michaelhtm: no, just following the FromACKTags. removing..

templates/pkg/resource/tags.go.tpl
michaelhtm: i see that..this could be part of the code-gen refactor. Especially since i'll try to combine the two functions..

templates/pkg/resource/tags.go.tpl
michaelhtm: it would be trickier to change the existing functions, as they are used in delta..
i believe calling them in this new function could be the way to go.
eg.

```go
func toACKTagsWithKeyOrder(tags []*svcapitypes.Tag) (acktags.Tags, []string) {
	result := acktags.NewTags()
	keyOrder := []string{}
	if tags == nil || len(tags) == 0 {
		return result, keyOrder
	}

	result = ToACKTags(tags)

	for _, t := range tags {
		if t.Key != nil {
			keyOrder = append(keyOrder, *t.Key)
		}
	}

	return result, keyOrder
}
```

templates/pkg/resource/tags.go.tpl
michaelhtm: actually vCopy might be needed, as retrieval from map returns a `string, bool`..

a-hilaly (approved):

a-hilaly (approved):
Thanks @michaelhtm ! Neat description :)

michaelhtm:
retest

a-hilaly (approved):

PR #573, opened 2025-03-13
Replace To/FromACKTags with to/fromACKTagsWithKeyOrder

michaelhtm:
Description of changes:
This change also ensures these functions are generated using 
go code instead of templating. This keeps the code cleaner and testable.

### DIFF
--- a/pkg/generate/ack/controller.go
+++ b/pkg/generate/ack/controller.go
@@ -203,6 +203,12 @@ var (
 		"GoCodeClearResolvedReferences": func(f *ackmodel.Field, targetVarName string, indentLevel int) string {
 			return code.ClearResolvedReferencesForField(f, targetVarName, indentLevel)
 		},
+		"GoCodeConvertToACKTags": func(r *ackmodel.CRD, sourceVarName string, targetVarName string, keyOrderVarName string, indentLevel int) string {
+			return code.GoCodeConvertToACKTags(r, sourceVarName, targetVarName, keyOrderVarName, indentLevel)
+		},
+		"GoCodeFromACKTags": func(r *ackmodel.CRD, sourceVarName string, orderVarName string, targetVarName string, indentLevel int) string {
+			return code.GoCodeFromACKTags(r, sourceVarName, orderVarName, targetVarName, indentLevel)
+		},
 	}
 )
 
--- a/pkg/generate/code/tags.go
+++ b/pkg/generate/code/tags.go
@@ -1,0 +1,157 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package code
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/aws-controllers-k8s/code-generator/pkg/model"
+)
+
+// GoCodeToACKTags returns Go code that converts Resource Tags
+// to ACK Tags. If Resource Tags field is of type list, we
+// also maintain and return the order of the list as a []string
+// 
+//
+//
+// Sample output:
+//
+//	 for _, k := range keyOrder {
+//	 	v, ok := tags[k]
+//	 	if ok {
+//	 		tag := svcapitypes.Tag{Key: &k, Value: &v}
+//	 		result = append(result, &tag)
+//	 		delete(tags, k)
+//	 	}
+//	 }
+//	 for k, v := range tags {
+//	 	tag := svcapitypes.Tag{Key: &k, Value: &v}
+//	 	result = append(result, &tag)
+//	 }
+func GoCodeConvertToACKTags(r *model.CRD, sourceVarName string, targetVarName string, keyOrderVarName string, indentLevel int) string {
+
+	out := "\n"
+	indent := strings.Repeat("\t", indentLevel)
+	tagField, err := r.GetTagField()
+	if err != nil {
+		panic("error: resource does not have tags. ignore in generator.yaml")
+	}
+
+	if tagField == nil {
+		return ""
+	}
+
+	tagFieldShapeType := tagField.ShapeRef.Shape.Type
+	keyMemberName := r.GetTagKeyMemberName()
+	valueMemberName := r.GetTagValueMemberName()
+
+	out += fmt.Sprintf("%sif len(%s) == 0 {\n", indent, sourceVarName)
+	out += fmt.Sprintf("%s\treturn %s, %s\n", indent, targetVarName, keyOrderVarName)
+	out += fmt.Sprintf("%s}\n", indent)
+
+	switch tagFieldShapeType {
+	case "list":
+		out += fmt.Sprintf("%sfor _, t := range %s {\n", indent, sourceVarName)
+		out += fmt.Sprintf("%s\tif t.%s != nil {\n", indent, keyMemberName)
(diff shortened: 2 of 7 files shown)


pkg/generate/ack/controller.go
```diff
@@ -203,6 +203,12 @@ var (
 		"GoCodeClearResolvedReferences": func(f *ackmodel.Field, targetVarName string, indentLevel int) string {
 			return code.ClearResolvedReferencesForField(f, targetVarName, indentLevel)
 		},
+		"GoCodeToACKTags": func(r *ackmodel.CRD, sourceVarName string, targetVarName string, keyOrderVarName string, indentLevel int) string {
```
a-hilaly, line 206:
```suggestion
		"GoCodeConvertToACKTags": func(r *ackmodel.CRD, sourceVarName string, targetVarName string, keyOrderVarName string, indentLevel int) string {
```

pkg/generate/code/tags.go
```diff
@@ -0,0 +1,153 @@
...
+//	 		tag := svcapitypes.Tag{Key: &k, Value: &v}
+//	 		result = append(result, &tag)
+//	 		delete(tags, k)
+//	 	}
+//	 }
+//	 for k, v := range tags {
+//	 	tag := svcapitypes.Tag{Key: &k, Value: &v}
+//	 	result = append(result, &tag)
+//	 }
+func GoCodeToACKTags(r *model.CRD, sourceVarName string, targetVarName string, keyOrderVarName string, indentLevel int) string {
```
a-hilaly, line 43: ditto

pkg/generate/code/tags.go
```diff
@@ -0,0 +1,153 @@
...
+
+	out := "\n"
+	indent := strings.Repeat("\t", indentLevel)
+	tagField, _ := r.GetTagField()
```
a-hilaly, line 47: Should we handle the error here?

pkg/generate/code/tags.go
```diff
@@ -0,0 +1,153 @@
...
+
+	switch tagFieldShapeType {
+	case "list":
+		out += fmt.Sprintf("%sfor _, t := range %s {\n", indent, sourceVarName)
+		out += fmt.Sprintf("%s\tif t.%s != nil {\n", indent, keyMemberName)
+		out += fmt.Sprintf("%s\t\t%s = append(%s, *t.%s)\n", indent, keyOrderVarName, keyOrderVarName, keyMemberName)
+		out += fmt.Sprintf("%s\t\tif t.%s != nil {\n", indent, valueMemberName)
+		out += fmt.Sprintf("%s\t\t\t%s[*t.%s] = *t.%s\n", indent, targetVarName, keyMemberName, valueMemberName)
+		out += fmt.Sprintf("%s\t\t} else {\n", indent)
+		out += fmt.Sprintf("%s\t\t\t%s[*t.%s] = *t.%s\n", indent, targetVarName, keyMemberName, valueMemberName)
```
a-hilaly, line 69:
```suggestion
		out += fmt.Sprintf("%s\t\t\t%s[*t.%s] = nil\n", indent, targetVarName, keyMemberName, valueMemberName)
```

pkg/generate/code/tags.go
```diff
@@ -0,0 +1,153 @@
...
+		out += fmt.Sprintf("%sfor _, k := range %s {\n", indent, orderVarName)
+		out += fmt.Sprintf("%s\tv, ok := %s[k]\n", indent, tagsSourceVarName)
+		out += fmt.Sprintf("%s\tif ok {\n", indent)
+		out += fmt.Sprintf("%s\t\ttag := svcapitypes.%s{%s: &k, %s: &v}\n", indent, tagFieldGoType, keyMemberName, valueMemberName)
+		out += fmt.Sprintf("%s\t\t%s = append(%s, &tag)\n", indent, targetVarName, targetVarName)
+		out += fmt.Sprintf("%s\t\tdelete(%s, k)\n", indent, tagsSourceVarName)
+		out += fmt.Sprintf("%s\t}\n", indent)
+		out += fmt.Sprintf("%s}\n", indent)
+	case "map":
+		out += fmt.Sprintf("%s_ = %s\n", indent, orderVarName)
```
a-hilaly, line 137: is this step needed?

pkg/generate/code/tags.go
```diff
@@ -0,0 +1,153 @@
...
+		panic(msg)
+	}
+
+	out += fmt.Sprintf("%sfor k, v := range %s {\n", indent, tagsSourceVarName)
+	if tagFieldShapeType == "list" {
+		out += fmt.Sprintf("%s\ttag := svcapitypes.%s{%s: &k, %s: &v}\n", indent, tagFieldGoType, keyMemberName, valueMemberName)
+		out += fmt.Sprintf("%s\t%s = append(%s, &tag)\n", indent, targetVarName, targetVarName)
+	} else {
+		out += fmt.Sprintf("%s\t%s[k] = &v\n", indent, targetVarName)
+	}
```
a-hilaly, line 149: Maybe for consistency we can continue using switch cases here too?

templates/pkg/resource/tags.go.tpl
a-hilaly, line 1: 👍 👍

a-hilaly (commented):
w00t! Thanks michael!

pkg/generate/code/tags.go
michaelhtm: what error?

michaelhtm: i would assume there is no error, since it's handled before the execution of this file

pkg/generate/code/tags.go
michaelhtm: i'm trying to avoid the `unused parameter` warning..

pkg/generate/code/tags.go
michaelhtm: of course

pkg/generate/code/tags.go
a-hilaly: `r.GetTagField()` returns an error as a second param

pkg/generate/code/tags.go
a-hilaly: Gotcha!

pkg/generate/code/tags.go
```diff
@@ -0,0 +1,154 @@
...
+	case "map":
+		out += fmt.Sprintf("%sfor k, v := range %s {\n", indent, sourceVarName)
+		out += fmt.Sprintf("%s\tif v == nil {\n", indent)
+		out += fmt.Sprintf("%s\t\t%s[k] = \"\"\n", indent, targetVarName)
+		out += fmt.Sprintf("%s\t} else {\n", indent)
+		out += fmt.Sprintf("%s\t\t%s[k] = *v\n", indent, targetVarName)
+		out += fmt.Sprintf("%s\t}\n", indent)
+		out += fmt.Sprintf("%s}\n", indent)
+	default:
+		msg := "error: tag type can only be a list of a map"
```
a-hilaly, line 83: one tiny nit:
```suggestion
		msg := "error: tag type can only be a list or a map"
```

pkg/generate/code/tags.go
```diff
@@ -0,0 +1,154 @@
...
+//	 	result = append(result, &tag)
+//	 }
+func GoCodeFromACKTags(r *model.CRD, tagsSourceVarName string, orderVarName string, targetVarName string, indentLevel int) string {
+	out := "\n"
+	indent := strings.Repeat("\t", indentLevel)
+	tagField, _ := r.GetTagField()
+
+	if tagField == nil {
+		return ""
+	}
```
a-hilaly, line 119:
```suggestion
	out := "\n"
	indent := strings.Repeat("\t", indentLevel)

	tagField, err := r.GetTagField()
	if err != nil {
	   panic("...")
	}
	if tagField == nil {
		return ""
	}
```

a-hilaly: I understand that `GetTagField` error is probably handled at some early stages of in code-gen. If we wanted to avoid checking for errors here, need different function signatures and different paradigm. It is always a good practice to check for errors.

a-hilaly (approved):
👍

rushmash91 (approved):
Yay less templates!! Thank you @michaelhtm 🚀

PR #574, opened 2025-03-18
Add support for custom shapes using simplified schema

rushmash91:
Description of changes:

- Added `CustomShapes` field to the Config struct to store user-defined custom shapes
- Implemented `InjectSimpleSchemaShapes` function that processes custom shapes and injects them into the SDK API model
- Added dependency on `github.com/kro-run/kro` package for schema handling


pkg/config/config.go
```diff
@@ -54,6 +54,23 @@ type Config struct {
...
+	//     list_of: ShapeName    # For arrays of the shape
+	//     map_of: ShapeName     # For maps with the shape as values
+	//
+	// Example:
+	//   custom_shapes:
+	//     ReplicaStatus:
+	//       Status: string
+	//       Region: string
+	//       Description: string
+
```
a-hilaly, line 72:
```suggestion
```

a-hilaly: https://pkg.go.dev/std is using these comments to generate documentation for the Go code.  Quoting from https://tip.golang.org/doc/comment

> “Doc comments” are comments that appear immediately before top-level package, const, func, type, and var declarations with **no intervening** newlines

pkg/sdk/simpleschema.go
```diff
@@ -0,0 +1,84 @@
...
+	for shapeName, fieldsMap := range h.cfg.CustomShapes {
+		// Convert fieldsMap to map[string]interface{} for ToOpenAPISpec
+		schemaObj := make(map[string]interface{})
+		for fieldName, fieldType := range fieldsMap {
+			schemaObj[fieldName] = fieldType
+		}
+
+		openAPISchema, err := simpleschema.ToOpenAPISpec(schemaObj)
+		if err != nil {
+			panic(fmt.Sprintf("Failed to parse schema for shape %s: %v", shapeName, err))
```
a-hilaly, line 42: Can we just return an error here?

pkg/sdk/simpleschema.go
```diff
@@ -0,0 +1,84 @@
...
+
+	awssdkmodel "github.com/aws-controllers-k8s/code-generator/pkg/api"
+	ackmodel "github.com/aws-controllers-k8s/code-generator/pkg/model"
+	simpleschema "github.com/kro-run/kro/pkg/simpleschema"
+)
+
+// InjectSimpleSchemaShapes processes custom shapes from the top-level custom_shapes
+// section and injects them into the SDK API model.
+// Only string types are supported - all other types will cause a panic.
+func (h *Helper) InjectSimpleSchemaShapes(sdkapi *ackmodel.SDKAPI) error {
```
a-hilaly, line 28: I guess we still need to call this function somewhere to inject the shapes?

pkg/sdk/simpleschema.go
```diff
@@ -0,0 +1,84 @@
...
+			if prop.Type != "string" && !strings.Contains(prop.Type, "string") {
+				panic(fmt.Sprintf("Field %s in shape %s has non-string type '%s'",
+					fieldName, shapeName, prop.Type))
+			}
+
+			shape.MemberRefs[fieldName] = &awssdkmodel.ShapeRef{
+				API: sdkapi.API,
+				Shape: &awssdkmodel.Shape{
+					API:       sdkapi.API,
+					ShapeName: fmt.Sprintf("%sType", fieldName),
```
a-hilaly, line 64: Curious to understand why we're calling it `%sType`
```suggestion
					ShapeName: fmt.Sprintf("%sType", fieldName),
```

pkg/sdk/simpleschema.go
```diff
@@ -0,0 +1,84 @@
...
+		openAPISchema, err := simpleschema.ToOpenAPISpec(schemaObj)
+		if err != nil {
+			panic(fmt.Sprintf("Failed to parse schema for shape %s: %v", shapeName, err))
+		}
+
+		shape := &awssdkmodel.Shape{
+			API:           sdkapi.API,
+			ShapeName:     shapeName,
+			Type:          "structure",
+			Documentation: "Generated from SimpleSchema",
```
a-hilaly, line 49:
```suggestion
			Documentation: "Custom ACK type",
```

a-hilaly (commented):
Thanks a lot Arush!

pkg/sdk/simpleschema.go
rushmash91: Yes!

pkg/sdk/helper.go
```diff
@@ -111,6 +111,7 @@ func (h *Helper) API(serviceModelName string) (*model.SDKAPI, error) {
 		sdkapi := model.NewSDKAPI(api, h.APIGroupSuffix)
 
 		h.InjectCustomShapes(sdkapi)
+		h.InjectSimpleSchemaShapes(sdkapi)
```
a-hilaly, line 114: Shall we make those one function? At least i believe we need to inject custom structs before we start injecting `map_of` and `list_of`

pkg/config/config.go
```diff
@@ -54,6 +54,22 @@ type Config struct {
...
 	// This will also change the helm chart and image names.
 	ControllerName string `json:"controller_name,omitempty"`
+	// CustomShapes defines custom structure types that can be referenced in resource fields.
+	// The outer map is keyed by shape name (e.g., "ReplicaStatus"), and the inner map
+	// is keyed by field name with string values representing field types.
+	//
+	// These shapes can be referenced in a resource's fields section using:
+	//   custom_field:
+	//     list_of: ShapeName    # For arrays of the shape
+	//     map_of: ShapeName     # For maps with the shape as values
```
a-hilaly, line 64: Maybe let's put this after the Example in L66

pkg/model/model.go
```diff
@@ -462,6 +462,14 @@ func (m *Model) GetTypeDefs() ([]*TypeDef, error) {
 		return tdefs[i].Names.Camel < tdefs[j].Names.Camel
 	})
 	m.processNestedFieldTypeDefs(tdefs)
+
+	// Add custom type definitions for custom shapes
+	var err error
+	tdefs, err = m.AddCustomTypeDefinitions(tdefs)
+	if err != nil {
+		return nil, fmt.Errorf("error adding custom type definitions: %w", err)
+	}
```
a-hilaly, line 471: maybe this function can just return the customTypes and we'll append them.

```suggestion
	customTypes, err := m.AddCustomTypeDefinitions(tdefs)
	if err != nil {
		return nil, fmt.Errorf("error adding custom type definitions: %w", err)
	}
```

pkg/model/model.go
```diff
@@ -1039,3 +1047,49 @@ func New(
 	m.ApplyShapeIgnoreRules()
 	return m, nil
 }
+
+// AddCustomTypeDefinitions adds type definitions for custom shapes to the typeDefs slice
+// It checks if a type def for the shape already exists and skips adding it if so.
+// It also checks if the shape is a structure and only adds type defs for structure types.
+func (m *Model) AddCustomTypeDefinitions(tdefs []*TypeDef) ([]*TypeDef, error) {
+	for _, customShape := range m.SDKAPI.CustomShapes {
```
a-hilaly, line 1055: why are we looping over `	for _, customShape := range m.SDKAPI.CustomShapes {`, aren't we supposed to loop over cfg.CustomShapes?

pkg/model/model.go
```diff
@@ -1039,3 +1047,49 @@ func New(
...
+		shapeExists := false
+		for _, tdef := range tdefs {
+			if tdef.Shape.ShapeName == shape.ShapeName {
+				shapeExists = true
+				break
+			}
+		}
+		if shapeExists {
+			continue
+		}
```
a-hilaly, line 1071: Should we error instead?

pkg/sdk/simpleschema.go
```diff
@@ -0,0 +1,150 @@
...
+
+		// Create and register the list shape for this custom shape
+		h.createAndRegisterListShape(sdkapi, shapeName, shapeRef)
+	}
+
+	return nil
+}
+
+// convertFieldsMapToSchemaObj converts a fields map to a schema object for OpenAPI spec generation
+func convertFieldsMapToSchemaObj(fieldsMap map[string]string) map[string]interface{} {
```
a-hilaly, line 59: nit:
```suggestion
func convertMapKeys(fieldsMap map[string]string) map[string]interface{} {
```

pkg/model/model.go
rushmash91: `InjectSimpleSchemaShapes` reads from `config` -> adds to `sdkapi`
`AddCustomTypeDefinitions` reads from `SDKAPI.CustomShapes` -> creates typedfs

do they are in `sdkapi`

pkg/sdk/custom_shapes.go
```diff
@@ -42,6 +42,7 @@ type customShapeInjector struct {
 // the SDKAPI object.
 func (h *Helper) InjectCustomShapes(sdkapi *ackmodel.SDKAPI) error {
 	injector := customShapeInjector{sdkapi}
+	h.InjectSimpleSchemaShapes(sdkapi)
```
a-hilaly, line 45: shall we declare this before injector? flow looks like:
-> we inject some custom structs
-> initialise the injector
-> inject list/map custom types.

Now i'm actually thinking should we move some of that logic to the customShapeInjector?

pkg/model/model.go
a-hilaly: ohh, makes sense!

pkg/model/model.go
```diff
@@ -1039,3 +1047,47 @@ func New(
...
```
a-hilaly, line 1069: can we just return an error here?

a-hilaly: also `shapeExists` variable is not needed

pkg/sdk/simpleschema.go
```diff
@@ -0,0 +1,109 @@
...
+	injector := customShapeInjector{sdkapi}
+	fieldShape := &awssdkmodel.Shape{
+		API:       sdkapi.API,
+		ShapeName: fieldName,
+		Type:      "string",
+	}
+
+	sdkapi.API.Shapes[fieldShape.ShapeName] = fieldShape
+	parentShape.MemberRefs[fieldName] = injector.createShapeRefForMember(fieldShape)
+}
```
a-hilaly, line 109: looks like this function can only add `string` shapes, should we make generic to support other types too (even thought high level abstractions don't support that) or should we name it to something like `addStringField`

go.mod
```diff
@@ -21,11 +19,16 @@ require (
...
-	golang.org/x/mod v0.17.0
+	golang.org/x/mod v0.24.0
 	k8s.io/apimachinery v0.31.0
 	sigs.k8s.io/controller-runtime v0.19.0
 )
 
+require (
+	github.com/kro-run/kro v0.2.2
+	k8s.io/apiextensions-apiserver v0.31.0
+)
```
a-hilaly, line 30: shall we just merge these in L32?

pkg/sdk/simpleschema.go
rushmash91: making it add string for now, can have other function later when other types are supported

pkg/model/model.go
rushmash91: Declaration

```
ReplicaCustomStatus:
      RegionName: string
      ReplicaStatus: string
      ReplicaStatusDescription: string
      ReplicaStatusPercentProgress: string
      ReplicaInaccessibleDateTime: string
      
```

Reference
```
        ReplicasDescriptions:
        custom_field:
          list_of: ReplicaCustomStatus
        is_read_only: true
```
  
 when `newList(memberShapeName string) ` is called to create the list in `InjectCustomShapes` it tries to create the same shape again. Need to verify that if its created its not done again.

rushmash91: if I panic instead i get `panic: shape ReplicaCustomStatus already exists`

go.mod
```diff
@@ -21,12 +19,14 @@ require (
...
 	github.com/stretchr/testify v1.9.0
-	golang.org/x/mod v0.17.0
+	golang.org/x/mod v0.24.0
 	k8s.io/apimachinery v0.31.0
 	sigs.k8s.io/controller-runtime v0.19.0
 )
 
 require (
+	github.com/kro-run/kro v0.2.2
+	k8s.io/apiextensions-apiserver v0.31.0
```
a-hilaly, line 29: i think if you run `go mod tidy` it's gonna fix ordering

go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/aws-controllers-k8s/code-generator
 
-go 1.22.0
-
-toolchain go1.22.4
+go 1.24.0
```
a-hilaly, line 3: was this automatically or manually set?

pkg/sdk/helper.go
```diff
@@ -129,3 +130,25 @@ func (h *Helper) ModelAndDocsPath(serviceModelName string) string {
 	)
 	return modelPath
 }
+
+// GetOriginalAPIShapeSet returns a map set (keys with empty struct values) of all original shape names
+// from the AWS service API model. This provides efficient lookups to check if a shape name exists.
+func (h *Helper) GetOriginalAPIShapeSet(serviceNameOrID string) (map[string]struct{}, error) {
+	modelPath := h.ModelAndDocsPath(serviceNameOrID)
+	apis, err := apiv2.LoadAPI(modelPath)
```
a-hilaly, line 138: i think one of the reasons why the code-generator is slower today is the increased runtime complexity of the apiv2.LoadApi. Not sure if we should take this route

pkg/sdk/helper.go
```diff
@@ -129,3 +130,25 @@ func (h *Helper) ModelAndDocsPath(serviceModelName string) string {
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, api := range apis {
+		_ = api.ServicePackageDoc()
+		shapeSet := make(map[string]struct{}, len(api.Shapes))
+		for shapeName := range api.Shapes {
+			shapeSet[shapeName] = struct{}{}
+			shapeSet[strings.ToLower(shapeName)] = struct{}{}
```
a-hilaly, line 148: do we need both lower and original casing?

pkg/sdk/helper.go
```diff
@@ -129,3 +130,25 @@ func (h *Helper) ModelAndDocsPath(serviceModelName string) string {
...
+// from the AWS service API model. This provides efficient lookups to check if a shape name exists.
+func (h *Helper) GetOriginalAPIShapeSet(serviceNameOrID string) (map[string]struct{}, error) {
+	modelPath := h.ModelAndDocsPath(serviceNameOrID)
+	apis, err := apiv2.LoadAPI(modelPath)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, api := range apis {
+		_ = api.ServicePackageDoc()
```
a-hilaly, line 144: What is this call used for?

pkg/sdk/simpleschema.go
```diff
@@ -0,0 +1,120 @@
...
+			MemberShapeName: nil,
+			ValueShapeName:  nil,
+		})
+	}
+
+	return nil
+}
+
+// convertFieldsMapToSchemaObj converts a fields map to a schema object for OpenAPI spec generation
+func convertMapKeys(fieldsMap map[string]string) map[string]interface{} {
```
a-hilaly, line 67: Sorry i meant `convertMapValues`..

pkg/model/model.go
a-hilaly: Should `custom_field.list_of` and `custom_field.map_of` inject a post fix to the shape name?

pkg/model/model.go
```diff
@@ -412,7 +412,7 @@ func (m *Model) GetTypeDefs() ([]*TypeDef, error) {
 	if m.typeDefs != nil {
 		return m.typeDefs, nil
 	}
-
+
```
a-hilaly, line 415: can you please run gofmt for all these files?

pkg/sdk/custom_shapes.go
```diff
@@ -41,6 +41,10 @@ type customShapeInjector struct {
 // into the list of shapes in the API and update the list of custom shapes in
 // the SDKAPI object.
 func (h *Helper) InjectCustomShapes(sdkapi *ackmodel.SDKAPI) error {
+	err := h.InjectSimpleSchemaShapes(sdkapi)
+	if err != nil {
+		return err
+	}
```
a-hilaly, line 47: Should we make `InjectSimpleSchemaShapes` part of `customShapeInjector` ? feels like that's where it should be living. :)

pkg/sdk/simpleschema.go
```diff
@@ -0,0 +1,115 @@
...
+
+package sdk
+
+import (
+	"fmt"
+
+	awssdkmodel "github.com/aws-controllers-k8s/code-generator/pkg/api"
+	ackmodel "github.com/aws-controllers-k8s/code-generator/pkg/model"
+	simpleschema "github.com/kro-run/kro/pkg/simpleschema"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
a-hilaly, line 22: nit: need to order the imports: checkout the standard here https://github.com/golang/tools/commit/ed69e84b1518b5857a9f4e01d1f9cefdcc45246e

pkg/sdk/simpleschema.go
```diff
@@ -0,0 +1,115 @@
...
+// section and injects them into the SDK API model.
+// Only string types are supported - all other types will cause a panic.
+func (h *Helper) InjectSimpleSchemaShapes(sdkapi *ackmodel.SDKAPI) error {
+	if len(h.cfg.CustomShapes) == 0 {
+		return nil
+	}
+
+	apiShapeNames := sdkapi.API.ShapeNames()
+	for shapeName, fieldsMap := range h.cfg.CustomShapes {
+		// check dublicates
```
a-hilaly, line 35:
```suggestion
		// check for duplicates
```

pkg/sdk/simpleschema.go
```diff
@@ -0,0 +1,115 @@
...
+		for _, as := range apiShapeNames {
+			if as == shapeName {
+				return fmt.Errorf("shapeName %s already exists in the API", shapeName)
```
a-hilaly, line 38:
```suggestion
				return fmt.Errorf("CustomType name %s already exists in the API", shapeName)
```

pkg/sdk/simpleschema.go
```diff
@@ -0,0 +1,115 @@
...
+func convertMapValues(fieldsMap map[string]string) map[string]interface{} {
+	schemaObj := make(map[string]interface{})
+	for fieldName, fieldType := range fieldsMap {
+		schemaObj[fieldName] = fieldType
+	}
+	return schemaObj
+}
+
+// createBaseShape creates a base structure shape with its member fields
+func (h *Helper) createBaseShape(
```
a-hilaly, line 71: How about `newStructureShape` - maybe when combined with the `customShapeInjector` we can reuse some of the functions it already have.

go.mod
a-hilaly, line 29: nit: need to execute `go mod tidy`

a-hilaly (commented):
Awesome, thanks a lot Arush!! Getting very close - a few minor comments:

pkg/model/model_dynamodb_test.go
```diff
@@ -107,3 +107,41 @@ func TestDynamoDB_Table(t *testing.T) {
...
+		"RegionStatusDescription",
+		"RegionStatusPercentProgress",
+		"RegionInaccessibleDateTime",
+	}
+
+	for _, fieldName := range expectedFields {
+		assert.Contains(replicasStateShape.MemberRefs, fieldName, "RepicasState shape is missing field: "+fieldName)
+		field := replicasStateShape.MemberRefs[fieldName]
+		assert.Equal("string", field.Shape.Type, "Field "+fieldName+" should be of type string")
+	}
```
a-hilaly, line 146: nit: can we also assert that the length of MemberRefs is `len(expectedFields)`

pkg/model/sdk_api.go
```diff
@@ -134,14 +134,14 @@ func (a *SDKAPI) GetShapeRefFromType(
 	// TODO(jaypipes): Only handling maps with string keys at the moment...
 	isMap := strings.HasPrefix(typeOverride, "map[string]")
 	if isMap {
-		elemType = typeOverride[11:len(typeOverride)]
+		elemType = typeOverride[11:]
```
a-hilaly, line 137: 👍

pkg/sdk/custom_shapes.go
```diff
@@ -18,8 +18,10 @@ import (
 	"fmt"
 
 	awssdkmodel "github.com/aws-controllers-k8s/code-generator/pkg/api"
+	simpleschema "github.com/kro-run/kro/pkg/simpleschema"
 
 	ackmodel "github.com/aws-controllers-k8s/code-generator/pkg/model"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
a-hilaly, line 24: The standard is first standard libraries, then external, then internal

```suggestion
	simpleschema "github.com/kro-run/kro/pkg/simpleschema"
	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"

	ackmodel "github.com/aws-controllers-k8s/code-generator/pkg/model"
	awssdkmodel "github.com/aws-controllers-k8s/code-generator/pkg/api"
```

pkg/sdk/custom_shapes.go
```diff
@@ -62,10 +67,99 @@ func (h *Helper) InjectCustomShapes(sdkapi *ackmodel.SDKAPI) error {
...
 }
 
+// convertMapValues converts a fields map to a schema object for OpenAPI spec generation
+func convertMapValues(fieldsMap map[string]interface{}) map[string]interface{} {
+	schemaObj := make(map[string]interface{})
+	for fieldName, fieldType := range fieldsMap {
+		schemaObj[fieldName] = fieldType
+	}
+	return schemaObj
+}
```
a-hilaly, line 116: Do you still need this function? input and output are identical

pkg/testdata/models/apis/dynamodb/0000-00-00/generator-with-custom-shapes.yaml
a-hilaly, line 1: 👍

a-hilaly (approved):
Awesome works on this @rushmash91 ! thanks a lot for the quick updates :)
/lgtm

PR #577, opened 2025-03-28
remove apigateway for shameList

rushmash91:
This pull request aims to remove API Gateway from the shameList, as the original reason for its inclusion is no longer valid. In the AWS SDK v1, there was an inconsistency in the naming convention between input and output shapes for API Gateway operations. Specifically, input shapes followed a pattern like `CreateDomainNameInput`, while output shapes were simply named `DomainName`. This inconsistency was one of the reasons API Gateway was placed on the shameList.

However, with the release of AWS SDK v2, this naming inconsistency has been resolved. The new SDK now uses a standardized naming pattern for both input and output shapes across API Gateway operations. As a result, the special handling that was required is no longer necessary.

For reference, we can compare the old naming convention in SDK v1 ([link](https://github.com/aws/aws-sdk-go/blob/main/service/apigateway/api.go#L5162)) with the new, consistent naming in SDK v2 ([link](https://github.com/aws/aws-sdk-go-v2/blob/main/service/apigateway/api_op_GetDomainName.go#L49)).


a-hilaly (approved):
👍 
/lgtm

PR #580, opened 2025-04-08
add support for `reconcile-resources` flag

rushmash91:
Description of changes:
https://github.com/aws-controllers-k8s/runtime/pull/176

This PR adds template changes to support the reconcile-resources flag which was recently added to the ACK runtime. This flag allows users to specify which resource kinds should be reconciled by a controller, providing more granular control over controller behavior.
When the flag is set, only the specified resource kinds will be reconciled by the controller. If unspecified, all resources will be reconciled (default behavior).

### DIFF
--- a/pkg/generate/ack/release.go
+++ b/pkg/generate/ack/release.go
@@ -21,6 +21,7 @@ import (
 	"github.com/aws-controllers-k8s/code-generator/pkg/generate/templateset"
 	ackmetadata "github.com/aws-controllers-k8s/code-generator/pkg/metadata"
 	ackmodel "github.com/aws-controllers-k8s/code-generator/pkg/model"
+	"github.com/aws-controllers-k8s/pkg/names"
 )
 
 var (
@@ -103,16 +104,9 @@ func Release(
 		releaseFuncMap(m.MetaVars().ControllerName),
 	)
 	metaVars := m.MetaVars()
-
-	crds, err := m.GetCRDs()
-	if err != nil {
-		return nil, err
-	}
-
-	reconcileResources := make([]string, 0)
-
-	for _, crd := range crds {
-		reconcileResources = append(reconcileResources, crd.Names.Camel)
+	reconcileResources := make([]string, len(metaVars.CRDNames))
+	for i, name := range metaVars.CRDNames {
+		reconcileResources[i] = names.New(name).Camel
 	}
 
 	releaseVars := &templateReleaseVars{
--- a/pkg/generate/ack/release.go
+++ b/pkg/generate/ack/release.go
@@ -104,6 +104,17 @@ func Release(
 	)
 	metaVars := m.MetaVars()
 
+	crds, err := m.GetCRDs()
+	if err != nil {
+		return nil, err
+	}
+
+	reconcileResources := make([]string, 0)
+
+	for _, crd := range crds {
+		reconcileResources = append(reconcileResources, crd.Names.Camel)
+	}
+
 	releaseVars := &templateReleaseVars{
 		metaVars,
 		ImageReleaseVars{
@@ -112,6 +123,7 @@ func Release(
 		},
 		metadata,
 		serviceAccountName,
+		reconcileResources,
 	}
 	for _, path := range releaseTemplatePaths {
 		outPath := strings.TrimSuffix(path, ".tpl")
@@ -140,4 +152,6 @@ type templateReleaseVars struct {
 	Metadata *ackmetadata.ServiceMetadata
 	// ServiceAccountName is the name of the ServiceAccount used in the Helm chart
 	ServiceAccountName string
+	// ReconcileResources contains all CRD names in their original format
+	ReconcileResources []string
 }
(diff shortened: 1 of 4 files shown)


templates/helm/values.yaml.tpl
```diff
@@ -114,6 +114,11 @@ watchNamespace: ""
 # You can set multiple labelsSelectors by providing a comma separated list of a=b arguments. e.g "label1=value1,label2=value2" 
 watchSelectors: ""
 
+# Set the value of reconcileResources to specify which resource kinds to reconcile.
+# You can set multiple resource kinds by providing a comma separated list. e.g "Kind1,Kind2"
+# If unspecified, all resources will be reconciled.
+reconcileResources: ""
```
a-hilaly, line 120: idea: what do you think of moving this parameter to L134 instead? so we can do something like:
```yaml
reconcile:
    resources:
    - Policy
    - Role
    - User
    - ...
```

a-hilaly (commented):
Thanks @rushmash91 ! left one thought below

templates/helm/values.yaml.tpl
a-hilaly: We could also generate a default array here, in a way that it contains all the possible Kinds. Users will just have to disable the Kinds they don't need.

rushmash91: Yes, thats better

rushmash91: added a list

pkg/generate/ack/release.go
```diff
@@ -104,6 +104,17 @@ func Release(
...
 
+	crds, err := m.GetCRDs()
+	if err != nil {
+		return nil, err
+	}
+
+	reconcileResources := make([]string, 0)
+
+	for _, crd := range crds {
+		reconcileResources = append(reconcileResources, strings.ToLower(crd.Names.Camel))
```
michaelhtm, line 115: why ToLower of Camel case names?
I don't think the kind this will be compared to will be ToLower.
https://github.com/aws-controllers-k8s/runtime/pull/176/files#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R623-R627

templates/helm/values.schema.json
```diff
@@ -239,6 +239,13 @@
         },
         "resourceMaxConcurrentSyncs": {
           "type": "object"
+        },
+        "resources": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "description": "List of resource kinds to reconcile. If empty, all resources will be reconciled."
```
michaelhtm, line 248: if we're populating the all resources in helm values.yaml by default, should empty still reconcile all resources?

templates/helm/values.yaml.tpl
```diff
@@ -137,6 +137,15 @@ reconcile:
...
   # resource.
   resourceMaxConcurrentSyncs: {}
+  
+  # Set the value of resources to specify which resource kinds to reconcile.
+  # If empty, all resources will be reconciled.
+  # If specified, only the listed resource kinds will be reconciled.
+  resources: []
+  # Available resource kinds for this controller:
+  {{- range $resource := .ReconcileResources }}
+#   - {{ $resource }}
```
michaelhtm, line 147: i thought we would populate this with all resources by default..

michaelhtm (commented):
Thanks @rushmash91 👍 
left a few inline comments

templates/helm/values.schema.json
rushmash91: the flag when left empty reconciles all kinds, we are not populating them, they are added to the values.yaml to uncomment if needed making it convenient eg need to ignore one..

templates/helm/values.yaml.tpl
rushmash91: No, default if empty. you add the resources if you decide to use the flag.

https://github.com/aws-controllers-k8s/runtime/pull/176/files#diff-17e55e07558643df844d27ad1fee3822c62d8fa4e5d9adfe6c33c7c28f377659R283

pkg/generate/ack/release.go
```diff
@@ -104,6 +104,17 @@ func Release(
...
+	crds, err := m.GetCRDs()
+	if err != nil {
+		return nil, err
+	}
+
+	reconcileResources := make([]string, 0)
+
+	for _, crd := range crds {
+		reconcileResources = append(reconcileResources, crd.Names.Camel)
+	}
```
a-hilaly, line 116: `metaVars` already has the CRD names https://github.com/aws-controllers-k8s/code-generator/blob/0909e7f0adb8ffe4120a8c20d5d58b991f2539e9/pkg/generate/templateset/vars.go#L48-L49

a-hilaly (commented):
Thanks @rushmash91 ! one comment below

pkg/generate/ack/release.go
rushmash91: So, convert to names to use camel case? done

michaelhtm (approved):
👍 
/lgtm
/hold

michaelhtm (approved):

PR #584, opened 2025-05-01
Upgrade aws-controllers-k8s/pkg to v0.0.16

knottnt:
Description of changes:
Upgrade aws-controllers-k8s/pkg to v0.0.16

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/aws-controllers-k8s/code-generator
 go 1.24.0
 
 require (
-	github.com/aws-controllers-k8s/pkg v0.0.15
+	github.com/aws-controllers-k8s/pkg v0.0.16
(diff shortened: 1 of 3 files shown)


knottnt:
@a-hilaly: It looks like some properties in .go files are being changed due to a regex being applied to "HTTP" in `names.go`. 

1. Is that okay? 
2. If it is okay, how does the team generally go about resolving mismatches between the current state of the "code-generator" repo if additional changes would need to be made? For example here there a few cases where code in hooks use the old capitalization.

knottnt:
Testing out alternative regex patterns for "Http" it looks like the current pattern matching only against "HTTP" is what's causing the large blast radius. Will close this PR in favor of using a better regex pattern in `pkg`.

PR #590, opened 2025-05-13
fix: Use correct CamelCase Kind names in reconcile resources

rushmash91:
Description of changes:

Fixed the generation of reconcile resources in the Helm chart to use proper CamelCase Kind names. The previous implementation using `metaVars.CRDNames` with `names.New(name).Camel` was incorrectly converting CRD names from their lowercase plural form, resulting in improper capitalization.

The previous approach used `metaVars.CRDNames` which contains lowercase plural forms (e.g., "preparedstatements", "workgroups"). When these were processed through `names.New(name).Camel`, it only capitalized the first letter without properly handling internal word breaks, leading to incorrect resource kind names.

Issue:
```yaml
resources:
  - Preparedstatements    # incorrect
  - Workgroups           # incorrect
```

Fix:
```yaml
resources:
  - PreparedStatement  
  - WorkGroup
```


a-hilaly (approved):
Noice!

PR #595, opened 2025-05-19
Allow zero and negative values for generated float32 bounds check

knottnt:
Issue #, if available:

Description of changes:
- Update `set_sdk.go setSDKForScalar()` to allow zero and negative values for float32 in bounds check

### DIFF
--- a/pkg/generate/code/set_sdk.go
+++ b/pkg/generate/code/set_sdk.go
@@ -1615,7 +1615,13 @@ func setSDKForScalar(
 				actualType[1],
 			)
 		} else {
-			out += fmt.Sprintf("%sif %s > math.Max%s32 || %s < math.%s32 {\n", indent, dereferencedVal, actualType[0], dereferencedVal, actualType[1])
+			out += fmt.Sprintf(
+				"%[1]sif %[2]s > math.Max%[3]s32 || %[2]s < math.%[4]s32 {\n",
+				indent,
+				dereferencedVal,
+				actualType[0],
+				actualType[1],
+			)
 		}
 		out += fmt.Sprintf("%s\treturn nil, fmt.Errorf(\"error: field %s is of type %s32\")\n", indent, ogShapeName.Original, actualType[2])
 		out += fmt.Sprintf("%s}\n", indent)
--- a/pkg/generate/code/set_sdk.go
+++ b/pkg/generate/code/set_sdk.go
@@ -369,7 +369,6 @@ func SetSDK(
 					indentLevel+1,
 				)
 				out += setSDKForScalar(
-					cfg, r,
 					memberName,
 					targetVarName,
 					inputShape.Type,
@@ -391,7 +390,6 @@ func SetSDK(
 				)
 			} else {
 				out += setSDKForScalar(
-					cfg, r,
 					memberName,
 					targetVarName,
 					inputShape.Type,
@@ -577,7 +575,6 @@ func SetSDKGetAttributes(
 			indent, sourceVarPath,
 		)
 		out += setSDKForScalar(
-			cfg, r,
 			memberName,
 			targetVarName,
 			inputShape.Type,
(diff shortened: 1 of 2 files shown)


pkg/generate/code/set_sdk.go
```diff
@@ -1575,7 +1575,17 @@ func setSDKForScalar(
 		dereferencedVal := ogShapeName.CamelLower + "Copy0"
 		out += fmt.Sprintf("%s%s := %s\n", indent, dereferencedVal, setTo)
 
-		out += fmt.Sprintf("%sif %s > math.Max%s32 || %s < math.%s32 {\n", indent, dereferencedVal, actualType[0], dereferencedVal, actualType[1])
+		if shape.Type == "float" {
+			out += fmt.Sprintf(
+				"%[1]sif %[2]s > math.Max%[3]s32 || %[2]s < -math.Max%[3]s32 || (%[2]s < math.%[4]s32 && !(%[2]s <= 0)) || (%[2]s > -math.%[4]s32 && !(%[2]s >= 0)) {\n",
```
michaelhtm, line 1580: This is a nice use! we should probably use it during the refactor!

pkg/generate/code/set_sdk.go
```diff
@@ -1575,7 +1575,17 @@ func setSDKForScalar(
...
+				indent,
+				dereferencedVal,
+				actualType[0],
+				actualType[1],
+			)
+		} else {
+			out += fmt.Sprintf("%sif %s > math.Max%s32 || %s < math.%s32 {\n", indent, dereferencedVal, actualType[0], dereferencedVal, actualType[1])
```
michaelhtm, line 1587: can we use the above here as well?

pkg/generate/code/set_sdk.go
michaelhtm, line 1: Can we modify the SmallestNonZeroFloat in `intOrFloat` instead? should be in L1564

michaelhtm (commented):
Great catch @knottnt!
left a few nits

pkg/generate/code/set_sdk.go
michaelhtm: One more nit here: can you show what the sample output of this line would be?

pkg/generate/code/set_sdk.go
michaelhtm: I meant 1591..I assumed you were no longer using SmallestNonZeroInt..

a-hilaly (approved):
👍 
/lgtm

knottnt:
@michaelhtm @a-hilaly 

Example of the new output shown below. The new check is quite a bit larger. Assuming we want to prevent `float64` values smaller than `math.SmallestNonzeroFloat32` we could reduce the size of the generated bounds check by adding a new function to the `runtime` package. Then in code-gen we simply call the new function. 
```go
if temperatureCopy0 > math.MaxFloat32 || temperatureCopy0 < -math.MaxFloat32 || (temperatureCopy0 < math.SmallestNonzeroFloat32 && !(temperatureCopy0 <= 0)) || (temperatureCopy0 > -math.SmallestNonzeroFloat32 && !(temperatureCopy0 >= 0)) {
	return nil, fmt.Errorf("error: field temperature is of type float32")
}
```

This is replacing the below check which does not allow zero or negative values.
```go
if temperatureCopy0 > math.MaxFloat32 || temperatureCopy0 < math.SmallestNonzeroFloat32 {
	return nil, fmt.Errorf("error: field temperature is of type float32")
}
```

pkg/generate/code/set_sdk_whitebox_test.go
```diff
@@ -0,0 +1,132 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
knottnt, line 1: Had to use a new file since the existing `set_sdk_test.go` has a circular dependency with the code package due to `testutils`.

PR #597, opened 2025-05-22
feat: rm.IsSynced returns requeueError when false

michaelhtm:
Description of changes:
This changes introduce a new field in generator.yaml SyncedConfig, which
lets us define custom reuqueue duration. If it is not defined, the
duration will default to the DefaultSyncPeriod


pkg/generate/code/synced.go
```diff
@@ -187,11 +195,11 @@ func fieldPathSafeEqual(
 		// if r.ko.Spec.ProvisionedThroughput == nil
 		out += fmt.Sprintf("\tif %s == nil {\n", subFieldPath)
 		// return false, nil
-		out += "\t\treturn false, nil\n"
+		out += fmt.Sprintf("\t\treturn false, %s\n", notSyncedErr)
```
a-hilaly, line 198: Why do nil return trigger a 30 seconds requeue?

pkg/generate/code/synced.go
a-hilaly, line 1: Can we get unit tests here?

pkg/config/resource.go
```diff
@@ -152,6 +152,8 @@ type SyncedCondition struct {
 	Path *string `json:"path"`
 	// In contains a list of possible values `Path` should be equal to.
 	In []string `json:"in"`
+	// RequeueError custom SyncPeriod in seconds
+	RequeueTime *int `json:"requeueTime,omitempty`
```
a-hilaly, line 156: what happens when someone provides 0?

michaelhtm: Should we panic code-gen?

pkg/generate/code/synced.go
michaelhtm: not sure i understand..

PR #604, opened 2025-06-18
Support Map of SecretKeyReferences

knottnt:
Issue #, if available:

Description of changes:
- Add Test case for mapping string to secrets references
- Update CleanGoType to update value type of map to SecretKeyReference if is_secret set.
- Add tests validating SetSDK() for map of secrets references

### DIFF
--- a/pkg/generate/code/set_sdk_test.go
+++ b/pkg/generate/code/set_sdk_test.go
@@ -14,6 +14,7 @@
 package code_test
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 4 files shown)


pkg/generate/code/set_sdk_test.go
```diff
@@ -4326,3 +4327,328 @@ func TestSetSDK_WAFv2_RuleGroup_Create(t *testing.T) {
...
+		}
+		res.DeadLetterConfig = f2
+	}
+	if r.ko.Spec.Description != nil {
+		res.Description = r.ko.Spec.Description
+	}
+	if r.ko.Spec.Environment != nil {
+		f4 := &svcsdktypes.Environment{}
+		if r.ko.Spec.Environment.Variables != nil {
+			f4f0 := map[string]string{}
```
knottnt, line 4378: No changes to SetSDK were necessary, but there don't appear to be any tests validating the generated code for a map of secrets references. I was also able to verify that key/value pair of secrets references are written out as expected using a modified version of the lambda-controller.

michaelhtm (approved):

pkg/model/types.go
```diff
@@ -58,6 +58,15 @@ func CleanGoType(
 
 		gt = "[]" + mgt
 		gtwp = "[]" + mgtwp
+	} else if shape.Type == "map" && fieldCfg != nil && fieldCfg.IsSecret {
```
a-hilaly, line 61: do you need to check for `fieldCfg != nil && fieldCfg.IsSecret` ? looks like the `list` case doesn't need it

knottnt: The reason I check both here is that I only want to enter this block if the field is both a map and is secret whereas the `list` block needs be entered regardless of the secret status.

a-hilaly (approved):
Neat, thanks Trevor 👍

PR #606, opened 2025-07-02
feat: add SkipIncompleteLateInitializeCheck

michaelhtm:
Description of changes:
When a field is marked for lateInitialize, the controller tries to
ensure that field is never nil.
If by chance the field is nil, the controller will reconcile every 5
seconds until the field is no longer nil.

Although we want certain fields lateInitialized, we don't expect them to
be nonNil.

In the example of RDS DBInstance, DBInstance contains
PerformanceInsightsKMSID, that we want to lateInitialize only when
PerformanceInsights is Enabled.

For now this change only allows us the incomplete check, but moving
forward, we may want to add certain conditions, to skip or not to
skip...


pkg/config/field.go
```diff
@@ -318,6 +318,9 @@ type LateInitializeConfig struct {
 	// MaxBackoffSeconds provide the maximum allowed backoff when retrying late initialization after an
 	// unsuccessful attempt.
 	MaxBackoffSeconds int `json:"max_backoff_seconds"`
+	// SkipIncompleteCheck skips the LateInitialization incomplete check for resources.
+	// TODO: (michaelhtm) we might want to add a SkipOnCondition/Condition
+	SkipIncompleteCheck bool `json:"skip_incomplete_check,omitempty"`
```
michaelhtm, line 323: Should this be a struct instead? it would be easier to add conditions in it

pkg/generate/code/late_initialize.go
```diff
@@ -276,7 +276,10 @@ func IncompleteLateInitialization(
 	indent := strings.Repeat("\t", indentLevel)
 	var lateInitFieldNames []string
 	lateInitConfigs := cfg.GetLateInitConfigs(r.Names.Original)
-	for fieldName := range lateInitConfigs {
+	for fieldName, lateInitConfig := range lateInitConfigs {
+		if lateInitConfig.SkipIncompleteCheck {
+			continue
+		}
```
knottnt, line 282: Would this mean that we only make a single attempt to read the field after initial creation? Is there any mechanism that would allow us to eventually sync this field is the AWS resource hadn't set a skipped field by the first post-create sync?

a-hilaly:
>Although we want certain fields lateInitialized, we don't expect them to
be nonNil.
In the example of RDS DBInstance, DBInstance contains
PerformanceInsightsKMSID, that we want to lateInitialize only when
PerformanceInsights is Enabled.

But why make an extra ReadOne API call if you could already get the info from the OG ReadOne call?

knottnt (approved):

PR #607, opened 2025-07-02
chore: increase `git fetch` default timeout to 180s

michaelhtm:
Description of changes:
The timeout being 30 seconds was causing a timeout error. 
Moving forward we may want to make certain changes to make the fetch faster

### DIFF
--- a/pkg/sdk/repo.go
+++ b/pkg/sdk/repo.go
@@ -32,7 +32,7 @@ import (
 const (
 	sdkRepoURLV2           = "https://github.com/aws/aws-sdk-go-v2"
 	defaultGitCloneTimeout = 180 * time.Second
-	defaultGitFetchTimeout = 30 * time.Second
+	defaultGitFetchTimeout = 180 * time.Second
 )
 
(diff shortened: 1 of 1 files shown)


pkg/sdk/repo.go
```diff
@@ -32,7 +32,7 @@ import (
 const (
 	sdkRepoURLV2           = "https://github.com/aws/aws-sdk-go-v2"
 	defaultGitCloneTimeout = 180 * time.Second
-	defaultGitFetchTimeout = 30 * time.Second
+	defaultGitFetchTimeout = 180 * time.Second
```
a-hilaly, line 35: This is just for fetching tags right? it is hard to believe that fetching tags can take more than 30seconds, are those are just refs in the .git folder?

michaelhtm: It is..aws-sdk-go-v2 has [885](https://github.com/aws/aws-sdk-go-v2/releases) tags to be fetched...maybe moving forward we can rethink this...

a-hilaly (approved):
/lgtm

PR #608, opened 2025-07-08
fix(helm): avoid null volumes in deployment

blanchardma:
Description of changes:

The AWS ACK Helm chart generates `"volumes": null` when no static secret `.Values.aws.credentials.secretName` or no extra volume `.Values.deployment.extraVolumes` was set. This can cause issues during the deployment, particularly with admission controllers such as policy engines. An example issue can be found here https://github.com/kyverno/policies/issues/1310 by Kyverno. 

This PR solves such problems without using a workaround, like deploying an unused emptyDir.

### DIFF
--- a/templates/helm/templates/deployment.yaml.tpl
+++ b/templates/helm/templates/deployment.yaml.tpl
@@ -148,6 +148,7 @@ spec:
         {{ "{{- if .Values.deployment.extraEnvVars -}}" }}
           {{ "{{ toYaml .Values.deployment.extraEnvVars | nindent 8 }}" }}
         {{ "{{- end }}" }}
+        {{  "{{- if or .Values.aws.credentials.secretName .Values.deployment.extraVolumeMounts }}" }} 
         volumeMounts:
         {{ "{{- if .Values.aws.credentials.secretName }}" }}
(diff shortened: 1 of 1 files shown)


templates/helm/templates/deployment.yaml.tpl
```diff
@@ -195,15 +195,17 @@ spec:
       hostPID: false
       hostNetwork: {{ "{{ .Values.deployment.hostNetwork }}" }}
       dnsPolicy: {{ "{{ .Values.deployment.dnsPolicy }}" }}
+      {{ "{{- if or .Values.aws.credentials.secretName .Values.deployment.extraVolumes }}" }}
```
knottnt, line 198: [volumeMounts](https://github.com/aws-controllers-k8s/code-generator/blob/a488d4e8a9222ed51e5690cb5850beb21cce5a1f/templates/helm/templates/deployment.yaml.tpl#L151-L159) has a similar pattern. Should it also be omitted when `.Values.aws.credentials.secretName` and `.Values.deployment.extraVolumeMounts` are not set?

blanchardma: Good point, I already integrated it into the commit.

knottnt (approved):

PR #613, opened 2025-07-25
feat: return error on missing required fields during adoption

michaelhtm:
fixes [#2557](https://github.com/aws-controllers-k8s/community/issues/2557)

Description of changes:
During resource Adoption, the ACK controllers only returned an error 
when there was a missing PrimaryKey. If a Resource requires more than 1 
field to describe a Resource (eg. Lambda alias requires Name and
FunctionName), the controller only verifies the Primary identifier (ARN, 
ID, Name) and nothing else.

With this change, we look into the SDK model to retrieve the required 
fields and return an error if all are not defined during 
`PopulateResourceFromAnnotation`

### DIFF
--- a/pkg/generate/code/set_resource.go
+++ b/pkg/generate/code/set_resource.go
@@ -843,10 +843,14 @@ func identifierNameOrIDGuardConstructor(
 // the required field for a read, or returning an error here
 // and returns a `MissingNameIdentifier` error:
 //
-//	if fields[${requiredField}] == "" {
-//	 return ackerrors.MissingNameIdentifier
+//  f0, ok := fields[${requiredField}]
+//	if !ok {
+//	 return ackerrors.NewTerminalError(fmt.Errorf("required field missing: ${requiredField}"))
 //	}
 func requiredFieldGuardContructor(
+	// requiredFieldVarName is the variable where the requiredField value
+	// will be stored
+	requiredFieldVarName string,
 	// String representing the fields map that contains the required
 	// fields for adoption
 	sourceVarName string,
@@ -856,7 +860,7 @@ func requiredFieldGuardContructor(
 	indentLevel int,
 ) string {
 	indent := strings.Repeat("\t", indentLevel)
-	out := fmt.Sprintf("%stmp, ok := %s[\"%s\"]\n", indent, sourceVarName, requiredField)
+	out := fmt.Sprintf("%s%s, ok := %s[\"%s\"]\n", indent, requiredFieldVarName, sourceVarName, requiredField)
 	out += fmt.Sprintf("%sif !ok {\n", indent)
 	out += fmt.Sprintf("%s\treturn ackerrors.NewTerminalError(fmt.Errorf(\"required field missing: %s\"))\n", indent, requiredField)
 	out += fmt.Sprintf("%s}\n", indent)
@@ -1258,11 +1262,11 @@ func SetResourceIdentifiers(
 //
 // ```
 //
-//	tmp, ok := field["brokerID"]
+//	primaryKey, ok := field["brokerID"]
 //	if  !ok {
-//		return ackerrors.MissingNameIdentifier
+//		return ackerrors.NewTerminalError(fmt.Errorf("required field missing: brokerID"))
 //	}
-//	r.ko.Status.BrokerID = &tmp
+//	r.ko.Status.BrokerID = &primaryKey
 //
 // ```
 //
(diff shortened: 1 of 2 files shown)


pkg/generate/code/set_resource.go
```diff
@@ -847,6 +847,7 @@ func identifierNameOrIDGuardConstructor(
 //	 return ackerrors.MissingNameIdentifier
 //	}
 func requiredFieldGuardContructor(
+	holdingVar string,
```
a-hilaly, line 850: what does holdingVar mean? does it need to be the first param?

pkg/generate/code/set_resource.go
```diff
@@ -847,6 +847,7 @@ func identifierNameOrIDGuardConstructor(
 //	 return ackerrors.MissingNameIdentifier
 //	}
```
a-hilaly, line 848: maybe this comment needs to be updated?

pkg/generate/code/set_resource.go
```diff
@@ -1539,6 +1542,8 @@ func setResourceIdentifierPrimaryIdentifier(
 func setResourceIdentifierPrimaryIdentifierAnn(
 	cfg *ackgenconfig.Config,
 	r *model.CRD,
+	// The variable used for primary key
+	requiredFieldVarName string,
```
a-hilaly, line 1546: is this variable needed a this level? doc updates?

pkg/generate/code/set_resource.go
```diff
@@ -1451,18 +1455,18 @@ func PopulateResourceFromAnnotation(
...
-			break
 		}
 
 		targetVarPath := fmt.Sprintf("%s%s", targetVarName, memberPath)
-		if isPrimaryIdentifier {
-			primaryKeyOut += requiredFieldGuardContructor(sourceVarName, targetField.Names.CamelLower, indentLevel)
+		if inputShape.IsRequired(memberName) || isPrimaryIdentifier {
+			primaryKeyOut += requiredFieldGuardContructor(fmt.Sprintf("f%d", memberIndex), sourceVarName, targetField.Names.CamelLower, indentLevel)
 			primaryKeyOut += setResourceIdentifierPrimaryIdentifierAnn(cfg, r,
+				fmt.Sprintf("&f%d", memberIndex),
```
knottnt, line 1464: p: `fmt.Sprintf("&f%d", memberIndex)` is used in a couple places. I'd consider assigning it to a variable so that we can give it a meaningful name.

pkg/generate/code/set_resource.go
```diff
@@ -1539,6 +1544,8 @@ func setResourceIdentifierPrimaryIdentifier(
 func setResourceIdentifierPrimaryIdentifierAnn(
 	cfg *ackgenconfig.Config,
 	r *model.CRD,
+	// The variable used for primary key
+	requiredFieldVarName string,
```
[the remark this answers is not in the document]
knottnt: If it isn't always the primary key then maybe the comment should use different language.

pkg/generate/code/set_resource_test.go
```diff
@@ -1800,11 +1799,11 @@ func TestSetResource_EKS_Cluster_PopulateResourceFromAnnotation(t *testing.T) {
 	require.NotNil(crd)
 
 	expected := `
-	tmp, ok := fields["name"]
+	f0, ok := fields["name"]
```
knottnt, line 1802: Q: What's the reason for using `f0` instead of a meaningful name here? Is it to help guarantee we don't duplicate a variable name?

michaelhtm: As seen here: https://github.com/aws-controllers-k8s/eks-controller/blob/e88fa6e13ba13d3124226e321072c67b6939f978/pkg/resource/nodegroup/resource.go#L106-L115

there is a bit of an inconsistency. The main thing we're trying to do is to ensure we require the fields in the input shape. In the link shared, only the nodegroup name is being enforced by the controller, whereas in the API definition it requires the ClusterName as well: https://docs.aws.amazon.com/eks/latest/APIReference/API_DescribeNodegroup.html

Since we expect this function to be used to retrieve required keys only from the input shape, whether they are primary keys or not, we can keep the variable naming more consistent by using `f0` `f1` etc

pkg/generate/code/set_resource.go
```diff
@@ -1258,30 +1261,30 @@ func SetResourceIdentifiers(
 //
 // ```
 //
-//	tmp, ok := field["brokerID"]
+//	primaryKey, ok := field["brokerID"]
 //	if  !ok {
-//		return ackerrors.MissingNameIdentifier
```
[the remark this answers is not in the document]
a-hilaly: So for a longtime this wasn't considered a terminal error?

knottnt (approved):
@michaelhtm thanks for resolving my.  Holding of on lgtm so that @a-hilaly can verify that their comments have been resolved.

pkg/generate/code/set_resource.go
```diff
@@ -1451,18 +1456,19 @@ func PopulateResourceFromAnnotation(
 		switch targetField.ShapeRef.Shape.Type {
 		case "list", "structure", "map":
 			panic("primary identifier '" + targetField.Path + "' must be a scalar type since NameOrID is a string")
-		default:
-			break
```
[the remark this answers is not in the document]
michaelhtm: all break does is exit the switch case. since we're already at the end i thought it was unnecessary.
here's an example: https://go.dev/play/p/aFKpTVGC6BT

a-hilaly (approved):
👍 
/lgtm

PR #615, opened 2025-07-30
feat: Store resource partition in Status

michaelhtm:
Description of changes:
Store resource partition as ACKResourceMetadata

### DIFF
--- a/templates/pkg/resource/identifiers.go.tpl
+++ b/templates/pkg/resource/identifiers.go.tpl
@@ -40,3 +40,12 @@ func (ri *resourceIdentifiers) Region() *ackv1alpha1.AWSRegion {
 	}
 	return nil
 }
+
+// Partition returns the AWS partition in which the reosurce exists, or
+// nil if this information is not known. 
+func (ri *resourceIdentifiers) Partition() *ackv1alpha1.AWSPartition {
+	if ri.meta != nil {
+		return ri.meta.Partition
+	}
+	return nil
+}
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.25.0
 
 require (
 	github.com/aws-controllers-k8s/pkg v0.0.23
-	github.com/aws-controllers-k8s/runtime v0.58.1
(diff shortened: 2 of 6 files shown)


templates/pkg/resource/identifiers.go.tpl
```diff
@@ -40,3 +40,12 @@ func (ri *resourceIdentifiers) Region() *ackv1alpha1.AWSRegion {
 	}
 	return nil
 }
+
+// Partition returns the AWS partition in which the reosurce exists, or
+// nil if this information is not known.
```
knottnt, line 45: nit

```suggestion
// Partition returns the AWS partition in which the resource exists, or
// nil if this information is not known. 
```

templates/pkg/resource/manager_factory.go.tpl
```diff
@@ -30,33 +30,46 @@ func (f *resourceManagerFactory) ResourceDescriptor() acktypes.AWSResourceDescri
...
+	region ackv1alpha1.AWSRegion,
+	roleARN ackv1alpha1.AWSResourceName,
+) (acktypes.AWSResourceManager, error) {
 	f.Lock()
 	defer f.Unlock()
 
+	// We use the account ID, region, partition, and role ARN to uniquely identify a
+	// resource manager. This helps us to avoid creating multiple resource
+	// managers for the same account/region/roleARN combination.
+	rmId := fmt.Sprintf("%s/%s/%s", id, region, roleARN)
```
knottnt, line 48: This doesn't appear to actually use partition.

michaelhtm: there is no cross partition support..using it as a cache key would not be necessary

knottnt: Okay, we might want to update the comment to reflect that then.

>We use the account ID, region, **partition**, and role ARN to uniquely identify a
	// resource manager

knottnt (approved):
/lgtm

knottnt (approved):
/lgtm

PR #621, opened 2025-09-16
Add only_set_unchanged_fields config to UpdateOperationConfig

knottnt:
Issue #, if available:

Description of changes:
- Add config field only_set_unchanged_fields to UpdateOperationConfig.
- Add CRD.OnlySetChangedFieldsOnUpdate() func to check if only_set_unchanged_fields value
- Add logic in SetResource to add delta.DifferentAt() checks when only_set_unchanged_fields is set.
- Create unit test to verify new SetResource() behavior

### DIFF
--- a/pkg/config/resource.go
+++ b/pkg/config/resource.go
@@ -339,12 +339,12 @@ type UpdateOperationConfig struct {
 	// will leverage `delta.DifferentAt` to check whether a field have changed or not
 	// before including it in the update request.
 	OmitUnchangedFields bool `json:"omit_unchanged_fields"`
-	// OnlySetChangedField instructs the code generator on how to generate logic for setting
+	// OnlySetChangedFields instructs the code generator on how to generate logic for setting
 	// the value of Spec fields after a successful Update operation in the `sdkUpdate` function.
 	// If the boolean is true, the code generator uses `delta.DifferentAt` to check if a
 	// field had changed or not before setting it's value to that returned by the Update AWS API's
 	// response. Defaults to false.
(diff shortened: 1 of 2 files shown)


knottnt:
new config field is used in this mq-controller [PR](https://github.com/aws-controllers-k8s/mq-controller/pull/104) to apply delta check to setting of spec fields based on update operation response

https://github.com/aws-controllers-k8s/mq-controller/blob/dd705d9ad7d04dc14c7ab6c9ad1d2d5a3e8dd6a6/pkg/resource/broker/sdk.go#L559

pkg/generate/code/set_resource_test.go
```diff
@@ -5022,3 +5023,150 @@ func TestSetResource_WAFv2_RuleGroup_ReadOne(t *testing.T) {
...
+	g := testutil.NewModelForServiceWithOptions(t, "mq", &testutil.TestingModelOptions{
+		GeneratorConfigFile: "generator-only-set-unchanged-fields.yaml",
+	})
+	op := model.OpTypeUpdate
+
+	crd := testutil.GetCRDByName(t, g, "Broker")
+	require.NotNil(crd)
+
+	expected := `
+	if delta.DifferentAt("Spec.AuthenticationStrategy") {
```
a-hilaly, line 5040: is this lambda comparing the latest after a ReadOne from AWS with desired from K8s? At first glance i don't think so

knottnt: This is the delta passed to `sdkUpdate`. So, it should be the a comparison of the desired and latest.

rushmash91 (commented):
Thanks @knottnt !

PR #624, opened 2025-09-19
fix populateResourceFromAnnotation field setting

michaelhtm:
Description of changes:
Since we know the adoptionFields are always going to be `map[string]string`, we don't need to use 
`setResourceForScalar`, which handles all types of scalar assignments.


a-hilaly (dismissed):
👍

a-hilaly (approved):
👍 
/lgtm

PR #651, opened 2025-12-12
chore: Add Kustomization.yaml for crds

kelvin-acosta:
Description of changes:

Currently, the generated helm charts have a crds/ directory which installs CRDs the first time the helm chart is installed. The problem with the CRDs directory is that it is not managed further in the helm lifecycle, for example with upgrade commands. This creates an alternative way to install the crds using kustomize.

I had first opened a PR in https://github.com/aws-controllers-k8s/eks-controller/pull/191 but was redirected to open a PR in this repo instead. 


kelvin-acosta:
@a-hilaly @jlbutler 

Looking to see if there's anything else I can do here to push this along.

PR #668, opened 2026-02-20
Fix map mutation during iteration in renameExportable

a-hilaly:
renameExportable() deletes and inserts map keys while iterating
MemberRefs. Go may revisit newly-inserted keys, overwriting
OriginalMemberName with the exported name instead of the original.
Snapshot keys before iterating to prevent this.

### DIFF
--- a/pkg/api/passes.go
+++ b/pkg/api/passes.go
@@ -3,6 +3,7 @@ package api
 import (
 	"fmt"
 	"regexp"
+	"sort"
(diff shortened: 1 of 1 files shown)


michaelhtm (approved):
👍 
/lgtm

michaelhtm:
fixed in https://github.com/aws-controllers-k8s/code-generator/pull/666
/close

PR #678, opened 2026-03-16
Include idempotency token by default

gustavodiaz7722:
## Make `ignore_idempotency_token` opt-out instead of opt-in

Idempotency token fields (e.g. `ClientToken`) are SDK implementation details auto-filled by middleware. Previously, each controller had to explicitly set `ignore_idempotency_token: true` to exclude them from CRDs.

This PR flips the default: idempotency token fields are now excluded automatically. Controllers that need to keep them can set `include_idempotency_token: true`.

### Changes
- Renamed `IgnoreIdempotencyToken` to `IncludeIdempotencyToken` in `ResourceConfig`
- Inverted filtering logic in `model.go`
- Updated affected tests (EC2, MQ, EMRContainers, EMR Serverless)
- Added opt-in validation test
- 

### DIFF
--- a/pkg/config/resource.go
+++ b/pkg/config/resource.go
@@ -116,12 +116,14 @@ type ResourceConfig struct {
 	// TagConfig contains instructions for the code generator to generate
 	// custom code for ensuring tags
 	TagConfig *TagConfig `json:"tags,omitempty"`
-	// IgnoreIdempotencyToken instructs the code generator to automatically
-	// exclude fields that have the smithy.api#idempotencyToken trait (or the
-	// equivalent idempotencyToken trait in the v1 SDK model). These fields are
-	// SDK implementation details that are auto-filled by the SDK middleware
-	// when nil and should not be exposed in the CRD.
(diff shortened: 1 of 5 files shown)


pkg/config/resource.go
```diff
@@ -116,12 +116,14 @@ type ResourceConfig struct {
...
-	// SDK implementation details that are auto-filled by the SDK middleware
-	// when nil and should not be exposed in the CRD.
-	IgnoreIdempotencyToken bool `json:"ignore_idempotency_token,omitempty"`
+	// IncludeIdempotencyToken instructs the code generator to keep fields
+	// that have the smithy.api#idempotencyToken trait (or the equivalent
+	// idempotencyToken trait in the v1 SDK model). By default, these fields
+	// are automatically excluded because they are SDK implementation details
+	// that are auto-filled by the SDK middleware when nil and should not be
+	// exposed in the CRD. Set this to true only if you need to expose
+	// idempotency token fields in the CRD.
```
knottnt, line 125: p: It would be good to include an example of when this would be needed or clarify that this is intended to avoid breaking CRD changes in resources that already have it.

PR #679, opened 2026-03-16
fix: fail on missing CRDs for `compat-check` command

michaelhtm:
Description of changes:
This command is meant to be ran inside an ACK controller directory.
Currently it has been silently failing if it can't discover missing
CRDs. With this change, we will fail if we can't detect CRD files.

### DIFF
--- a/cmd/ack-generate/command/crd_compat_check.go
+++ b/cmd/ack-generate/command/crd_compat_check.go
@@ -70,7 +70,7 @@ func checkCRDCompat(cmd *cobra.Command, args []string) error {
 
 	if len(crdFiles) == 0 {
 		fmt.Println("No CRD files found, nothing to check.")
-		return nil
+		return fmt.Errorf("no CRD files found in %v", optCRDPaths)
 	}
 
 	hasBreaking := false


cmd/ack-generate/command/crd_compat_check.go
```diff
@@ -70,7 +70,7 @@ func checkCRDCompat(cmd *cobra.Command, args []string) error {
 
 	if len(crdFiles) == 0 {
 		fmt.Println("No CRD files found, nothing to check.")
-		return nil
+		return fmt.Errorf("no CRD files found in %v", optCRDPaths)
```
knottnt, line 73: Q: Under what scenarios was this silently skipping where we would have wanted to fail the test? Wondering if we should be performing this check against the CRDs as they exist in the baseRef? If the baseRef actually had no CRDs previously it seems like we would want to skip this check since there are no CRDs to break.

michaelhtm: This is running on the open PR, and looks back at BaseRef (main)
If we're adding a CRD, it is categorized as New.

Here's the test that was passing silently: https://github.com/aws-controllers-k8s/s3-controller/pull/212

PR #681, opened 2026-03-23
Support cross-resource references in nested map types

gustavodiaz7722:
### Problem

The ACK code generator did not support cross-resource references for fields nested inside map types. When a `references` config pointed to a field inside a map (e.g., `PhysicalTableMap.S3Source.DataSourceARN` in QuickSight DataSet), the generator would produce an error during code generation. This blocked controllers like QuickSight from using K8s-native resource references for fields within map-typed structures.

### Solution

Extended two functions in `pkg/generate/code/resource_reference.go`:

- **`iterReferenceValues()`** — Added a `"map"` case to the shape type switch. The new case mirrors the existing `"list"` case: it generates a `for key, val := range` loop over map entries, tracks iteration depth, and advances the field access prefix through the map value. Previously this function only handled `"structure"` and `"list"` shapes, returning an error for maps.

- **`buildIndexBasedFieldAccessorWithOffset()`** — Extended the list-shape condition to also match `"map"` shapes (`cur.ShapeRef.Shape.Type == "list" || cur.ShapeRef.Shape.Type == "map"`). This ensures index-based field accessors are correctly built when traversing map entries during reference resolution and clearing.

### Testing

- Added a QuickSight test fixture (`generator-with-nested-references.yaml`) that configures `PhysicalTableMap.CustomSql.DataSourceArn` as a cross-resource reference to DataSource.
- Added 3 unit tests:
  - `Test_ResolveReferencesForField_SingleReference_WithinMap` — verifies reference resolution code generation for map-nested fields
  - `Test_ClearResolvedReferencesForField_SingleReference_WithinMap` — verifies reference clearing code generation
  - `Test_ReferenceFieldsValidation_MapNestedReference` — verifies field validation accepts map-nested references
- All 113 existing tests continue to pass (backward compatible).

### Affected resources

Any ACK controller with references inside map-typed fields. Immediate use case: QuickSight DataSet's `PhysicalTableMap` entries referencing DataSource ARNs.

### DIFF
--- a/pkg/generate/code/resource_reference_test.go
+++ b/pkg/generate/code/resource_reference_test.go
@@ -568,3 +568,103 @@ func Test_ClearResolvedReferencesForField_SingleReference_WithinMultipleSlices(t
 	require.NoError(err)
 	assert.Equal(expected, got)
 }
+
+func Test_ResolveReferencesForField_SingleReference_WithinMap(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+
+	g := testutil.NewModelForServiceWithOptions(t, "quicksight",
+		&testutil.TestingModelOptions{
+			GeneratorConfigFile: "generator-with-nested-references.yaml",
+		})
+
+	crd := testutil.GetCRDByName(t, g, "DataSet")
+	require.NotNil(crd)
+
+	expected :=
+		`	for f0idx, f0iter := range ko.Spec.PhysicalTableMap {
+		if f0iter.CustomSQL != nil {
+			if f0iter.CustomSQL.DataSourceRef != nil && f0iter.CustomSQL.DataSourceRef.From != nil {
+				hasReferences = true
+				arr := f0iter.CustomSQL.DataSourceRef.From
+				if arr.Name == nil || *arr.Name == "" {
+					return hasReferences, fmt.Errorf("provided resource reference is nil or empty: PhysicalTableMap.CustomSQL.DataSourceRef")
+				}
+				namespace := ko.ObjectMeta.GetNamespace()
+				if arr.Namespace != nil && *arr.Namespace != "" {
+					namespace = *arr.Namespace
+				}
+				obj := &svcapitypes.DataSet{}
+				if err := getReferencedResourceState_DataSet(ctx, apiReader, obj, *arr.Name, namespace); err != nil {
+					return hasReferences, err
+				}
+				ko.Spec.PhysicalTableMap[f0idx].CustomSQL.DataSourceARN = (*string)(obj.Status.ACKResourceMetadata.ARN)
+			}
+		}
+	}
+`
+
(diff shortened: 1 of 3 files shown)


pkg/generate/code/resource_reference_test.go
```diff
@@ -568,3 +568,103 @@ func Test_ClearResolvedReferencesForField_SingleReference_WithinMultipleSlices(t
...
+	require.NotNil(crd)
+
+	expected :=
+		`	for f0idx, f0iter := range ko.Spec.PhysicalTableMap {
+		if f0iter.CustomSQL != nil {
+			if f0iter.CustomSQL.DataSourceRef != nil && f0iter.CustomSQL.DataSourceRef.From != nil {
+				hasReferences = true
+				arr := f0iter.CustomSQL.DataSourceRef.From
+				if arr.Name == nil || *arr.Name == "" {
+					return hasReferences, fmt.Errorf("provided resource reference is nil or empty: PhysicalTableMap.CustomSQL.DataSourceRef")
```
knottnt, line 591: Q: should we include the map key in this error message? If there are multiple entries of the same object type it would help users identify the problematic part of their spec.

pkg/generate/code/resource_reference_test.go
```diff
@@ -568,3 +568,103 @@ func Test_ClearResolvedReferencesForField_SingleReference_WithinMultipleSlices(t
...
+	g := testutil.NewModelForServiceWithOptions(t, "quicksight",
+		&testutil.TestingModelOptions{
+			GeneratorConfigFile: "generator-with-nested-references.yaml",
+		})
+
+	crd := testutil.GetCRDByName(t, g, "DataSet")
+	require.NotNil(crd)
+
+	expected :=
+		`	for f0idx, f0iter := range ko.Spec.PhysicalTableMap {
```
knottnt: nit: It looks like we're re-using the list variables names here. Changing the name to be map specific could make the generated code more readable. Example names would be `f0key` and `f0value`.

knottnt (commented):
Thanks @gustavodiaz7722. Left a couple comments. Also have you been able to test the generated code locally with the quicksight-controller?

gustavodiaz7722:
@knottnt Yes, I have tested these changes locally with new e2e test for verifying cross referencing is working correctly. I haven't updated the existing open PR with the new test yet, have them locally staged.

pkg/generate/code/resource_reference_test.go
gustavodiaz7722: I think for now we can leave as is, since it follows similar behavior for list type references. I agree though it's poor UX if we don't include key/element identifier. 

For now I think it's okay to leave as is. I can follow up on this in another PR if that's okay

pkg/generate/code/resource_reference_test.go
gustavodiaz7722: Yup, agreed. Will change this in new revision.

michaelhtm (approved):

PR #683, opened 2026-03-27
Handle double and long shape types in setSDKForScalar for list members

gustavodiaz7722:
### Issue

The QuickSight controller analysis resource fails to compile with errors like:

```
pkg/resource/analysis/sdk.go:6223:30: cannot use f2f6elemf1f0f1iter
  (variable of type *float64) as float64 value in assignment
```

This affects all `[]float64` list fields in the CRD-to-SDK conversion path, producing 10+ compile errors across the Analysis resource's `sdk.go`.

### Root Cause

`setSDKForScalar` in `pkg/generate/code/set_sdk.go` has an `intOrFloat` map that handles narrowing conversions for `"integer"` (int64→int32) and `"float"` (float64→float32) shape types. These types need range checks and casts because the CRD uses 64-bit types while the SDK uses 32-bit types.

However, `"double"` (float64) and `"long"` (int64) shape types were not handled. When these appear as list members:

- CRD type: `[]*float64` — loop iterator is `*float64`
- SDK type: `[]float64` — target element is `float64`

The code fell through to the default `else` branch which strips the `*` dereference via `strings.TrimPrefix(setTo, "*")`, generating `elem = iter` instead of `elem = *iter`.

### Fix

Added an explicit branch for `"double"` and `"long"` types in `setSDKForScalar`. Unlike `"integer"`/`"float"`, these are already native 64-bit Go types, so no range check or narrowing cast is needed — just correct pointer dereference in list context.

### DIFF
--- a/pkg/generate/code/set_sdk.go
+++ b/pkg/generate/code/set_sdk.go
@@ -1765,6 +1765,16 @@ func setSDKForScalar(
 		}
 		out += fmt.Sprintf("%stemp := int32(%s)\n", indent, setTo)
 		out += fmt.Sprintf("%s%s = &temp\n", indent, targetVarPath)
+	} else if shape.Type == "double" || shape.Type == "long" {
+		targetVarPath := targetVarName
+		if targetFieldName != "" {
+			targetVarPath += "." + targetFieldName
+		}
+		if isListMember {
+			out += fmt.Sprintf("%s%s = %s\n", indent, targetVarPath, setTo)
+		} else {
+			out += fmt.Sprintf("%s%s = %s\n", indent, targetVarPath, strings.TrimPrefix(setTo, "*"))
+		}
 	} else {
 
 		targetVarPath := targetVarName
--- a/pkg/generate/code/set_sdk_test.go
+++ b/pkg/generate/code/set_sdk_test.go
@@ -6656,3 +6656,55 @@ func TestSetSDK_QuickSight_DataSet_Update(t *testing.T) {
 	// should also use value-type access
 	assert.NotContains(got, "= &f6elemf0f0f0iter.Time")
 }
+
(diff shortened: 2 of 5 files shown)


pkg/generate/code/set_sdk.go
```diff
@@ -1765,6 +1765,16 @@ func setSDKForScalar(
...
+	} else if shape.Type == "double" || shape.Type == "long" {
+		targetVarPath := targetVarName
+		if targetFieldName != "" {
+			targetVarPath += "." + targetFieldName
+		}
+		if isListMember {
+			out += fmt.Sprintf("%s%s = %s\n", indent, targetVarPath, setTo)
+		} else {
+			out += fmt.Sprintf("%s%s = %s\n", indent, targetVarPath, strings.TrimPrefix(setTo, "*"))
+		}
```
michaelhtm, line 1777: how are double and long handled in `setResource`?

gustavodiaz7722: I missed checking setResource direction for double and long, since quicksight Analysis resource did not have any compile errors on translating sdk to crd.

I would assume for setResource we would do an opposite operation, where we would add prefix & for float and double.

gustavodiaz7722: I think for now we can leave setResource alone, since there doesn't appear to be any compile issues.

michaelhtm: can we support it in this PR, if it's not already supported?

gustavodiaz7722: I think the problem with supporting this right now is we don't have any concrete examples to test this feature in setResource. I guess we could create a white box test to simulate this.

michaelhtm (approved):
/lgtm

PR #684, opened 2026-04-02
feat: add new_resource_manager_client_options hook to manager template

tcsavage:
Issue #, if available: aws-controllers-k8s/community#2237

Description of changes:

In order to support some S3-compatible APIs (e.g. LocalStack on a local dev cluster) it may be necessary to force the S3 client to use "path-style" URLs (e.g. http://host/bucket/key).

This PR adds a hook injection point in `newResourceManager` to allow service controllers to pass custom options to the AWS SDK client constructor.

This will let the S3 controller set `UsePathStyle` based on the value set in its controller config.

I've also added changes to the Helm chart to support configuring this option in the S3 controller.

Associated PRs:
- [s3-controller#214](https://github.com/aws-controllers-k8s/s3-controller/pull/214) uses this hook to control `UsePathStyle`.
- [runtime#233](https://github.com/aws-controllers-k8s/runtime/pull/233) adds a new config option used by the `s3-controller` to set `UsePathStyle`.

### DIFF
--- a/templates/helm/values.schema.json.tpl
+++ b/templates/helm/values.schema.json.tpl
@@ -181,10 +181,12 @@
           "type": "boolean",
           "default": false
         },
+        {{- if eq .ControllerName "s3" }}
         "endpoint_use_path_style": {
           "type": "boolean",
           "default": false
         },
+        {{- end }}
         "credentials": {
           "description": "AWS credentials information",
           "properties": {
--- a/pkg/generate/ack/release.go
+++ b/pkg/generate/ack/release.go
@@ -32,7 +32,7 @@ var (
 		"helm/templates/_helpers.tpl.tpl",
 		"helm/Chart.yaml.tpl",
 		"helm/values.yaml.tpl",
-		"helm/values.schema.json",
+		"helm/values.schema.json.tpl",
 		"helm/templates/NOTES.txt.tpl",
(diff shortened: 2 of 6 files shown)


templates/helm/values.schema.json.tpl
```diff
@@ -181,6 +181,10 @@
           "type": "boolean",
           "default": false
         },
+        "endpoint_use_path_style": {
+          "type": "boolean",
+          "default": false
+        },
```
knottnt, line 188: It seems like this should only be included in the s3-controller's values.schema.json if we're omitting the corresponding value in the deployment template for our other controllers. To make it clear that this file is being templated could you also change the filename to `values.schema.json.tpl` and update the corresponding [entry](https://github.com/aws-controllers-k8s/code-generator/blob/c55e8227b1ff5ac6a8b8e1421a2f2076fa6a77bd/pkg/generate/ack/release.go#L35) in `release.go`. 
```suggestion
        {{- if eq .ControllerName "s3" }}
        "endpoint_use_path_style": {
          "type": "boolean",
          "default": false
        },
        {{- end }}
```

knottnt (commented):
Thanks @tcsavage! Just a left a comment. Overall looks good.

knottnt (approved):
/lgtm

PR #685, opened 2026-04-06
Add updateable and deletable status guards for resources

sapphirew:
Issue #, if available:

Description of changes:

Add generator.yaml configuration and template functions that allow service controllers to declare which status field values permit update and delete operations. When a resource is not in an allowed state, the generated code requeues the operation with a configurable delay instead of attempting the API call.

New config blocks `updateable` and `deletable` on ResourceConfig accept a list of status field path + allowed-values conditions and an optional requeue_after_seconds override (default 30s). Template functions GoCodeResourceIsUpdateable and GoCodeResourceIsDeletable are injected into sdk_update, sdk_update_custom, sdk_update_set_attributes, and sdk_delete templates respectively.

### DIFF
--- a/pkg/generate/code/resource_guard.go
+++ b/pkg/generate/code/resource_guard.go
@@ -1,0 +1,188 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package code
+
+import (
+	"fmt"
+	"strings"
+
+	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/config"
+	"github.com/aws-controllers-k8s/code-generator/pkg/model"
+)
+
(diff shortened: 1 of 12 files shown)


pkg/generate/code/updateable.go
knottnt, line 1: nit: If both updatable and deletable checks are in this file a more generic name might be better.

templates/pkg/resource/sdk_update_custom.go.tpl
```diff
@@ -5,6 +5,7 @@ func (rm *resourceManager) sdkUpdate(
 	latest *resource,
 	delta *ackcompare.Delta,
 ) (*resource, error) {
+{{ GoCodeResourceIsUpdateable .CRD "latest" 1 }}
```
knottnt, line 8: I think this can be removed. This template is for using a fully custom update function. So, combining it with generated code would be unexpected.

knottnt (commented):
Thanks @sapphirew. Just left a couple comments. Also did you have a chance to test these new generator.yaml configs on an existing controller?

pkg/generate/code/resource_guard.go
sapphirew: refactored to resource_guard.go

templates/pkg/resource/sdk_update_custom.go.tpl
sapphirew: makes sense, removed

sapphirew:
> Thanks @sapphirew. Just left a couple comments. Also did you have a chance to test these new generator.yaml configs on an existing controller?

@knottnt Thanks for reviewing this PR. Here's an example for ec2-controller: https://github.com/sapphirew/ack-ec2-controller/pull/1/changes

knottnt (approved):
Thanks @sapphirew!
/lgtm

PR #686, opened 2026-04-08
feat: add CompareResourceForPreDelete and DeltaForPreDelete support

sapphirew:
Add pre-delete delta comparison support to the code generator. Controllers can opt specific fields into pre-delete comparison via `compare.pre_delete_include: true` in generator.yaml, so that fields like `DeletionProtectionEnabled` are synced before deletion even when they use `compare.is_ignored` for normal reconciliation.

Config changes:
- Add `PreDeleteInclude` bool to `CompareFieldConfig` (field.go)
- Add `PreDeleteSyncConfig` with `CompareAll to ResourceConfig` (resource.go)

Code generation:
- Add `CompareResourceForPreDelete` function that emits comparison code
  only for opted-in fields (or all fields when compare_all is true)
- Register `GoCodeCompareForPreDelete` template function (controller.go)
- Add `newResourceDeltaForPreDelete` in delta.go.tpl
- Add `DeltaForPreDelete` method on `resourceDescriptor` in descriptor.go.tpl

Tests:
- Four tests covering no-opt-in (empty), explicit opt-in, compare_all,
  and empty-without-opt-in scenarios
- Two new S3 test fixture generator YAMLs

runtime changes: https://github.com/aws-controllers-k8s/runtime/pull/234

Issue #, if available:

Description of changes:

### DIFF
--- a/pkg/generate/code/compare.go
+++ b/pkg/generate/code/compare.go
@@ -988,3 +988,212 @@ func fastCompareTypes(
 	}
 	return out, needToCloseBlock, nil
 }
+
+// CompareResourceForPreDelete returns the Go code that traverses a set of two
+// Resources, adding differences between the two Resources to an
+// `ackcompare.Delta`.
+//
+// When the resource has `pre_delete_sync.compare_all: true` set, ALL fields
+// are included (even those with `compare.is_ignored: true`), mirroring
+// CompareResource without the is_ignored skip.
+//
+// Otherwise (the default), only fields explicitly opted in via
+// `compare.pre_delete_include: true` are included; all other fields are
+// skipped. This opt-in approach reduces blast radius so controllers only get
+// pre-delete comparison on fields they've deliberately configured.
+func CompareResourceForPreDelete(
+	cfg *ackgenconfig.Config,
+	r *model.CRD,
+	// String representing the name of the variable that is of type
+	// `*ackcompare.Delta`. We will generate Go code that calls the `Add()`
+	// method of this variable when differences between fields are detected.
+	deltaVarName string,
+	// String representing the name of the variable that represents the first
+	// CR under comparison. This will typically be something like "a.ko". See
+	// `templates/pkg/resource/delta.go.tpl`.
+	firstResVarName string,
+	// String representing the name of the variable that represents the second
+	// CR under comparison. This will typically be something like "b.ko". See
+	// `templates/pkg/resource/delta.go.tpl`.
+	secondResVarName string,
+	// Number of levels of indentation to use
+	indentLevel int,
+) (string, error) {
+	out := "\n"
+
+	resConfig := cfg.GetResourceConfig(r.Names.Camel)
+
+	// Determine whether to include all fields in the pre-delete delta.
+	// If compare_all is set at the resource level, include all fields
+	// (even is_ignored ones). Otherwise, only include fields explicitly
+	// opted in via compare.pre_delete_include: true.
+	compareAll := false
+	if resConfig != nil && resConfig.PreDeleteSync != nil {
+		compareAll = resConfig.PreDeleteSync.CompareAll
+	}
+
+	tagField, err := r.GetTagField()
+	if err != nil {
+		return "", err
+	}
+
+	// We need a deterministic order to traverse our top-level fields...
+	specFieldNames := []string{}
+	for fieldName := range r.SpecFields {
+		specFieldNames = append(specFieldNames, fieldName)
+	}
+	sort.Strings(specFieldNames)
+
+	for _, fieldName := range specFieldNames {
+		specField := r.SpecFields[fieldName]
+		indent := strings.Repeat("\t", indentLevel)
+		firstResAdaptedVarName := firstResVarName + cfg.PrefixConfig.SpecField
+		firstResAdaptedVarName += "." + specField.Names.Camel
+		secondResAdaptedVarName := secondResVarName + cfg.PrefixConfig.SpecField
+		secondResAdaptedVarName += "." + specField.Names.Camel
+
+		var fieldConfig *ackgenconfig.FieldConfig
+		var compareConfig *ackgenconfig.CompareFieldConfig
+
+		if resConfig != nil {
+			fieldConfig = resConfig.GetFieldConfig(fieldName)
+		}
+		if fieldConfig != nil {
+			compareConfig = fieldConfig.Compare
+		}
+
+		if !compareAll {
+			// Default opt-in behavior: only include fields with pre_delete_include
+			if compareConfig == nil || !compareConfig.PreDeleteInclude {
+				continue
+			}
+		}
+
+		// this is the "path" to the field within the structs being compared.
+		// This is passed down into the compareXXX functions recursively and
+		// appended to with each level of nested structs we recurse into.
+		fieldPath := strings.TrimPrefix(
+			cfg.PrefixConfig.SpecField+"."+specField.Names.Camel, ".",
+		)
+
+		// Use equality.Semantic.Equalities.DeepEqual for comparing Reference fields because
+		// some of reference fields are list of pointer to structs and
+		// DeepEqual is easy way to compare them
+		if specField.IsReference() {
+			out += fmt.Sprintf("%sif !equality.Semantic.Equalities.DeepEqual(%s, %s) {\n",
+				indent, firstResAdaptedVarName, secondResAdaptedVarName)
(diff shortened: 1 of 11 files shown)


pkg/generate/code/compare.go
```diff
@@ -988,3 +988,346 @@ func fastCompareTypes(
 	}
 	return out, needToCloseBlock, nil
 }
+
+// CompareResourceForPreDelete returns the Go code that traverses a set of two
+// Resources, adding differences between the two Resources to an
+// `ackcompare.Delta`. Unlike CompareResource, this function does NOT skip
+// fields where compareConfig.IsIgnored is true. This is used for pre-delete
+// sync to ensure fields like DeletionProtectionEnabled are included in the
+// delta comparison.
```
michaelhtm: should we make this delta comparison opt-in? Only compute delta on specific fields defined in generator.yaml?
In dsql example, we could just compare DeletionProtection field only

michaelhtm (commented):
Thanks @sapphirew 
left one comment below

pkg/generate/code/compare.go
sapphirew: good call, that would provide fine-control on specific fields for a controller. 

The current delta comparison can be enabled by a flag like this instead:
```
resources:
  Cluster:
    pre_delete_sync:
      compare_all: true
```

pkg/generate/code/compare.go
```diff
@@ -988,3 +988,273 @@ func fastCompareTypes(
...
+	}
+
+	// We need a deterministic order to traverse our top-level fields...
+	specFieldNames := []string{}
+	for fieldName := range r.SpecFields {
+		specFieldNames = append(specFieldNames, fieldName)
+	}
+	sort.Strings(specFieldNames)
+
+	for _, fieldName := range specFieldNames {
```
knottnt: p: The logic for creating the go comparison code appears to be almost exactly the same as `CompareResource` which makes sense. However, I'm wondering if instead of copying the code we could extract it into a shared function and split out the logic for field selection. That way we don't have to maintain the comparison generation logic in two places.

pkg/generate/code/compare.go
```diff
@@ -988,3 +988,273 @@ func fastCompareTypes(
...
+	compareAll := false
+	if resConfig != nil && resConfig.PreDeleteSync != nil {
+		compareAll = resConfig.PreDeleteSync.CompareAll
+	}
+
+	specFieldNames := []string{}
+	for fieldName := range r.SpecFields {
+		specFieldNames = append(specFieldNames, fieldName)
+	}
+	sort.Strings(specFieldNames)
```
knottnt: p: This logic appears in at least three places. Might be worth moving this to a utility function for better re-usability.

templates/pkg/resource/descriptor.go.tpl
```diff
@@ -59,6 +59,20 @@ func (d *resourceDescriptor) Delta(a, b acktypes.AWSResource) *ackcompare.Delta
...
+// copy of b (observed) with only those differing fields overwritten from a
+// (desired). Used during pre-delete sync.
+func (d *resourceDescriptor) DeltaForPreDelete(
+	a, b acktypes.AWSResource,
+) (*ackcompare.Delta, acktypes.AWSResource) {
+	delta, merged := newResourceDeltaForPreDelete(a.(*resource), b.(*resource))
+	if merged == nil {
+		return delta, nil
+	}
+	return delta, merged
```
knottnt: Q: If `merged` is nil already why do we need to change the return to explicitly return nil?

knottnt (commented):
@sapphirew left a few comments. Also, it would be helpful to see a WIP PR showing off these changes applied to a service controller. It makes it a bit easier to understand how the generated code will actually work.

pkg/generate/code/compare.go
sapphirew: Sure, extracted to `compareResourceFields` function

pkg/generate/code/compare.go
sapphirew: Extracted to `sortedSpecFieldNames` helper function

templates/pkg/resource/descriptor.go.tpl
sapphirew: Good call, removed this if condition

pkg/generate/code/compare.go
```diff
@@ -129,20 +80,14 @@ func CompareResource(
 			compareConfig = fieldConfig.Compare
 		}
 
-		if compareConfig != nil && compareConfig.IsIgnored {
+		if !includeField(compareConfig) {
 			continue
 		}
 
-		// this is the "path" to the field within the structs being compared.
```
knottnt, line 136: These removed comments still seem to be relevant. Is there a reason for removing them?

templates/pkg/resource/descriptor.go.tpl
```diff
@@ -59,6 +59,17 @@ func (d *resourceDescriptor) Delta(a, b acktypes.AWSResource) *ackcompare.Delta
     return newResourceDelta(a.(*resource), b.(*resource))
 }
 
+// DeltaForPreDelete returns an `ackcompare.Delta` that includes fields
+// configured for pre-delete comparison, and a merged resource that is a deep
+// copy of b (observed) with only those differing fields overwritten from a
+// (desired). Used during pre-delete sync.
+func (d *resourceDescriptor) DeltaForPreDelete(
```
knottnt, line 67: The associated runtime appears to assume that this function is optional and performs a type check to exit the pre-delete sync early if the resource doesn't implement this function. However, from this template it looks like this function will always be implemented regardless of whether the resource includes a pre-delete synced field. What's the purpose of that runtime type check?

https://github.com/sapphirew/ack-runtime/blob/a2382caa3636d1cc7364c76e2812c854dab3c561/pkg/runtime/reconciler.go#L1108

CODE_GENERATOR_FIX_BLOB_UNION.md
knottnt, line 1: This can be removed

templates/pkg/resource/descriptor.go.tpl
sapphirew: Good catch, actually it should be optional. Will conditionally generate this function

pkg/generate/code/compare.go
sapphirew: got dropped during the refactor. Added them back

CODE_GENERATOR_FIX_BLOB_UNION.md
sapphirew: oops, removed

pkg/generate/code/compare_test.go
```diff
@@ -669,6 +669,130 @@ func TestCompareResource_IAM_Role_IAMPolicy(t *testing.T) {
...
+	// With pre_delete_include: true on ACL, only ACL should appear in the
+	// pre-delete delta output (other fields are not opted in).
+	expected := `
+	if ackcompare.HasNilDifference(a.ko.Spec.ACL, b.ko.Spec.ACL) {
+		delta.Add("Spec.ACL", a.ko.Spec.ACL, b.ko.Spec.ACL)
+	} else if a.ko.Spec.ACL != nil && b.ko.Spec.ACL != nil {
+		if *a.ko.Spec.ACL != *b.ko.Spec.ACL {
+			delta.Add("Spec.ACL", a.ko.Spec.ACL, b.ko.Spec.ACL)
+		}
+	}
```
michaelhtm, line 796: can we add test cases for nested fields?

pkg/generate/code/compare.go
```diff
@@ -988,3 +1041,107 @@ func fastCompareTypes(
...
+
+// MergeResourceForPreDelete outputs Go code that copies only the pre-delete sync
+// configured fields from the source variable onto the target variable. This is
+// used to build a merged resource that starts as a deep copy of observed and
+// has only the pre-delete sync fields overwritten from desired.
+//
+// The generated code is a series of simple field assignments for each field
+// that has `compare.pre_delete_include: true` (or all fields when
+// `pre_delete_sync.compare_all: true`).
+func MergeResourceForPreDelete(
```
michaelhtm, line 1193: can we add a unit test for this one?

sapphirew: yes, added

pkg/generate/code/compare_test.go
sapphirew: yes, added

knottnt (approved):

knottnt (approved):
/lgtm

PR #688, opened 2026-04-13
Fix OriginalShapeName handling for union types

jlbutler:
Came across an issue w/ handling of `OriginalShapeName` for union types. 

PR #657  fixed the Input/Output suffix renaming issue for structure shapes by using OriginalShapeName to reference the correct SDK type name. However, the fix was not applied to union type code paths.

When a union shape name ends in "Input", e.g. `FooInput`, code-gen renames it w/ an underbar, e.g. `FooInput_` to avoid collisions with operation input types. The structure fix restores the original name for SDK references, but `setSDKForUnion`, `setResourceForUnion`, and `varEmptyConstructorSDKType` (when `RealType` == "union") were not covered.

Discovered while building the AgentCore `Memory` resource (https://github.com/aws-controllers-k8s/bedrockagentcorecontrol-controller/pull/16), which has several union types ending in "Input" (`MemoryStrategyInput`, `CustomConfigurationInput`, `TriggerConditionInput`).

### DIFF
--- a/pkg/api/passes_test.go
+++ b/pkg/api/passes_test.go
@@ -974,3 +974,91 @@ func TestValidateShapeNameMethod(t *testing.T) {
 		})
 	}
 }
+
+// TestRenamedUnionShapePreservesOriginalName verifies that when a union
+// shape is renamed (e.g. MemoryStrategyInput -> MemoryStrategyInput_) by
+// renameIOSuffixedShapeNames, the OriginalShapeName is preserved. The
+// downstream SDK code generation functions (setSDKForUnion,
+// setResourceForUnion, varEmptyConstructorSDKType) use OriginalShapeName
+// to emit correct SDK type references.
+func TestRenamedUnionShapePreservesOriginalName(t *testing.T) {
+	a := &API{
+		name: "testapi",
+		Metadata: Metadata{
+			APIVersion:          "0000-00-00",
+			EndpointPrefix:      "testapi",
+			JSONVersion:         "1.1",
+			Protocol:            "json",
+			ServiceAbbreviation: "TestAPI",
+			ServiceFullName:     "Test API",
+			SignatureVersion:    "v4",
+		},
+		Operations: map[string]*Operation{
+			"CreateMemory": {
+				Name:      "CreateMemory",
+				InputRef:  ShapeRef{ShapeName: "CreateMemoryInput"},
+				OutputRef: ShapeRef{ShapeName: "CreateMemoryOutput"},
+			},
+		},
+		Shapes: map[string]*Shape{
+			"CreateMemoryInput": {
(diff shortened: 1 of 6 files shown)


knottnt (approved):
/lgtm

jlbutler:
I went down the code paths but failed to see that `goType` is used for both SDK type references AND CRD type generation. By using `OriginalShapeName` unconditionally, we're also changing the CRD type names, which breaks the stutter renaming.  So, we were hitting this for any rename (stutter removal included). We only need this for the suffix rename case we hit.

`goType` is the wrong place for this change, needs to be downstream only in cases we know we're in SDK-specific code paths where we are generating type references (not CRD types). Refreshing PR to:
- add union in set_sdk.go (downstream)
- add back in the initial `OriginalShapeName` override from first pass

knottnt (commented):
@jlbutler could we add test cases in set_resource_test.go and set_sdk_test.go to validate that the SDK type with an `Input` suffix is properly handled. I know we've tested in a WIP branch for the bedrockagentcorecontrol-controller, but having the tests will help us avoid a regression for that generated code. 

We'd mostly just need to validate that the expected line appears in the generated code and that the `Input_` version is not present. @gustavodiaz7722 made some good examples of this in the quicksight tests he added.

[set_resource.go test](https://github.com/aws-controllers-k8s/code-generator/blob/60fc9a5dbc6357d163d96c64e498efa46e4955a4/pkg/generate/code/set_resource_test.go#L5447)
[set_sdk.go test](https://github.com/aws-controllers-k8s/code-generator/blob/60fc9a5dbc6357d163d96c64e498efa46e4955a4/pkg/generate/code/set_sdk_test.go#L6582)

#689 already added test data for bedrockagentcorecontrol, but you'll likely need to add a new test generator.yaml [here](https://github.com/aws-controllers-k8s/code-generator/tree/main/pkg/testdata/models/apis/bedrock-agentcore-control/0000-00-00). You can find an example of a similar bedrock-agentcore-control test [here](https://github.com/aws-controllers-k8s/code-generator/blob/60fc9a5dbc6357d163d96c64e498efa46e4955a4/pkg/generate/code/set_sdk_test.go#L6726).

knottnt (approved):
/lgtm

knottnt:
DynamoDB failure is related to change in API validation.

```
operation error DynamoDB: CreateGlobalTable, https response error StatusCode: 400, RequestID: IGOCMORBBUBNR768N664UIHKMFVV4KQNSO5AEMVJF66Q9ASUAAJG, api error ValidationException: One or more parameter values were invalid: DynamoDB global tables version 2017.11.29 is not supported. We recommend using DynamoDB global tables version 2019.11.21, instead of version 2017.11.29 (Legacy)."}
```

PR #689, opened 2026-04-13
Supported nested union fields

knottnt:
Issue #, if available:

Description of changes:
This PR adds support for nested union types to the code-generator. Prior to this change if a `union` shape contained a member that was itself a `union` the code-generator would produce non-compiling code in two ways:

1. When setting the values from the API response the generated code would not use a type switch for the nested union member and attempt directly access fields as if they were a `structure`.
2. When setting the SDK payload the generated code would attempt to dereference the union member.

To address this this PR applies two fixes.
1. In `setResourceForUnion` check if the member is a union and force `Shape.Type` to `union` before recursing
2. In `setSDKForUnion` prevent dereference of member value if member is a union itself.


jlbutler:
This PR addresses the other issues I came across w/ Bedrock AgentCore - the first one is addressed in PR #688

pkg/generate/code/set_sdk.go
```diff
@@ -1937,7 +1937,7 @@ func setSDKForUnion(
 					return "", err
 				}
 				out += containerOut
-				if memberShape.Type == "list" {
+				if memberShape.Type == "list" || memberShape.Type == "union" {
```
jlbutler, line 1944: this addresses the deref issue i was hitting, nice.

i guess one nit - this depends on us earlier setting `Type` above (line ~1895). if we ever refactor in the future, this is maybe not obvious. worth a comment linking this?

PR #690, opened 2026-04-16
feat(sdk): Add per-service SDK tag fallback for model fetching

sapphirew:
Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/2859
Description of changes:

When a new AWS service is released, its model JSON is published under a per-service tag (e.g. service/s3files/v1.0.0) before the next core SDK tag is cut. This adds automatic fallback to per-service tags when the core tag returns HTTP 404.

Changes:
- Extend EnsureModel with per-service URL fallback and separate cache
- Add --aws-service-sdk-version CLI flag to ack-generate
- Add version resolution priority chain (flag > metadata > empty)
- Record aws_service_sdk_version in ack-generate-metadata.yaml
- Add AWS_SERVICE_SDK_VERSION passthrough in build-controller.sh
- Add unit tests for fallback, cache separation, URL construction, metadata round-trip, and EnsureSemverPrefix idempotence

### DIFF
--- a/cmd/ack-generate/command/apis.go
+++ b/cmd/ack-generate/command/apis.go
@@ -64,7 +64,8 @@ func saveGeneratedMetadata(cmd *cobra.Command, args []string) error {
 		optGenVersion,
 		filepath.Join(optOutputPath, "apis"),
 		ackmetadata.UpdateReasonAPIGeneration,
-		optAWSSDKGoVersion,
+		sdkVersion,
+		svcSDKVersion,
 		optGeneratorConfigPath,
 	)
 	if err != nil {
--- a/cmd/ack-generate/command/root.go
+++ b/cmd/ack-generate/command/root.go
@@ -33,6 +33,7 @@ var (
 	defaultCacheDir            string
 	optCacheDir                string
 	optAWSSDKGoVersion         string
+	optAWSServiceSDKVersion    string
 	defaultTemplateDirs        []string
 	optTemplateDirs            []string
 	defaultServicesDir         string
@@ -123,6 +124,9 @@ func init() {
 	)
 	rootCmd.PersistentFlags().StringVar(
 		&optAWSSDKGoVersion, "aws-sdk-go-version", "", "Version of github.com/aws/aws-sdk-go-v2 used to generate apis and controllers files",
+	)
+	rootCmd.PersistentFlags().StringVar(
+		&optAWSServiceSDKVersion, "aws-service-sdk-version", "", "Per-service SDK version (e.g. v1.0.0) for fetching model from per-service tag",
 	)
 	rootCmd.PersistentFlags().StringVar(
 		&optServiceAccountName, "service-account-name", "", "The name of the ServiceAccount used for ACK service controller",
--- a/pkg/metadata/generation_metadata_test.go
+++ b/pkg/metadata/generation_metadata_test.go
@@ -1,0 +1,124 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package metadata
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/ghodss/yaml"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestGenerationMetadata_RoundTrip validates that serializing GenerationMetadata
+// to YAML and deserializing back produces identical field values, including the
+// aws_service_sdk_version field when present and its absence when empty.
+func TestGenerationMetadata_RoundTrip(t *testing.T) {
+	tests := []struct {
+		name                 string
+		awsServiceSDKVersion string
+		expectFieldInYAML    bool
+	}{
+		{
+			name:                 "with per-service SDK version",
+			awsServiceSDKVersion: "v1.0.0",
+			expectFieldInYAML:    true,
+		},
+		{
+			name:                 "without per-service SDK version",
+			awsServiceSDKVersion: "",
+			expectFieldInYAML:    false,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			original := GenerationMetadata{
+				APIVersion:           "v1alpha1",
+				APIDirectoryChecksum: "abc123",
+				LastModification: lastModificationInfo{
+					Reason: UpdateReasonAPIGeneration,
+				},
+				AWSSDKGoVersion:      "v1.41.5",
+				AWSServiceSDKVersion: tc.awsServiceSDKVersion,
+				ACKGenerateInfo: ackGenerateInfo{
+					Version:   "v0.58.0",
+					GoVersion: "go1.25.3",
+					BuildDate: "2026-04-14T18:02:39Z",
+					BuildHash: "a9e2cea",
+				},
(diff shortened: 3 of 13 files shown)


pkg/sdk/fetch.go
```diff
@@ -26,86 +26,174 @@ import (
...
-	// Fetch from GitHub
-	url := fmt.Sprintf(sdkModelURLTemplate, sdkVersion, modelName)
-	util.Tracef("EnsureModel: fetching %s\n", url)
+	// Fetch from GitHub (core tag)
+	coreURL := fmt.Sprintf(sdkModelURLTemplate, sdkVersion, modelName)
+	util.Tracef("EnsureModel: fetching %s\n", coreURL)
 
 	fetchStart := time.Now()
+	coreStatus, coreBody, err := httpGet(ctx, coreURL)
+	if err != nil {
```
knottnt, line 143: Q: Does this not throw an error when no `sdkVersion` is set?

pkg/sdk/fetch.go
```diff
@@ -26,86 +26,174 @@ import (
...
+			return "", err
+		}
+		util.Tracef("EnsureModel: fetched %s (%s)\n", modelName, time.Since(fetchStart))
+		util.Tracef("EnsureModel total: %s\n", time.Since(totalStart))
+		return basePath, nil
+	}
+
+	// Core tag returned non-200 (likely 404)
+	if coreStatus == http.StatusNotFound && serviceSDKVersion != "" {
+		// Fallback to per-service tag
```
knottnt: Q: If the per-service-tag is able to represent a newer version of the api model file should it take precedence?

pkg/sdk/fetch.go
```diff
@@ -26,86 +26,174 @@ import (
...
 // JSON file for a given service and SDK version. If the file is already cached,
 // it returns immediately. Otherwise, it fetches the file from GitHub.
 //
+// When the core SDK tag URL returns HTTP 404 and serviceSDKVersion is non-empty,
+// EnsureModel retries using the per-service tag URL format. Per-service models
+// are cached under a separate path to prevent collisions with core tag models.
+//
 // The returned string is the base path to use with NewHelper — it mirrors the
 // SDK repo directory structure so that ModelAndDocsPath works unchanged.
 func EnsureModel(
```
knottnt, line 47: p: having the logic for checking both the core and per-service versions is making this function quite a bit more complex. Wondering if it would be possible to make the decision on whether or not to use the core vs per-service version outside of this function and base the parameters we pass into this function on that decision.

cmd/ack-generate/command/root.go
```diff
@@ -124,6 +125,9 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(
 		&optAWSSDKGoVersion, "aws-sdk-go-version", "", "Version of github.com/aws/aws-sdk-go-v2 used to generate apis and controllers files",
 	)
+	rootCmd.PersistentFlags().StringVar(
```
knottnt, line 128: Q: Should this flag be mutually exclusive with `aws-sdk-go-version`? What are the scenarios where we'd want both to be set?

knottnt (commented):
Thanks @sapphirew. Left some comments. Mostly questions about how we want to use the new flag.

pkg/sdk/fetch.go
sapphirew: Agreed. The per-service API models are always more up to date until the core sdk catches up in a few weeks. It should take precedence when set

pkg/sdk/fetch.go
sapphirew: I think we can always use per-service version when set. If core sdk version is set at the same time, return an error or warn

cmd/ack-generate/command/root.go
sapphirew: Makes sense. If core sdk version is set at the same time, we can return an error

pkg/sdk/fetch.go
sapphirew: Nope, it would return a 4xx, which would fall through to L102 below to the coreStatus check

pkg/metadata/generation_metadata_test.go
```diff
@@ -0,0 +1,124 @@
...
+			assert.Equal(t, original.AWSServiceSDKVersion, restored.AWSServiceSDKVersion)
+			assert.Equal(t, original.ACKGenerateInfo, restored.ACKGenerateInfo)
+			assert.Equal(t, original.GeneratorConfigInfo, restored.GeneratorConfigInfo)
+		})
+	}
+}
+
+// TestGenerationMetadata_BackwardCompatibility verifies that YAML without the
+// aws_service_sdk_version field deserializes without error and the field
+// defaults to empty string (Requirement 4.3).
```
knottnt, line 99: nit: `(Requirement 4.3).` doesn't refer to anything.
```suggestion
// defaults to empty string.
```

pkg/metadata/generation_metadata_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			original := GenerationMetadata{
+				APIVersion:           "v1alpha1",
+				APIDirectoryChecksum: "abc123",
+				LastModification: lastModificationInfo{
+					Reason: UpdateReasonAPIGeneration,
+				},
+				AWSSDKGoVersion:      "v1.41.5",
```
knottnt, line 54: It would be good to also test that `AWSSDKGoVersion` is emitted as expected.

pkg/sdk/fetch.go
```diff
@@ -26,86 +26,194 @@ import (
...
+// URL. It never falls back to the core SDK tag.
+func ensureModelPerService(
+	ctx context.Context,
+	cacheDir string,
+	modelName string,
+	serviceSDKVersion string,
+	totalStart time.Time,
+) (string, error) {
+	normalizedSvcVer := EnsureSemverPrefix(serviceSDKVersion)
+	basePath := filepath.Join(cacheDir, "models", "service", modelName, normalizedSvcVer)
```
knottnt, line 73: Doing some testing `modelName` doesn't appear to work as-is for all controllers. For example the `route53-controller` uses `route-53` as the model name, but the service package directory uses `route53`. Using `svcAlias` here instead may be a better general fit. Going forward we may also need to add a new generator.yaml `sdk_names:pkg_name` config similar to the `sdk_names:model_name` config that's causing route53's issue.

https://github.com/aws-controllers-k8s/route53-controller/blob/6b66359fac93b613ef69b68ac6e185137b9247f6/generator.yaml#L23

cmd/ack-generate/command/apis.go
```diff
@@ -64,7 +64,8 @@ func saveGeneratedMetadata(cmd *cobra.Command, args []string) error {
 		optGenVersion,
 		filepath.Join(optOutputPath, "apis"),
 		ackmetadata.UpdateReasonAPIGeneration,
-		optAWSSDKGoVersion,
```
knottnt, line 67: Testing this PR against the `s3-controller` I see that after calling `make build-controller SERVICE=s3 AWS_SERVICE_SDK_VERSION=v1.99.1` the ack-generate-metadata.yaml file looks like the below and includes both `aws_sdk_go_version` and `aws_service_sdk_version`. Should only one of these fields be emitted?

```
ack_generate_info:
  build_date: "2026-03-31T19:59:08Z"
  build_hash: a9e2ceaadfc00a742e2ea2b6d6c68348f03e52a5
  build_date: "2026-04-17T23:51:02Z"
  build_hash: 2f10e7b367cd01230a7916e4956c61f97be8cf13
  go_version: go1.26.1
  version: v0.58.0-3-ga9e2cea
api_directory_checksum: 379f9aa07359a58fabb591e795271b44a678b20c
  version: v0.58.0-5-g2f10e7b
api_directory_checksum: 29a8845f7c3bf83c6d20107a582b35d7a912d93a
api_version: v1alpha1
aws_sdk_go_version: v1.41.5
aws_service_sdk_version: v1.99.1
generator_config_info:
  file_checksum: 035386df2e486fac01a2dbedf247c36bbaeec7f3
  original_file_name: generator.yaml
```

knottnt (commented):
Thanks @sapphirew! Left a few more comments.

PR #692, opened 2026-04-17
Fix nested secrets and names conversion

visit1985:
Issue #, if available:

This PR is related to https://github.com/aws-controllers-k8s/community/issues/2813.

During my attempt to generate the ack-dms-controller, I identified the following problems:

- Nested secret fields with acronyms (e.g., `SASLPassword`, `SSLClientKeyPassword`) could not be found in `MemberFields` map because the lookup used raw AWS SDK names (`SaslPassword`, `SslClientKeyPassword`) while the map is keyed by normalized camelCase names.
- Secret field setters for nested struct members were not being automatically suppressed based on the `is_secret: true` configuration.

Description of changes:

- All secret fields (top-level and nested) marked with `is_secret: true` automatically suppress SDK→CRD output setters.
- Field config lookups are now consistently case-insensitive across all code paths.
- Acronym normalization (e.g., `Sasl` → `SASL`) no longer causes member field lookup failures.

### DIFF
--- a/pkg/generate/code/set_resource.go
+++ b/pkg/generate/code/set_resource.go
@@ -237,6 +237,9 @@ func SetResource(
 		if setCfg != nil && setCfg.IgnoreResourceSetter() {
 			continue
 		}
+        if inSpec && f.FieldConfig != nil && f.FieldConfig.IsSecret {
+            continue
+        }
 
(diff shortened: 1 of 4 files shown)


michaelhtm:
Thanks @visit1985 
/test all

visit1985:
Looks like a flaky test. I re-generated ec2-controller locally, and there is no change in the output.

/retest

PR #697, opened 2026-05-06
fix(compare): support scalar slice element types in compareSlice

sapphirew:
Issue #, if available: N/A

Description of changes:

Add long, integer, double, float, and boolean to the existing list/map case in compareSlice(). These scalar pointer slice types (e.g. []*int64) previously hit the default error case with 'unsupported element type in compareSlice: long', forcing service controllers to use compare.is_ignored workarounds.

Uses DeepEqual (same as list/map/structure cases) since there is no typed slice comparison helper in the runtime for these types.

Add s3files test fixture with AccessPoint resource containing a PosixUser.SecondaryGIDs field ([]*int64) to exercise the new code path.


knottnt (approved):
/lgtm

PR #699, opened 2026-05-08
feat: Generate cross-namespace reference validation and --enable-cross-namespace flag

sapphirew:
Issue #, if available:

Description of changes:

Generate cross-namespace reference **validation** into controller code and expose the `--enable-cross-namespace` flag, aligned with the runtime changes in aws-controllers-k8s/runtime#254.

### Resource references

Replace the legacy unconditional namespace-override block in `ResolveReferencesForField` with a call to the shared runtime helper `ackrt.ResolveCrossNamespaceReference`. This centralizes cross-namespace reference validation: when `--enable-cross-namespace` is disabled, references targeting a different namespace are rejected with a terminal error; when enabled (Phase 1), the runtime sets the `ACK.Advisory` deprecation notice. Generated code passes `&ko.Status.Conditions` to the helper; it does not manage the notice itself.

### Secret references

Cross-namespace validation for secrets now lives entirely in the runtime's `SecretValueFromReference` (runtime#254), so the generator emits the **unchanged** call:

```go
tmpSecret, err := rm.rr.SecretValueFromReference(ctx, r.ko.Spec.MasterUserPassword)
```

No per-call validation block and no resource-handle argument are generated. The runtime reaches the owning resource via a context-stash (set in `Reconcile()`), which keeps `SecretValueFromReference` signature-compatible and avoids breaking hand-written callers (e.g. RDS `db_cluster/custom_update.go`).

> An earlier revision of this PR threaded the owning resource into the generated call (`SecretValueFromReference(ctx, r, …)`) to match a proposed runtime signature change. Both were reverted once the runtime adopted the context-stash; the generated secret call is back to its original 2-arg shape.

### Helm chart

Expose the `--enable-cross-namespace` flag (default `true` for Phase 1 warning behavior), covering resource references, secret references, and field exports.

### Changes
- `resource_reference.go`: emit `ResolveCrossNamespaceReference` call with `EnableCrossNamespace` and conditions pointer
- `set_sdk.go`: secret fields emit the unchanged `SecretValueFromReference(ctx, source)` (validation handled in runtime)
- `*_test.go`: expected output strings updated to match the helper/call shapes
- `deployment.yaml.tpl`: add `--enable-cross-namespace` flag to container args
- `values.yaml.tpl`: add `enableCrossNamespace: true` with description
- `values.schema.json.tpl`: add `enableCrossNamespace` schema entry
- `go.mod`: temporary `replace` pinning the runtime to the runtime#254 branch commit until it merges and a release is tagged

### Dependency / merge order
This PR depends on runtime#254 (the `ACK.Advisory` notice + secret validation). The `replace` directive must be dropped and the `require` line bumped to the tagged runtime release once #254 merges.


kprahulraj:
Doc changes is added - https://github.com/aws-controllers-k8s/docs/pull/37

pkg/generate/code/resource_reference.go
```diff
@@ -228,9 +228,29 @@ func ResolveReferencesForField(field *model.Field, sourceVarName string, indentL
 			outPrefix += fmt.Sprintf("%s\treturn hasReferences, fmt.Errorf(\"provided resource reference is nil or empty: %s\")\n", innerIndent, refFieldPath)
 			outPrefix += fmt.Sprintf("%s}\n", innerIndent)
 
-			outPrefix += fmt.Sprintf("%snamespace := ko.ObjectMeta.GetNamespace()\n", innerIndent)
-			outPrefix += fmt.Sprintf("%sif arr.Namespace != nil && *arr.Namespace != \"\" {\n", innerIndent)
-			outPrefix += fmt.Sprintf("%s\tnamespace = *arr.Namespace\n", innerIndent)
+			outPrefix += fmt.Sprintf("%snamespace, isCrossNs, err := ackrt.ValidateCrossNamespaceReference(\n", innerIndent)
```
knottnt: Q: does the function comment need to be updated to reflect the new output?

templates/helm/templates/deployment.yaml.tpl
```diff
@@ -104,6 +104,8 @@ spec:
         - "$(FEATURE_GATES)"
 {{ "{{- end }}" }}
         - {{ "--enable-carm={{ .Values.enableCARM }}" }}
+        - {{ "--enable-cross-namespace={{ .Values.enableCrossNamespace }}" }}
+        - {{ "--enable-cross-namespace-references={{ .Values.enableCrossNamespaceReferences }}" }}
```
knottnt: I don't see this flag in the associated runtime changes.

templates/helm/values.yaml.tpl
```diff
@@ -180,6 +180,17 @@ leaderElection:
...
+# Enable cross-namespace behavior including resource references, secret references,
+# and field exports (default = true). When false, the controller rejects any operation
+# that crosses namespace boundaries.
+enableCrossNamespace: true
+
+# Enables cross-namespace resolution of AWSResourceReference entries.
+# When false (default), the controller rejects any reference whose
+# Namespace field differs from the namespace of the resource containing
+# the reference.
+enableCrossNamespaceReferences: false
```
knottnt: I don't see this flag in the associated runtime changes or in the values.schema.json.tpl.

pkg/generate/code/resource_reference.go
sapphirew: Good catch — updated the function comment to reflect the new generated output (ValidateCrossNamespaceReference call with three return values, error handling, and isCrossNs warning/condition logic).

templates/helm/templates/deployment.yaml.tpl
sapphirew: You're right — this was a leftover from the previous spec. The `--enable-cross-namespace-references` flag was superseded by `--enable-cross-namespace` in the runtime. Removed it.

templates/helm/values.yaml.tpl
sapphirew: Same issue — removed the stale `enableCrossNamespaceReferences` entry. Only `enableCrossNamespace: true` remains, which matches the runtime flag and is already in values.schema.json.tpl.

pkg/generate/code/resource_reference.go
```diff
@@ -156,31 +184,6 @@ func ReferenceFieldsValidation(
...
-//							if err := getReferencedResourceState_Bucket(ctx, apiReader, obj, *arr.Name, namespace); err != nil {
-//								return hasReferences, err
-//							}
-//							ko.Spec.Notification.LambdaFunctionConfigurations[f0idx].Filter.Key.FilterRules[f1idx].Value = (*string)(obj.Spec.Name)
-//						}
-//					}
-//				}
-//			}
-//		}
-//	}
```
michaelhtm, line 183: nit: why is this removed?

pkg/generate/code/resource_reference_test.go
```diff
@@ -384,9 +504,29 @@ func Test_ResolveReferencesForField_SingleReference_WithinMultipleSlices(t *test
...
+									"ownerNamespace", ko.ObjectMeta.GetNamespace(),
+									"targetNamespace", *arr.Namespace,
+									"referenceName", *arr.Name,
+								)
+								crossNsMsg := fmt.Sprintf("Cross-namespace resource reference detected: "+
+									"resource in namespace %q references %q in namespace %q. "+
+									"Cross-namespace behavior will be disabled by default in a future release. "+
+									"Set --enable-cross-namespace=true to preserve this behavior.",
+									ko.ObjectMeta.GetNamespace(), *arr.Name, *arr.Namespace)
+								setCrossNamespaceCondition(ko, crossNsMsg)
```
michaelhtm: can we make this block a helper function? maybe in runtime?

templates/pkg/resource/sdk.go.tpl
```diff
@@ -339,3 +341,29 @@ func (rm *resourceManager) terminalAWSError(err error) bool {
...
+		crossNsCond = &ackv1alpha1.Condition{
+			Type: ackv1alpha1.ConditionTypeCrossNamespaceOptInRequired,
+		}
+		ko.Status.Conditions = append(ko.Status.Conditions, crossNsCond)
+	}
+	now := metav1.Now()
+	crossNsCond.LastTransitionTime = &now
+	crossNsCond.Status = corev1.ConditionTrue
+	crossNsCond.Message = &message
+}
```
michaelhtm: nit: this could also be a runtime function, similar to https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/runtime/reconciler.go#L555

templates/pkg/resource/references.go.tpl
```diff
@@ -134,3 +141,6 @@ func (rm *resourceManager) resolveReferenceFor{{ $field.FieldPathWithUnderscore
 {{ end -}}
 {{ end -}}
 
+{{ if .CRD.HasReferenceFields -}}
+{{ end -}}
+
```
michaelhtm: what does this part do?

pkg/generate/code/resource_reference.go
sapphirew: Good catch — restored. The new sample shows the simpler generated output that uses `ackrt.HandleCrossNamespaceReference` from runtime#249.

pkg/generate/code/resource_reference_test.go
sapphirew: Done — moved into runtime as `ackrt.HandleCrossNamespaceReference` (see https://github.com/aws-controllers-k8s/runtime/pull/249). The block in generated code collapses from ~14 lines to a single helper call.

templates/pkg/resource/sdk.go.tpl
sapphirew: Good call — done. The lookup-or-create logic now lives in the runtime as `ackrt.SetCrossNamespaceOptInRequired`, mirroring the field export reconciler pattern you linked. The per-resource generated function is gone.

templates/pkg/resource/references.go.tpl
sapphirew: Nothing — that was leftover dead code from when `setCrossNamespaceCondition` was briefly defined here before being moved. Removed it.

pkg/generate/code/resource_reference.go
```diff
@@ -130,6 +136,22 @@ func ReferenceFieldsValidation(
...
+//		if err != nil {
+//			return hasReferences, err
+//		}
+//		if isCrossNs {
+//			ko.Status.Conditions = ackrt.HandleCrossNamespaceReference(
+//				ctx, ko.Status.Conditions,
+//				ackrt.CrossNamespaceRefKindResource,
+//				ko.ObjectMeta.GetNamespace(), *arr.Namespace, *arr.Name,
+//			)
+//		}
```
michaelhtm: i was hoping this entire section would be inside the helper function..and we won't need to return `isCrossNs`

sapphirew: sure, moved to the helper function

michaelhtm (approved):

pkg/generate/code/set_sdk.go
knottnt, line 1: Including this in the generated sdk.go files will get most of the SecretKeyReference usages, but I believe there are cases where Secrets are used in custom update functions. For example, RDS's MasterUserPassword is set in a [custom_update.go file](https://github.com/aws-controllers-k8s/rds-controller/blob/381268b247a2f4cd51784f4805fa687321274045/pkg/resource/db_cluster/custom_update.go#L579) and won't be updated automatically. Is there any way we can move this validation to happen in SecretValueFromReference entirely in runtime?

sapphirew: Good call, going to move secret validation into runtime's [SecretValueFromReference](https://github.com/aws-controllers-k8s/runtime/blob/14be98880374b9722ce4c3a077bcee2437664cc1/pkg/runtime/reconciler.go#L125) from the current change in sdk.go.

sapphirew: runtime PR: https://github.com/aws-controllers-k8s/runtime/pull/254

go.mod
```diff
@@ -103,3 +103,5 @@ require (
 	sigs.k8s.io/randfill v1.0.0 // indirect
 	sigs.k8s.io/structured-merge-diff/v6 v6.3.0 // indirect
 )
+
+replace github.com/aws-controllers-k8s/runtime => github.com/sapphirew/ack-runtime v0.0.0-20260608212517-9bb3e738fe97
```
knottnt: This will need to be removed before merging.

knottnt (approved):

PR #700, opened 2026-05-08
Complete implementation for NilEqualsZeroValue

### LINKED ISSUE #2873: [SNS] Permanent Topic drift due to missing DisplayName field
**Describe the bug**
Topics are permanently out of sync and reconciled due to drift in default field DisplayName

```
"diff":[{"Path":{"Parts":["Spec","DisplayName"]},"A":null,"B":""}]
 ```

**Steps to reproduce**
Create a SNS topic with no DisplayName field.

**Expected outcome**
CR is in sync, no drift detected.

**Environment**

* Kubernetes version
* Using EKS (yes/no), if so version? 
* AWS service targeted - SNS

gustavodiaz7722:
### Problem

AWS APIs often return empty strings for unset optional fields (e.g., SNS Topic `DisplayName`). The desired CR state has `nil` for unspecified fields, causing permanent drift on every reconciliation.

### Solution

Wire the existing `NilEqualsZeroValue` config in `CompareFieldConfig` into the `compareNil` code generation function. When configured with `compare.nil_equals_zero_value: true` in `generator.yaml`, the generated delta code calls `ackcompare.IsNilEqualsZero()` to suppress false drift.

This is one-way: `nil` desired + zero-value latest = no drift. But zero-value desired + `nil` latest = real drift reported.

### Changes

- `pkg/generate/code/compare.go` — Generate `ackcompare.IsNilEqualsZero` guard inside the nil-difference block when config is set
- `pkg/generate/code/compare_test.go` — Unit test validating generated output for SNS Topic with `nil_equals_zero_value`
- `pkg/testdata/models/apis/sns/0000-00-00/generator-nil-equals-zero-value.yaml` — Test fixture

### Testing
- Validated that SNS controller creates correct code and verified logic with additional delta unit test. 

### Dependencies

Requires runtime with `IsNilEqualsZero` in `pkg/compare/zero.go`.

Fixes: https://github.com/aws-controllers-k8s/community/issues/2873

### DIFF
--- a/pkg/generate/code/compare.go
+++ b/pkg/generate/code/compare.go
@@ -301,6 +301,8 @@ func compareNil(
 	out := ""
 	indent := strings.Repeat("\t", indentLevel)
 
+	nilEqualsZeroValue := compareConfig != nil && compareConfig.NilEqualsZeroValue
+
 	switch shape.Type {
 	case "boolean", "string", "character", "byte", "short", "integer", "long",
 		"float", "double", "timestamp", "structure", "jsonvalue":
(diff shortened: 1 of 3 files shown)


sapphirew (commented):
~~can we test this change in sns controller? I suppose you will be using this new feature for [DisplayName](https://github.com/aws-controllers-k8s/sns-controller/blob/main/generator.yaml#L82). So preferably add a new e2e test to cover this case and configure code generator to run it https://github.com/aws-controllers-k8s/test-infra/blob/main/prow/jobs/jobs_config.yaml#L101~~

nvm, looks like there is already a unit test

sapphirew (approved):
/lgtm

PR #702, opened 2026-05-12
fix: prevent aws:* tag leakage into CR spec during LateInitialize

michaelhtm:
Description of changes:
ReadOne may mutate its input via mirrorAWSTags, which injects aws:*
tags from the observed AWS state. In LateInitialize, the same object
passed to ReadOne was then used by lateInitializeFromReadOneOutput to
build the resource that gets patched back to the CR. This caused
AWS-managed tags (e.g. aws:cloudformation:*) to leak into the CR spec
after late initialization.

Pass a separate DeepCopy to ReadOne so that latestCopy remains clean
for lateInitializeFromReadOneOutput.


sapphirew (approved):
/lgtm

michaelhtm:
fixing here: https://github.com/aws-controllers-k8s/runtime/pull/246

PR #705, opened 2026-05-18
fix: ensure identifier lookup uses renames

michaelhtm:
Description of changes:
When looking for identifiers, code-gen expected generator.yaml fields to
contain the normal name (eg. Id instead of ID).
With this change, we ensure to check for both `Id` and `ID` when looking
for the identifier


knottnt (approved):
/lgtm

PR #707, opened 2026-05-22
Fix error generating custom fields for shared Statement shape in Wafv2

knottnt:
Issue #, if available: [2893](https://github.com/aws-controllers-k8s/community/issues/2893)

Description of changes:
  - Fix generation failure when multiple resources inject custom nested fields into the same underlying SDK shape (e.g. WAFv2's Statement shape shared between WebACL and RuleGroup)
  - Add conflict detection: error with a clear message when overlapping custom fields have different types
  - Log a warning when a compatible duplicate injection is skipped, including the shared shape name
  - Add WAFv2 integration test validating the shared-shape scenario end-to-end

Context:
  PR #691 (Support custom list and map fields, ac02b64) introduced addMemberShapRef with a strict duplicate check that errored unconditionally when a member already existed. This broke controllers like wafv2-controller where both RuleGroup and WebACL configure custom nested fields (e.g. Rules.Statement.AndStatement) that resolve to the same shared SDK Statement shape. The second resource's injection would fail with:

```
Error: resource "WebACL", custom nested field "Rules.Statement.AndStatement": member "AndStatement" already exists in shape of type "structure"
```

### DIFF
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -825,7 +825,10 @@ func addMemberShapRef(shapeRef, memberShapeRef *awssdkmodel.ShapeRef, fieldName 
 		return fmt.Errorf("unsupported shape type %q for adding member %q", shapeRef.Shape.Type, fieldName)
 	}
 	if _, exists := memberRefs[fieldName]; exists {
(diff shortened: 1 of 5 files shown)


michaelhtm (approved):
/lgtm

PR #708, opened 2026-05-26
fix: resolve Go field names correctly in late_initialize code generation

michaelhtm:
Description of changes:
The late_initialize template was using config field keys directly as Go
struct accessors. For fields from unpack_attributes_map (e.g. SQS
"SqsManagedSseEnabled"), the config key differs from the Go field name
("SQSManagedSSEEnabled"). This caused compilation errors in generated
controllers.

### DIFF
--- a/pkg/generate/code/late_initialize.go
+++ b/pkg/generate/code/late_initialize.go
@@ -22,6 +22,16 @@ import (
 	"github.com/aws-controllers-k8s/code-generator/pkg/model"
 )
 
+func fieldGoName(r *model.CRD, configName string) string {
+	if f, ok := r.SpecFields[configName]; ok {
+		return f.Names.Camel
+	}
+	if f, ok := r.Fields[configName]; ok {
+		return f.Names.Camel
+	}
+	return configName
+}
+
 // FindLateInitializedFieldNames outputs the code to create a sorted slice of fieldNames to
 // late initialize. This slice helps with short circuiting the AWSResourceManager.LateInitialize()
 // method if there are no fields to late initialize.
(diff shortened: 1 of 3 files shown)


michaelhtm:
generated https://github.com/aws-controllers-k8s/sqs-controller/pull/107 with these changes

pkg/generate/code/late_initialize.go
```diff
@@ -20,8 +20,23 @@ import (
 
 	ackgenconfig "github.com/aws-controllers-k8s/code-generator/pkg/config"
 	"github.com/aws-controllers-k8s/code-generator/pkg/model"
+	"github.com/aws-controllers-k8s/pkg/names"
 )
 
+// fieldGoName resolves a config field name to its Go-exported name using the
+// CRD's field model. For attribute-unpacked fields, the config key may differ
+// from the Go name (e.g. "SqsManagedSseEnabled" -> "SQSManagedSSEEnabled").
+func fieldGoName(r *model.CRD, configName string) string {
```
knottnt, line 25: Q: It seems like this should be a problem we've already solved for the `sdk.go` code. How are we determining the Go field name there and can we re-use it here? Ideally, we should have one way of doing this that is shared by all of our generation functions.

pkg/generate/code/late_initialize.go
knottnt, line 1: It would be good to add a new test case to cover this.

knottnt (commented):
Left a comment to see if we can use a shared utility for determining the Go field name that should be used. I'd guess we already have one that is being used elsewhere.

pkg/generate/code/late_initialize.go
michaelhtm: There isn't a standalone shared utility for this today.. sdk.go does it inline since it already has the field from iterating operation shapes. late_initialize is the only codepath starting from bare config keys.
Modified the function to lookup the existing field, instead of re-initializing names.New()

knottnt (approved):
/lgtm

PR #712, opened 2026-06-03
Fix zero-value defaultValue trait only in member ref using pointer SDK value

knottnt:
Issue #, if available:

Description of changes:
Fixes bug when generating code for SDK member fields that had `smith.api#defaultValue` trait set to Go zero-value only in member reference and not the underlying target shape. When generating code for such a field the ACK code-generator both failed to check the `defaultValue` trait set at the member ref level and incorrectly treated the field as a non-pointer type when the default used a non-zero value. 

Additionally, to avoid regressions in controllers that were previously ignoring member level defaultValue traits support for reading the `addedDefault` and `clientOptional` traits to support forcing fields with those traits to use pointer types.

Smithy references:
[GoPointableIndex.isMemberPointable](https://github.com/aws/smithy-go/blob/3f6ece24dd0646e1f43ce05d980bcb4880a6e7fe/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/knowledge/GoPointableIndex.java#L166)
[NullableIndex.CLIENT_ZERO_VALUE_V1_NO_INPUT](https://github.com/smithy-lang/smithy/blob/d62a9fca81f30d4e7773e58fb3c67d3ab682dfb6/smithy-model/src/main/java/software/amazon/smithy/model/knowledge/NullableIndex.java#L101)

- Add IsNonPointerInSDK() to encompass logic for when to use a pointer value for an SDK shape.
- Update set_resource and set_sdk to use IsNonPointerInSDK() instead of directly calling HasDefaultValue().
- Add test case for Bedrock AgentCore Browser resource

### DIFF
--- a/pkg/api/shape.go
+++ b/pkg/api/shape.go
@@ -1059,13 +1059,68 @@ func (s *Shape) IsEnum() bool {
 	return s.Type == "string" && len(s.Enum) > 0
 }
 
-// HasDefaultValue returns whether this shape has a default value
+// HasDefaultValue returns whether this shape has a default value.
+// A ShapeRef has a default value if either the ref itself carries an explicit
+// default (set from the member's "smithy.api#default" trait) or the underlying
+// Shape carries one (e.g. the injected PrimitiveBoolean shape). The sentinel
+// value "<nil>" means a previously-assigned default was explicitly cleared.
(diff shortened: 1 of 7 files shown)


michaelhtm (approved):
/lgtm

PR #713, opened 2026-06-18
feat: support Smithy intEnum shapes

naclonts:
Issue #, if available:

Description of changes: Adds support for Smithy `intEnum` shapes, which previously panicked the generator with `Unsupported shape type: intEnum`. An intEnum is surfaced as a plain integer (`*int64`), passing the value through unchanged at the SDK boundary — chosen over the human-friendly member name for forward-compatibility with new enum values.

Rather than mutating the model shape, downstream decision points (`IsNonPointerInSDK`, scalar set_sdk/set_resource, compare) recognize intEnum directly. Includes a mwaaserverless model fixture (`EngineVersion`) and tests.

  MWAA Serverless `EngineVersion` intEnum: https://github.com/aws/aws-sdk-go-v2/blob/main/codegen/sdk-codegen/aws-models/mwaa-serverless.json#L776

### DIFF
--- a/pkg/apiv2/converter.go
+++ b/pkg/apiv2/converter.go
@@ -150,12 +150,6 @@ func buildAPI(shapes map[string]Shape) (*awssdkmodel.API, string, error) {
 		case "enum":
 			newApi.Shapes[name].Type = "string"
 			addEnumRef(newApi.Shapes[name], shape)
-		// intEnum shapes are surfaced on the ACK side as named string enums
-		// (the human-friendly member name), with the underlying integer values
-		// captured so the SDK boundary can convert name <-> int.
-		case "intEnum":
-			newApi.Shapes[name].Type = "string"
-			addIntEnumRef(newApi.Shapes[name], shape)
 		// union is considered to be an interface. needs to be handled
 		// using custom code/generator.yaml renaming so it can point to
 		// the correct structure.
@@ -213,14 +207,27 @@ func createApiOperation(shape Shape, name, serviceAlias string) *awssdkmodel.Ope
 func createApiShape(shape Shape) (*awssdkmodel.Shape, error) {
 	isException := shape.isException()
 
+	// Smithy `intEnum` shapes are non-pointer value fields in aws-sdk-go-v2, so
+	// normalize them to `integer` and tag a DefaultValue. HasDefaultValue() then
+	// routes them through the existing value-type (non-pointer) SDK field path.
+	shapeType := shape.Type
+	isIntEnum := shapeType == "intEnum"
+	if isIntEnum {
+		shapeType = "integer"
+	}
+
 	apiShape := &awssdkmodel.Shape{
-		Type:       shape.Type,
+		Type:       shapeType,
 		Exception:  isException,
 		MemberRefs: make(map[string]*awssdkmodel.ShapeRef),
 		MemberRef:  awssdkmodel.ShapeRef{},
 		KeyRef:     awssdkmodel.ShapeRef{},
 		ValueRef:   awssdkmodel.ShapeRef{},
 		Required:   []string{},
+	}
+
+	if isIntEnum {
+		apiShape.DefaultValue = "0"
 	}
 
 	if isException {
@@ -318,23 +325,6 @@ func addEnumRef(apiShape *awssdkmodel.Shape, shape Shape) {
 		enumValue, ok := member.Traits["smithy.api#enumValue"]
 		if ok {
 			memberName = enumValue.(string)
-		}
-		apiShape.Enum = append(apiShape.Enum, memberName)
-	}
-	slices.Sort(apiShape.Enum)
-}
-
-// addIntEnumRef populates a (now string-typed) shape from a Smithy intEnum.
-// The member name becomes the enum's string value (what the user sets on the
-// ACK side), and the integer from the `smithy.api#enumValue` trait is recorded
-// in IntEnumValues so the SDK conversion code can map name <-> int.
-func addIntEnumRef(apiShape *awssdkmodel.Shape, shape Shape) {
-	apiShape.IntEnumValues = map[string]int64{}
-	for memberName, member := range shape.MemberRefs {
-		if enumValue, ok := member.Traits["smithy.api#enumValue"]; ok {
-			if f, ok := enumValue.(float64); ok {
-				apiShape.IntEnumValues[memberName] = int64(f)
-			}
 		}
 		apiShape.Enum = append(apiShape.Enum, memberName)
 	}
--- a/pkg/apiv2/converter.go
+++ b/pkg/apiv2/converter.go
@@ -206,14 +206,28 @@ func createApiOperation(shape Shape, name, serviceAlias string) *awssdkmodel.Ope
 // from the apiv2 Shape.
 func createApiShape(shape Shape) (*awssdkmodel.Shape, error) {
 	isException := shape.isException()
+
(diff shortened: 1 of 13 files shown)


pkg/generate/code/set_sdk_whitebox_test.go
knottnt, line 1: In addition to the white box tests it would also be good to setup some test cases in [set_resource_test.go](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/code/set_resource_test.go) and [set_sdk_test.go](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/code/set_resource_test.go) for the particular service controller use case. For setting up the test you can add the controller model and a sample generator.yaml file to the [pkg/test_data](https://github.com/aws-controllers-k8s/code-generator/tree/main/pkg/testdata) directory.

pkg/apiv2/converter.go
```diff
@@ -206,8 +206,18 @@ func createApiOperation(shape Shape, name, serviceAlias string) *awssdkmodel.Ope
 // from the apiv2 Shape.
 func createApiShape(shape Shape) (*awssdkmodel.Shape, error) {
 	isException := shape.isException()
+
+	// Smithy `intEnum` shapes are non-pointer value fields in aws-sdk-go-v2, so
+	// normalize them to `integer` and tag a DefaultValue. HasDefaultValue() then
+	// routes them through the existing value-type (non-pointer) SDK field path.
+	shapeType := shape.Type
+	isIntEnum := shapeType == "intEnum"
```
knottnt: Instead of manipulating the shape it might be cleaner to update the [IsNonPointerInSDK()](https://github.com/aws-controllers-k8s/code-generator/blob/2970ca9b3789515150e27ab80630175a8c77cba4/pkg/api/shape.go#L1096)

knottnt (commented):
@naclonts from the [Smithy docs](https://smithy.io/2.0/spec/simple-types.html#intenum) is look like intEnum provides a name for each numeric value. From a usability perspective should we be accepting that human friendly name instead of the integer?

pkg/generate/code/set_sdk_whitebox_test.go
naclonts: Sounds good, added some test cases there, with new mwaaserverless models.

naclonts:
> @naclonts from the [Smithy docs](https://smithy.io/2.0/spec/simple-types.html#intenum) is look like intEnum provides a name for each numeric value. From a usability perspective should we be accepting that human friendly name instead of the integer?

Yeah, that sounds like a better user experience. Just updated to take an approach similar to `Enum` fields - Let me know what you think.

pkg/generate/code/set_sdk_test.go
```diff
@@ -6987,3 +6987,84 @@ func TestSetSDK_LambdaMicrovms_MicrovmImage_Create_ValidateField(t *testing.T) {
...
+		}
+		if r.ko.Spec.EncryptionConfiguration.Type != nil {
+			f3.Type = svcsdktypes.EncryptionType(*r.ko.Spec.EncryptionConfiguration.Type)
+		}
+		res.EncryptionConfiguration = f3
+	}
+	if r.ko.Spec.EngineVersion != nil {
+		switch *r.ko.Spec.EngineVersion {
+		case "ONE":
+			res.EngineVersion = svcsdktypes.EngineVersion(1)
```
knottnt: Hmm, this a tough one. From what I can tell this generated code won't be forward compatible when the AWS service adds new enum values since we won't have a mapping for the string name of the new value. Just to double check with the previous approach of using an integer would the aws-sdk-go-v2 library allow us to send an integer value not included in `svcsdktypes.EngineVersion`?

naclonts: That's a really good catch, I think you're right. The original int-based approach should be forwards-compatible. 

That makes me lean toward going back to the int-based approach... Especially considering the [MWAA controller's enum is literally just "ONE"](https://github.com/aws/aws-sdk-go-v2/blob/main/codegen/sdk-codegen/aws-models/mwaa-serverless.json#L776), there's not much devex improvement in seeing "ONE" versus 1, IMO. Thoughts?

knottnt: Yeah I agree, the lack of forward compatibility is a big drawback and isn't worth any potential ease of use from using a human friendly string. My main concern would be that another API uses an enum name that is less generic and not as obviously tied to the integer value. Although at that point the service team should be using a string enum instead.

Taking a look at the [aws-cli ](https://docs.aws.amazon.com/cli/latest/reference/mwaa-serverless/create-workflow.html#:~:text=engine%2Dversion,-%28integer)I can see that it is using an integer representation as well for engine version.

naclonts: Yep, makes sense. Updated this PR back to the integer-based approach. I kept the mwaa-serverless fixture for test coverage.

pkg/apiv2/converter.go
```diff
@@ -216,6 +226,10 @@ func createApiShape(shape Shape) (*awssdkmodel.Shape, error) {
 		Required:   []string{},
 	}
 
+	if isIntEnum {
```
knottnt: Instead of mutating the shape we read from the model I think it would better to update the [isNonPointerInSDK()](https://github.com/aws-controllers-k8s/code-generator/blob/0a6b791fa10a6140d862be334a6891868ee588eb/pkg/api/shape.go#L1096) function to properly interpret intEnums

pkg/apiv2/converter.go
```diff
@@ -206,8 +206,18 @@ func createApiOperation(shape Shape, name, serviceAlias string) *awssdkmodel.Ope
...
 func createApiShape(shape Shape) (*awssdkmodel.Shape, error) {
 	isException := shape.isException()
+
+	// Smithy `intEnum` shapes are non-pointer value fields in aws-sdk-go-v2, so
+	// normalize them to `integer` and tag a DefaultValue. HasDefaultValue() then
+	// routes them through the existing value-type (non-pointer) SDK field path.
+	shapeType := shape.Type
+	isIntEnum := shapeType == "intEnum"
+	if isIntEnum {
+		shapeType = "integer"
```
knottnt: I'm guessing this is being done to force `intEnum` down the same code-paths in set_resource/set_sdk. IMO it would be cleaner to instead update the downstream decision points to handle intEnum. Mutating the underlying model seems like it could be fragile and difficult to refactor if we ever need change how intEnum is handled in the future.

knottnt: How are the other integer variants, short/long, being handled?

knottnt (commented):
Thanks @naclonts. The integer approach makes sense, but was wondering if we could avoid mutating the model shape.

knottnt (approved):
/lgtm

pkg/apiv2/converter.go
naclonts: Updated `intEnum` to be handled in the downstream decision points instead of re-assigning its type.  

> How are the other integer variants, short/long, being handled?

Looks like they're considered non-pointer only if they have a non-zero default. I added a a condition for IntEnum to explicitly be considered a non-pointer on the SDK side (in `isNonPointerInSDK`). I think this matches the general [Smithy](https://github.com/aws/smithy-go/blob/main/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/knowledge/GoPointableIndex.java#L166-L177) (and then [SDK](https://github.com/aws/aws-sdk-go-v2/blob/main/service/mwaaserverless/types/enums.go#L24)) approach.

PR #716, opened 2026-06-24
fix: Inject Smithy prelude Unit shape in v2 model converter

knottnt:
Issue #, if available:

Description of changes:
Models with a union (or structure) member targeting smithy.api#Unit — the Smithy prelude empty-structure type used for valueless union variants, e.g. lambda-microvms' PortSpecification.allPorts — panicked during model load with 'unable resolve reference, Unit'.

The v2->v1 converter injects bare prelude scalar shapes (String, Integer, ...) so members targeting them resolve, but Unit was missing from that set. Enum members also target smithy.api#Unit but are converted to plain string values without creating a shape ref (addEnumRef), so only union/structure members hit the panic.

Inject Unit as an empty structure, mirroring how aws-sdk-go-v2 itself models it. Reachable empty Unit fields are then handled by the existing empty_shapes generator.yaml mechanism, the same as any other empty structure.

Add pkg/apiv2 converter tests covering the injected shape and an end-to-end union->Unit reference-resolution repro.

### DIFF
--- a/pkg/apiv2/converter.go
+++ b/pkg/apiv2/converter.go
@@ -390,6 +390,7 @@ func injectCustomPrimitiveShapes() map[string]*awssdkmodel.Shape {
 		"PrimitiveLong",
 		"String",
 		"Timestamp",
+		"Unit",
 	}
 
 	for _, t := range types {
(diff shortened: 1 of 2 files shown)


sapphirew (approved):
/lgtm

PR #717, opened 2026-06-27
Remove outdated aws-sdk-go v1 anti-collision logic for Validate fields

knottnt:
Issue #, if available:

Description of changes:
The code-generator's renameCollidingFields pass (pkg/api/passes.go) unconditionally renamed any shape member named Validate to Validate_. This was a holdover from aws-sdk-go v1, where every generated input/request struct implemented a Validate() error method (via aws/request), so a member literally named Validate would collide with it. aws-sdk-go-v2 restructured generated types and no longer emits that method, so the rename now produces references to a non-existent SDK field — e.g. svcsdktypes.MicrovmImageHooks.Validate_ — which fails to compile for any API with a Validate member (such as Lambda MicroVMs' MicrovmImageHooks). This PR removes the obsolete Validate case from collides() so the member maps to the real Validate field, updates the TestCollidingFields expectation, and adds SetSDK/SetResource regression tests against the Lambda MicroVMs model.

- Remove "Validate" from collides() in passes.go
- Update passes_test.go to reflect new behavior
- Add set_resource.go and set_sdk.go tests to validate that Validate field does not erroneously have "_" appended.

### DIFF
--- a/pkg/api/passes.go
+++ b/pkg/api/passes.go
@@ -429,8 +429,7 @@ func renameCollidingField(name string, v *Shape, field *ShapeRef) {
 func collides(name string) bool {
 	switch name {
 	case "String",
-		"GoString",
-		"Validate":
+		"GoString":
 		return true
 	}
 	return false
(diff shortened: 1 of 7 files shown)


michaelhtm (approved):
/lgtm

PR #718, opened 2026-06-29
fix: Preserve nested struct references on readback

gustavodiaz7722:
## Description

Cross-resource references that live inside a **nested struct** field are dropped from the resource's `.spec` after the controller reads the resource back from AWS. References resolve correctly (`ACK.ReferencesResolved=True`) and the resolved value is applied to AWS, but the `*Ref` field disappears from the spec after readback.

### Root cause

References are preserved across reconciles by the generated `ClearResolvedReferences`, which is called on `latest` in the runtime's `patchResourceMetadataAndSpec`. It only clears a concrete value when the sibling `*Ref` field is present on the resource.

Setting a resource's fields from an API response (`SetResource`) rebuilds nested struct fields from scratch and reassigns the parent struct:

```go
f16 := &svcapitypes.LambdaConfigType{}   // fresh struct, no *Ref fields
// ...populate concrete fields from the response...
ko.Spec.LambdaConfig = f16               // parent replaced -> *Ref fields lost
```

Because the response carries only the concrete value (e.g. an ARN) and has no concept of the `*Ref` field, the `*Ref` values the user supplied are discarded. With no `*Ref` field on `latest`, `ClearResolvedReferences` can't clear the concrete value and the subsequent merge patch writes the concrete value while dropping the `*Ref` field.

Top-level references are unaffected, because the `*Ref` field is a sibling that is never touched when setting fields from the response. Nested references break because the whole parent struct is reassigned. This is the same class of issue seen with ECR `encryptionConfiguration.kmsKeyRef`, EKS `computeConfig.nodeRoleRef`, CloudFront `distributionConfig.webACLRef`, and OpenSearch `cognitoOptions.*Ref`.

### Change

Add a `PreserveReferenceFields` code generator that emits code to copy nested `*Ref` values back from the original object into the rebuilt resource after the response is mapped, creating parent structs as needed:

```go
if r.ko.Spec.LambdaConfig != nil && r.ko.Spec.LambdaConfig.PreSignUpRef != nil {
    if ko.Spec.LambdaConfig == nil {
        ko.Spec.LambdaConfig = &svcapitypes.LambdaConfigType{}
    }
    ko.Spec.LambdaConfig.PreSignUpRef = r.ko.Spec.LambdaConfig.PreSignUpRef
}
```

This makes nested references behave like top-level references: `latest` carries both the `*Ref` and the concrete value, so `ClearResolvedReferences` does its job and the patch preserves the reference.

The code is emitted for the create, update, read-one, and read-many output paths, guarded by `HasReferenceFields`. Top-level references are skipped (already preserved) and references nested within lists are skipped (their concrete values are restored by index during resolution).

### Testing

- `go build ./...`, `gofmt`, and `go vet` are clean.
- Added `Test_ReferenceFieldsPreservation_NestedReference`; the full `pkg/generate/...` suite passes.
- Regenerated a controller with nested struct references and confirmed correct output for single-level and deeply nested cases.

### DIFF
--- a/templates/pkg/resource/sdk_find_read_one.go.tpl
+++ b/templates/pkg/resource/sdk_find_read_one.go.tpl
@@ -48,9 +48,6 @@ func (rm *resourceManager) sdkFind(
 {{ $hookCode }}
 {{- end }}
 {{ GoCodeSetReadOneOutput .CRD "resp" "ko" 1 }}
-{{- if .CRD.HasReferenceFields }}
-	rm.preserveReferenceFields(r.ko, ko)
-{{- end }}
 	rm.setStatusDefaults(ko)
 {{- if $setOutputCustomMethodName := .CRD.SetOutputCustomMethodName .CRD.Ops.ReadOne }}
 	// custom set output from response
--- a/templates/pkg/resource/sdk_find_read_one.go.tpl
+++ b/templates/pkg/resource/sdk_find_read_one.go.tpl
@@ -48,6 +48,9 @@ func (rm *resourceManager) sdkFind(
 {{ $hookCode }}
 {{- end }}
 {{ GoCodeSetReadOneOutput .CRD "resp" "ko" 1 }}
+{{- if .CRD.HasReferenceFields }}
+{{ GoCodePreserveReferenceFields .CRD "r.ko" "ko" 1 }}
+{{- end }}
 	rm.setStatusDefaults(ko)
 {{- if $setOutputCustomMethodName := .CRD.SetOutputCustomMethodName .CRD.Ops.ReadOne }}
 	// custom set output from response
--- a/templates/pkg/resource/sdk_find_read_one.go.tpl
+++ b/templates/pkg/resource/sdk_find_read_one.go.tpl
@@ -49,7 +49,7 @@ func (rm *resourceManager) sdkFind(
 {{- end }}
 {{ GoCodeSetReadOneOutput .CRD "resp" "ko" 1 }}
 {{- if .CRD.HasReferenceFields }}
-{{ GoCodePreserveReferenceFields .CRD "r.ko" "ko" 1 }}
+	rm.preserveReferenceFields(r.ko, ko)
 {{- end }}
 	rm.setStatusDefaults(ko)
 {{- if $setOutputCustomMethodName := .CRD.SetOutputCustomMethodName .CRD.Ops.ReadOne }}
--- a/templates/pkg/resource/sdk_update.go.tpl
+++ b/templates/pkg/resource/sdk_update.go.tpl
@@ -45,7 +45,7 @@ func (rm *resourceManager) sdkUpdate(
 {{- end }}
 {{ GoCodeSetUpdateOutput .CRD "resp" "ko" 1 }}
 {{- if .CRD.HasReferenceFields }}
-{{ GoCodePreserveReferenceFields .CRD "desired.ko" "ko" 1 }}
+	rm.preserveReferenceFields(desired.ko, ko)
 {{- end }}
 	rm.setStatusDefaults(ko)
 {{- if $setOutputCustomMethodName := .CRD.SetOutputCustomMethodName .CRD.Ops.Update }}
--- a/templates/pkg/resource/sdk_update.go.tpl
+++ b/templates/pkg/resource/sdk_update.go.tpl
@@ -44,9 +44,6 @@ func (rm *resourceManager) sdkUpdate(
 {{ $hookCode }}
 {{- end }}
 {{ GoCodeSetUpdateOutput .CRD "resp" "ko" 1 }}
-{{- if .CRD.HasReferenceFields }}
-	rm.preserveReferenceFields(desired.ko, ko)
-{{- end }}
 	rm.setStatusDefaults(ko)
 {{- if $setOutputCustomMethodName := .CRD.SetOutputCustomMethodName .CRD.Ops.Update }}
 	// custom set output from response
--- a/templates/pkg/resource/sdk_update.go.tpl
+++ b/templates/pkg/resource/sdk_update.go.tpl
@@ -44,6 +44,9 @@ func (rm *resourceManager) sdkUpdate(
 {{ $hookCode }}
 {{- end }}
 {{ GoCodeSetUpdateOutput .CRD "resp" "ko" 1 }}
+{{- if .CRD.HasReferenceFields }}
+{{ GoCodePreserveReferenceFields .CRD "desired.ko" "ko" 1 }}
+{{- end }}
 	rm.setStatusDefaults(ko)
 {{- if $setOutputCustomMethodName := .CRD.SetOutputCustomMethodName .CRD.Ops.Update }}
 	// custom set output from response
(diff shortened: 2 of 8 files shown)


gustavodiaz7722:
## Manual Validation Results

Built this branch's `ack-generate`, regenerated a controller with a struct-nested reference, and verified end-to-end on a live EKS cluster (us-west-2) that the nested `*Ref` is now preserved across create, readback, and update.

### Test subject

`cognitoidentityprovider` UserPool `lambdaConfig.*Ref` (struct-nested references to `lambda.Function`) — the same shape used in this PR's `PreserveReferenceFields` docstring. These fields **failed** before the fix (the `*Ref` was dropped from `.spec` and replaced with the concrete ARN after the Describe readback).

### What the regeneration produced

Regenerating the controller with this branch added preservation blocks to `pkg/resource/user_pool/sdk.go` (~300 lines) for every `lambdaConfig.*Ref`, in **both** the readback path and the update path, e.g.:

```go
// readback (sdkFind)
if r.ko.Spec.LambdaConfig != nil && r.ko.Spec.LambdaConfig.PreSignUpRef != nil {
    if ko.Spec.LambdaConfig == nil {
        ko.Spec.LambdaConfig = &svcapitypes.LambdaConfigType{}
    }
    ko.Spec.LambdaConfig.PreSignUpRef = r.ko.Spec.LambdaConfig.PreSignUpRef
}

// update (sdkUpdate)
if desired.ko.Spec.LambdaConfig != nil && desired.ko.Spec.LambdaConfig.PreSignUpRef != nil {
    ...
    ko.Spec.LambdaConfig.PreSignUpRef = desired.ko.Spec.LambdaConfig.PreSignUpRef
}
```

### Results

Each field PASSes only if: `ReferencesResolved=True`, `ResourceSynced=True`, concrete field **absent** from `.spec`, and `*Ref` **present** with the right target.

| Scenario | preSignUpRef | postConfirmationRef | customMessageRef | preAuthenticationRef | postAuthenticationRef |
| --- | --- | --- | --- | --- | --- |
| Before fix (controller PR as-is) | ❌ | ❌ | ❌ | ❌ | ❌ |
| After fix — create + readback | ✅ | ✅ | ✅ | ✅ | ✅ |
| After fix — post-update | ✅ | ✅ | ✅ | ✅ | ✅ |

**Create / readback** (`check_ref.py`):
```
  ACK.ResourceSynced: True ✅
  ACK.ReferencesResolved: True ✅
  ✅ lambdaConfig.preSignUpRef: concrete=None, ref_name=ack-test-cognito-fn
  ✅ lambdaConfig.postConfirmationRef: concrete=None, ref_name=ack-test-cognito-fn
  ✅ lambdaConfig.customMessageRef: concrete=None, ref_name=ack-test-cognito-fn
  ✅ lambdaConfig.preAuthenticationRef: concrete=None, ref_name=ack-test-cognito-fn
  ✅ lambdaConfig.postAuthenticationRef: concrete=None, ref_name=ack-test-cognito-fn
RESULT: PASS ✅
```

**Update path** — changed a non-reference mutable field (`policies.passwordPolicy.minimumLength` 8 → 12) to force an `UpdateUserPool`, keeping the 5 refs. The update was applied (`describe-user-pool` shows `MinimumLength: 12`) and all refs survived:
```
  ACK.ResourceSynced: True ✅
  ACK.ReferencesResolved: True ✅
  ✅ lambdaConfig.preSignUpRef: concrete=None, ref_name=ack-test-cognito-fn
  ✅ lambdaConfig.postConfirmationRef: concrete=None, ref_name=ack-test-cognito-fn
  ✅ lambdaConfig.customMessageRef: concrete=None, ref_name=ack-test-cognito-fn
  ✅ lambdaConfig.preAuthenticationRef: concrete=None, ref_name=ack-test-cognito-fn
  ✅ lambdaConfig.postAuthenticationRef: concrete=None, ref_name=ack-test-cognito-fn
RESULT: PASS ✅
```

**AWS-side confirmation** (`aws cognito-idp describe-user-pool`) — the resolved values were applied and stayed applied after the update:
```json
{
  "MinLen": 12,
  "Lambda": {
    "PreSignUp": "arn:aws:lambda:us-west-2:<acct>:function:ack-test-cognito-fn",
    "CustomMessage": "arn:aws:lambda:us-west-2:<acct>:function:ack-test-cognito-fn",
    "PostConfirmation": "arn:aws:lambda:us-west-2:<acct>:function:ack-test-cognito-fn",
    "PreAuthentication": "arn:aws:lambda:us-west-2:<acct>:function:ack-test-cognito-fn",
    "PostAuthentication": "arn:aws:lambda:us-west-2:<acct>:function:ack-test-cognito-fn"
  }
}
```

No perpetual diff after convergence (`"desired resource state has changed"` count = 0 over a 60s window once synced).

### Scope note

I also regenerated `elbv2` with this branch as a control: its generated code was **unchanged**, because its failing references (`Listener.defaultActions[].targetGroupRef`, `Rule.actions[].targetGroupRef`) are nested inside **lists**, which this fix intentionally skips (per the `PreserveReferenceFields` comments). So this PR resolves the **struct-nested** overwrite (cognito `lambdaConfig`, and by the same mechanism opensearch `cognitoOptions`, cloudfront `distributionConfig`, ECR `encryptionConfiguration`, etc.). List-nested references would need a separate follow-up.

### Reproduction
- Regenerated controller: `SERVICE=cognitoidentityprovider AWS_SDK_GO_VERSION=v1.32.6 make build-controller`
- Built/pushed image with `ko`, deployed via Helm, created a `lambda.Function` + `UserPool` using only `*Ref` fields.
- Verified with the standard `check_ref.py` (concrete absent + `*Ref` present + both conditions `True`).

templates/pkg/resource/sdk_update.go.tpl
```diff
@@ -44,6 +44,9 @@ func (rm *resourceManager) sdkUpdate(
 {{ $hookCode }}
 {{- end }}
 {{ GoCodeSetUpdateOutput .CRD "resp" "ko" 1 }}
+{{- if .CRD.HasReferenceFields }}
```
knottnt: There's a good number of resources that use a fully custom update function which won't be impacted by updating this template. Would it be possible to apply this in the ACK runtime?

templates/pkg/resource/sdk_find_read_one.go.tpl
```diff
@@ -48,6 +48,9 @@ func (rm *resourceManager) sdkFind(
 {{ $hookCode }}
 {{- end }}
 {{ GoCodeSetReadOneOutput .CRD "resp" "ko" 1 }}
+{{- if .CRD.HasReferenceFields }}
```
knottnt: Note, some fields may be set in either in the `setOutputCustomMethodName` function or in a `sdk_read_one_post_set_output` hook.

templates/pkg/resource/sdk_update.go.tpl
gustavodiaz7722: Good point, we should create a universal fix that will apply regardless if we have custom update function. I've refactored current implementation to use optional interface approach in runtime. 

https://github.com/aws-controllers-k8s/runtime/pull/257

PR #721, opened 2026-07-12
Add support for renamed nested fields

visit1985:
Issue #, if available:

https://github.com/aws-controllers-k8s/community/issues/2173

Description of changes:

In order to support references in WAFv2, it is necessary to rename (add some prefix to) nested `IPSetReferenceStatement.ARN` and `RuleGroupReferenceStatement.ARN` fields. This is an attempt to support this in the code-generator.


michaelhtm:
Hey @visit1985, Thanks for the contribution!
I'm not sure i understand tho how this change addresses the linked issue. 
Can you maybe open a wafv2 PR showing what this would look like?

visit1985:
> Hey @visit1985, Thanks for the contribution! I'm not sure i understand tho how this change addresses the linked issue. Can you maybe open a wafv2 PR showing what this would look like?

@michaelhtm, https://github.com/aws-controllers-k8s/community/issues/2173 requires to enable references on WebACLs and RuleGroups `RuleGroupReferenceStatement.ARN` and `IPSetReferenceStatement.ARN`.

Reference field names are generated by removing the identifier suffix (`ARN`) and adding `Ref(s)` as suffix.

https://github.com/aws-controllers-k8s/code-generator/blob/db232581a560896c2dc461a244f96d5bf3191ec6/pkg/model/field.go#L318-L323

So, we either rename the `ARN` field before creating the reference, or we allow `ref(s)` without a prefix as field name.

This PR fixes additional name conversion issues in nested field code paths that I didn't catch in https://github.com/aws-controllers-k8s/code-generator/pull/692.
e.g. `Rules.Statement.IPSetReferenceStatement.ARN` to `IPSetARN`

I'm aware that the rename of the `ARN` field would be a breaking change in the WAFv2 CRDs. If you prefer to allow `ref(s)` without a prefix for that specific case, I can open another PR for that. Nevertheless, this PR may be beneficial for other undiscovered edge cases.

Let me know what you think.

knottnt:
> > Hey @visit1985, Thanks for the contribution! I'm not sure i understand tho how this change addresses the linked issue. Can you maybe open a wafv2 PR showing what this would look like?
> 
> @michaelhtm, [aws-controllers-k8s/community#2173](https://github.com/aws-controllers-k8s/community/issues/2173) requires to enable references on WebACLs and RuleGroups `RuleGroupReferenceStatement.ARN` and `IPSetReferenceStatement.ARN`.
> 
> Reference field names are generated by removing the identifier suffix (`ARN`) and adding `Ref(s)` as suffix.
> 
> https://github.com/aws-controllers-k8s/code-generator/blob/db232581a560896c2dc461a244f96d5bf3191ec6/pkg/model/field.go#L318-L323
> 
> So, we either rename the `ARN` field before creating the reference, or we allow `ref(s)` without a prefix as field name.
> 
> This PR fixes additional name conversion issues in nested field code paths that I didn't catch in #692. e.g. `Rules.Statement.IPSetReferenceStatement.ARN` to `IPSetARN`
> 
> I'm aware that the rename of the `ARN` field would be a breaking change in the WAFv2 CRDs. If you prefer to allow `ref(s)` without a prefix for that specific case, I can open another PR for that. Nevertheless, this PR may be beneficial for other undiscovered edge cases.
> 
> Let me know what you think.

@visit1985 I don't think we'd want to rename the existing fields since, as you've noted, that would be a breaking change to the CRDs. Having a bare `ref` field with no prefix could be confusing. Maybe instead we could update the [ReferencesConfig](https://github.com/aws-controllers-k8s/code-generator/blob/db232581a560896c2dc461a244f96d5bf3191ec6/pkg/config/field.go#L355) to allow for a custom prefix to supplied. For WAFv2 we could then provide the `IPSet` and `RuleGroupReferenceStatement`as custom prefixes.

For this PR, I do agree that the ability to rename nested fields could be useful for future ACK resources.

cc: @michaelhtm

visit1985:
> @visit1985 I don't think we'd want to rename the existing fields since, as you've noted, that would be a breaking change to the CRDs. Having a bare `ref` field with no prefix could be confusing. Maybe instead we could update the [ReferencesConfig](https://github.com/aws-controllers-k8s/code-generator/blob/db232581a560896c2dc461a244f96d5bf3191ec6/pkg/config/field.go#L355) to allow for a custom prefix to supplied. For WAFv2 we could then provide the `IPSet` and `RuleGroupReferenceStatement`as custom prefixes.

@knottnt, ok, I reverted the WAFv2 test, which also fixes the failing unit-test. I will create a separate PR for the suggested custom prefix.

> For this PR, I do agree that the ability to rename nested fields could be useful for future ACK resources.

I fixed the s3-controller-test, but I need help to understand whats wrong with that dynamodb-controller-test. I re-generated the dynamodb controller from this branch before and after my latest fix. It doesn't cause any naming differences.

PR #723, opened 2026-07-16
add support for secret object, and allow adding CRD validation field

mkl262:
Issue #, if available:
required for aws-controllers-k8s/acm-controller#104

Description of changes:
* Allow generated APIs to reference Secrets whose data keys are controlled by the controller.
* support creation of x-kubernetes-validations fields in crd definitions

### DIFF
--- a/pkg/config/field.go
+++ b/pkg/config/field.go
@@ -422,6 +422,10 @@ type FieldConfig struct {
 	// IsSecret instructs the code generator that this field should be a
 	// SecretKeyReference.
 	IsSecret bool `json:"is_secret"`
+	// IsSecretReference instructs the code generator that this field should be
+	// a name/namespace-only SecretReference. Use this when the controller owns
+	// the Secret data key convention instead of accepting a user-selected key.
+	IsSecretReference bool `json:"is_secret_reference"`
 	// IsImmutable indicates that the field is enforced as immutable at the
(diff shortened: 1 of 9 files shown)


pkg/config/field.go
```diff
@@ -422,6 +422,10 @@ type FieldConfig struct {
 	// IsSecret instructs the code generator that this field should be a
 	// SecretKeyReference.
 	IsSecret bool `json:"is_secret"`
+	// IsSecretReference instructs the code generator that this field should be
+	// a name/namespace-only SecretReference. Use this when the controller owns
+	// the Secret data key convention instead of accepting a user-selected key.
+	IsSecretReference bool `json:"is_secret_reference"`
```
michaelhtm: i'm not sure if the difference between Secret and SecretReference is clear just from the names.

michaelhtm: maybe it can be something like `isTLSSecret`?

mkl262: I thought of it as a generlized catch all for secrets that have a predefined structure, such as docker config, TLS, or ssh type secrets.

mkl262: @michaelhtm Do you want me to rename it?

mkl262:
@michaelhtm I made the change, could you please review it?

PR #725, opened 2026-07-31
Add support for applying is_iam_policy and is_document custom comparisons to nested fields

knottnt:
Issue #, if available:

Description of changes:
Fixes custom comparison configs `is_iam_policy` and `is_document` being silently ignored when applied to a nested field.

- Update CompareStruct to check the field config for both is_iam_policy and is_document
- Add unit test to validate generated comparison code when is_iam_policy and is_document are set for a nested field

### DIFF
--- a/pkg/generate/code/compare.go
+++ b/pkg/generate/code/compare.go
@@ -881,6 +881,16 @@ func CompareStruct(
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


sapphirew (approved):
/lgtm

PR #726, opened 2026-08-03
Pin ACK runtime to v0.62.0

sapphirew:
Pin ACK runtime to `v0.62.0`.

Runtime `v0.62.0` includes the `ignore-field-drift` annotation feature ([runtime#256](https://github.com/aws-controllers-k8s/runtime/pull/256)), which lets users tell ACK to stop reconciling drift on specific spec fields. The feature is entirely in the generic runtime reconciler and is gated behind the `IgnoreFieldDrift` feature gate (Alpha, disabled by default), so no code-generator template changes are required.

Merging this and tagging code-generator `v0.62.0` triggers the Prow job that regenerates all controllers against the new runtime.

The `version/minor` label is applied so downstream controller releases get a minor bump, allowing them to advertise the new feature.


knottnt (approved):
/lgtm

PR #729, opened 2026-08-12
Remove overwrite of Notice from code-generator

knottnt:
Issue #, if available:

Description of changes:
In some cases contributors may add their own entries the service controller's NOTICE file. This PR updates the code-generator's build-controller.sh script to no longer overwrite those additions.


scripts/build-controller.sh
```diff
@@ -271,4 +271,3 @@ cp "$ROOT_DIR"/CODE_OF_CONDUCT.md "$SERVICE_CONTROLLER_SOURCE_PATH"/CODE_OF_COND
 cp "$ROOT_DIR"/CONTRIBUTING.md "$SERVICE_CONTROLLER_SOURCE_PATH"/CONTRIBUTING.md
 cp "$ROOT_DIR"/GOVERNANCE.md "$SERVICE_CONTROLLER_SOURCE_PATH"/GOVERNANCE.md
 cp "$ROOT_DIR"/LICENSE "$SERVICE_CONTROLLER_SOURCE_PATH"/LICENSE
-cp "$ROOT_DIR"/NOTICE "$SERVICE_CONTROLLER_SOURCE_PATH"/NOTICE
```
jlbutler, line 274: if we wanted to be nitty, could skip copy only if the file already exists... but i think this is fine.

jlbutler (approved):

scripts/build-controller.sh
knottnt: We do already populate the Notice with the controller-bootstrap tool. So, it should already exist once the repo is setup.

https://github.com/aws-controllers-k8s/controller-bootstrap/blob/main/templates/NOTICE.tpl

PR #730, opened 2026-08-13
Bump pkg to v0.0.26

knottnt:
Issue #, if available:

Description of changes:
Upgrades pkg to v0.0.26 to apply https://github.com/aws-controllers-k8s/pkg/pull/50

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/aws-controllers-k8s/code-generator
 go 1.25.0
 
 require (
-	github.com/aws-controllers-k8s/pkg v0.0.25
+	github.com/aws-controllers-k8s/pkg v0.0.26
 	github.com/aws-controllers-k8s/runtime v0.62.0
 	github.com/aws/aws-sdk-go v1.49.0
 	github.com/aws/aws-sdk-go-v2 v1.32.7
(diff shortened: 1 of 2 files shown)


rdoshi1:
The required `ec2-controller-test` is failing during bootstrap because the shared test account has reached its VPC quota. Could someone clean up the stale VPCs or address the quota, then comment `/test ec2-controller-test` on this PR?

rdoshi1:
Latest retry progressed past VPC creation but failed during EC2 bootstrap with `TransitGatewayLimitExceeded`. Cleanup then raised an `AttributeError` because the Transit Gateway was never created.

This PR only updates the pkg dependency in `go.mod` and `go.sum`, and the remaining checks pass, so this appears unrelated to the PR. Could a test-infra owner clean up the shared account’s Transit Gateways or restore its quota, then rerun `/test ec2-controller-test`?

Thank you!

rdoshi1:
The latest `ec2-controller-test` progressed past the previous VPC and Transit Gateway quota failures. This time, 85 tests passed, while 7 failed and 3 errored because `status.vpcID` or `status.transitGatewayID` was not populated before the tests read it.

The controller logs also show shared-account cleanup issues, including an authorization failure while releasing an Elastic IP and subnet dependency violations. Since this PR only updates the pkg dependency and the controller successfully populated vpcID for other resources in the same run, this still appears likely to be shared test-account state or timing rather than a regression caused by this PR.

Could a test-infra owner clean up the shared EC2 resources and run `ec2-controller-test` against both this PR and current main? That comparison should confirm whether the failure is environmental or specific to this change. Thank you!

rdoshi1:
The latest retry passed all other checks, but EC2 and EKS failed on what appear to be unrelated shared-environment/timing issues:
- EC2: missing `status.vpcID/status.transitGatewayID` values and cleanup failures.
- EKS: the test `checked authenticationMode` before the asynchronous AWS update completed.

Could a test-infra owner run these tests against both current main and this PR to confirm they are unrelated, then rerun:
```
  /test ec2-controller-test
  /test eks-controller-test
```

PR #731, opened 2026-08-13
fix(helm): expose IgnoreFieldDrift feature gate

Guerlielton:
## Summary

Add support for configuring the `IgnoreFieldDrift` feature gate in the ACK controller Helm values.

## Changes

- Added the `IgnoreFieldDrift` feature gate to the Helm values.
- Default value is set to `false` to preserve the current behavior.

## Impact

No functional changes are introduced. The feature remains disabled by default and can be enabled when needed.


Guerlielton:
Hi @knottnt, I don't think it's supposed to. I changed something from the DynamoDB and Lambda side because the test failed.

sapphirew:
known flaky tests for acm controller due to permission issue. Going to merge it for now. 
/approve

PR #732, opened 2026-08-13
feat: add field-level custom_sync generator config

gustavodiaz7722:
## Description

Some fields are managed by an AWS API separate from the resource's `Update` operation — tags via `CreateOrUpdateTags`, a logging configuration via its own `Put` call. Controllers reconcile these today by hand-writing an `sdk_update_pre_build_request` hook that invokes a sync function and then short-circuits out of `sdkUpdate` when nothing else differs.

Every controller doing this duplicates the same block. Worse, adding a second such field means widening the existing `DifferentExcept` call by hand, and forgetting to makes the resource silently stop applying legitimate updates, with no error surfaced anywhere.

This PR adds `custom_sync` as a field-level config that generates the boilerplate.

```yaml
resources:
  AutoScalingGroup:
    fields:
      Tags:
        custom_sync: {}
```

The sync function itself stays hand-written. It's a method on `resourceManager`, matching every other seam where generated code calls user-supplied code (`custom_implementation`, `set_output_custom_method_name`, and so on):

```go
func (rm *resourceManager) syncTags(
    ctx context.Context, desired *resource, latest *resource,
) error
```

The method name is always `sync<FieldName>` and is not configurable, so it's identical across every controller and can be found from the field name alone. `CustomSyncConfig` is an empty struct rather than a bool, so options can be added later without breaking the `generator.yaml` files that adopt `custom_sync: {}` now — promoting a bool to a struct would break under the strict unmarshalling `config.New` performs.

## Generated code

Into `sdkUpdate`:

```go
	updatedDesired := desired.DeepCopy()
	updatedDesired.SetStatus(latest)
	if delta.DifferentAt("Spec.Tags") {
		err = rm.syncTags(ctx, desired, latest)
		if err != nil {
			return nil, err
		}
	}
	if !delta.DifferentExcept("Spec.Tags") {
		return rm.concreteResource(updatedDesired), nil
	}
```

Into `sdkCreate`:

```go
	if ko.Spec.Tags != nil {
		msg := "Secondary sync required; resource will be requeued"
		ackcondition.SetSynced(&resource{ko}, corev1.ConditionFalse, &msg, nil)
	}
```

A `custom_sync` field is only ever applied in the update path, so immediately after create it exists in Spec but has not been pushed to AWS. Marking the resource unsynced makes the runtime requeue after `requeue.DefaultRequeueAfterDuration` rather than waiting for the full resync period, and the condition message tells the user a further sync is coming without their involvement.

The message is a generic constant rather than a list of the pending field paths. The nil checks guarding it are evaluated at runtime, but a generated message can only be assembled from every configured field, so naming them would over-report whenever a user populates only a subset.

With multiple `custom_sync` fields, the create-path guard collapses into one block, since `SetSynced` overwrites rather than accumulates:

```go
	if ko.Spec.LogDeliveryConfigurations != nil || ko.Spec.Tags != nil {
		msg := "Secondary sync required; resource will be requeued"
		ackcondition.SetSynced(&resource{ko}, corev1.ConditionFalse, &msg, nil)
	}
```

On the update side, all paths land in a single `DifferentExcept` call:

```go
	if !delta.DifferentExcept("Spec.LogDeliveryConfigurations", "Spec.Tags") {
```

That auto-widening is the main argument for generating this rather than documenting a snippet.

## Placement

Position in `sdk_update.go.tpl` is load-bearing in both directions, and is recorded in a comment on `CustomSyncUpdate`:

- **After** the `updateable.when` guard, so a resource in a state where mutations aren't allowed is requeued before any out-of-band API call is made, and so the short-circuit can't return success past the guard.
- **Before** the Update operation's `custom_implementation`, which returns from `sdkUpdate` directly and would otherwise skip the sync entirely.

## Validation

Rejected at generation time, since each otherwise produces a controller that compiles but misbehaves at runtime:

| Config | Why it's rejected |
| --- | --- |
| `custom_sync` on a nested field | Emitted code builds a `Spec.<Field>` delta path and a nil check off `ko.Spec` |
| `custom_sync` with `is_read_only` | Field lands in Status, where there's no desired value to sync toward |
| `custom_sync` with `compare.is_ignored` | Field never enters the delta, so the sync never runs and the short-circuit fires every reconcile |
| `custom_sync` with no Update operation | Generated code lives in `sdkUpdate`, so the sync would never be called |

## Testing

`make test` passes — 15 packages, no failures.

New unit tests cover the emitters (single field, two fields, and the no-config inert case) and the model accessors plus all validation errors. Fixtures are added under `pkg/testdata/models/apis/elasticache/0000-00-00/`.

Verified end to end by regenerating `autoscaling-controller`, which has exactly this pattern hand-written today:

- With no `custom_sync` configured, generated `sdk.go` is **byte-identical** to before, confirming the emitters are inert for resources that don't use the feature.
- With `custom_sync` on `Tags` and both hand-written hook templates deleted, the generated code matches what those hooks produced. The controller builds and its tests pass.
- Placement relative to `updateable.when` was confirmed by temporarily adding a guard to `AutoScalingGroup` and inspecting the generated output.

## Notes for reviewers

- Ordering across multiple `custom_sync` fields is alphabetical, which keeps generated output stable. Config-driven ordering is deferred until a resource needs it; note that YAML document order isn't recoverable here, since `ResourceConfig.Fields` is a map and the loader routes through `sigs.k8s.io/yaml`. An explicit `order:` key would be the way to add it.
- `custom_sync` is restricted to top-level Spec fields. Beyond the codegen reasons, `DifferentExcept` matches with `Path.Contains`, so allowing nested paths would let one diff match two except-paths and skew the count.
- No test asserts template placement, since the emitters are unit-tested in isolation. Happy to add a rendered-output assertion if you'd like that ordering pinned.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


michaelhtm (approved):

a-hilaly (approved):
👍

michaelhtm (approved):
/lgtm

PR #738, opened 2026-08-27
feat: generate EnsureReferences to preserve nested cross-resource references

### LINKED ISSUE #2431: The ACK Lambda Controller modifies the object spec
**Describe the bug**
I encountered an issue while creating a Function using `vpcConfig.securityGroupRefs`:

```yaml
status:
  ackResourceMetadata:
    arn: arn:aws:lambda:us-east-1:1234:function:pr13344879357-ci-va6-cw2splunk
    ownerAccountID: "1234"
    region: us-east-1
  conditions:
  - lastTransitionTime: "2025-04-17T11:22:25Z"
    message: Reference resolution failed
    reason: 'both resource reference wrapper and ID cannot be used together: VPCConfig.SecurityGroupIDs,VPCConfig.SecurityGroupRefs'
    status: Unknown
    type: ACK.ReferencesResolved
  lastModified: 2025-04-17T09:41:44.372+0000
  lastUpdateStatus: Successful
  revisionID: 93e97a3b-c3b0-451c-af76-b2928cb415bd
  state: Active
```

My spec contains these lines:
```yaml
    securityGroupRefs:
      - from:
          name: {{ include "cluster-build.clusterFullName" $ }}-cw2splunklambda
```

**Steps to reproduce**

I've also created a simple manifest with this in its spec:

```yaml
  vpcConfig:
    securityGroupRefs:
      - from:
          name: pr13344879357-ci-va6-cw2splunklambda
```

And the controller took this snippet of code and transformed it into:

```yaml
  vpcConfig:
    securityGroupIDs:
    - sg-042a062cc8a002db3
```

```sh
kubectl get securitygroup -n ci-clusters  pr13344879357-ci-va6-cw2splunklambda 
NAME                                   ID
pr13344879357-ci-va6-cw2splunklambda   sg-042a062cc8a002db3
```

**Expected outcome**

The ACK Lambda Controller accepts both securityGroupRefs and securityGroupIDs / just one of them, but it does not transform securityGroupRefs into securityGroupIDs.

I assume that AWS Lambda template doesn't support reference by name 
**Environment**

* Kubernetes version 1.30
* Using EKS (yes/no), if so version?  1.30
* AWS service targeted (S3, RDS, etc.) lambda-controller:1.6.1

gustavodiaz7722:
## Summary

A cross-resource reference (`*Ref`) is generated as a **sibling** of the concrete field it resolves into — `spec.vpcConfig.subnetRefs` next to `spec.vpcConfig.subnetIDs`. A resource manager builds its return value from an AWS API response, which has no concept of a reference, so rebuilding the containing struct drops every `*Ref` inside it.

That disables `ClearResolvedReferences`, which suppresses a resolved value only while it can still see the sibling:

```go
if ko.Spec.VPCConfig != nil {
	if len(ko.Spec.VPCConfig.SubnetRefs) > 0 {   // false once the *Ref is gone
		ko.Spec.VPCConfig.SubnetIDs = nil
	}
}
```

So the spec patch **deletes the declared `*Ref` and stores the resolved value in its place** — what aws-controllers-k8s/community#2431 reports: a declared `securityGroupRefs` replaced by `securityGroupIDs`.

Reconciliation continues until the manifest is applied again, from Helm, Argo, Flux or `kubectl apply`. That apply restores the `*Ref` beside the now-stored value, and `validateReferenceFields` rejects the pair:

```
message: Reference resolution failed
reason: 'both resource reference wrapper and ID cannot be used together:
         VPCConfig.SubnetIDs,VPCConfig.SubnetRefs'
```

This PR generates an `EnsureReferences` method that restores the missing reference from the declared resource.

Together with aws-controllers-k8s/runtime#267 this fixes aws-controllers-k8s/community#2431, and addresses the struct-nested half of aws-controllers-k8s/community#2361.

## Requires runtime v0.64.0

aws-controllers-k8s/runtime#267 defines the optional `ReferenceEnsurer` interface and calls the method on what a resource manager returns from `Create` and from `Update`. It deliberately does not call it on `AdoptionPolicy_Adopt`, where replacing a declared spec is the intended behaviour. It is merged and released as [v0.64.0](https://github.com/aws-controllers-k8s/runtime/releases/tag/v0.64.0).

**The `go.mod` bump to v0.64.0 is included in this PR** rather than left as a follow-up. `auto-generate-controllers` resolves the runtime version from this module's `go.mod` (`cd/auto-generate/auto-generate-controllers.sh:91`), not from the latest runtime release, so merging the generator while the pin still read v0.63.0 would open fleet-wide PRs adding a generated `EnsureReferences` against a runtime that never calls it. The bump is `go.mod` and `go.sum` only, with no generated-output changes.

Against an older runtime the method is inert but harmless: it compiles and is simply never invoked, which is the state a controller is in between this merge and its own regeneration. Controllers generated before the method existed are unaffected and opt in by regenerating.

## What is emitted

| Reference | Emitted | Why |
|---|---|---|
| top-level (`spec.xRef`) | nothing | every write path starts from a `DeepCopy`, and the `*Ref` is a sibling of the concrete field, so nothing rebuilds it |
| through structs (`spec.a.xRef`, `spec.a.xRefs`) | assign the reference | it has one fixed address; every value the service reported stands |
| through a list (`spec.l[].xRef`) | nothing | it has no fixed address; out of scope, see below |

A reference field that is itself a list (`*Refs`, whose concrete sibling is a list of scalars) belongs in the **struct** row: the list is the leaf, not part of the path, so nothing has to be indexed to reach it.

```go
if desiredKO.Spec.VPCConfig != nil {
	if len(desiredKO.Spec.VPCConfig.SecurityGroupRefs) > 0 {
		if latestKO.Spec.VPCConfig == nil {
			latestKO.Spec.VPCConfig = &svcapitypes.VPCConfig{}
		}
		if len(latestKO.Spec.VPCConfig.SecurityGroupRefs) == 0 {
			latestKO.Spec.VPCConfig.SecurityGroupRefs = desiredKO.Spec.VPCConfig.SecurityGroupRefs
		}
	}
	if len(desiredKO.Spec.VPCConfig.SubnetRefs) > 0 {
		if latestKO.Spec.VPCConfig == nil {
			latestKO.Spec.VPCConfig = &svcapitypes.VPCConfig{}
		}
		if len(latestKO.Spec.VPCConfig.SubnetRefs) == 0 {
			latestKO.Spec.VPCConfig.SubnetRefs = desiredKO.Spec.VPCConfig.SubnetRefs
		}
	}
}
```

Only the reference is written, so nothing the service populated is touched.

Four properties of the shape above are load-bearing.

**Each container is materialised on the target.** Generated set-output code rebuilds a struct from the response and nils it when the response omits it -- lambda's `sdkCreate` has `} else { ko.Spec.VPCConfig = nil }`. Guarding on the container being present on *both* objects would therefore skip the reference in exactly the case it most needs restoring, and the spec patch would then delete the whole declared block rather than just the reference. Across the controllers, 36 of the 64 struct-nested containers can be nil'd this way inside `sdkCreate`/`sdkUpdate`. The hand-maintained hooks named below already materialise the container for this reason.

**The source-side guards enclose it.** A container is only constructed once the declared resource is known to hold a reference to put in it, so a resource that declares the container but no reference leaves the target untouched and no empty container reaches the patch.

**The guards are nested, not concatenated.** A single combined condition reached 661 characters on `ecs/CapacityProvider`'s three-level path, which gofmt does not wrap. Nesting also mirrors `ClearResolvedReferences`, which walks the same paths. This bounds the *guards*; an assignment line is as long as its two field paths and gofmt wraps neither form, so `ecs/CapacityProvider`'s deepest reference still produces a 210-character one, exactly as the hand-written hooks below already do.

**A container's guard is emitted once, however many references sit inside it.** The two references in `VPCConfig` above share one `!= nil` guard, and a container nested inside another is guarded within its parent rather than restating the parent's chain: `ecs/CapacityProvider` emits one guard tree over its three nested containers instead of four separate chains, 75 lines rather than 87. The materialisation stays per-reference, because it has to sit inside the guard establishing that the source holds a reference to put in the container -- sharing it would need those guards OR'd together, which is the concatenation the nesting exists to avoid. It is a nil-check either way, so repeating it costs nothing.

## This codifies an existing pattern

Restoring a nested `*Ref` from the declared resource is not new — three controllers already hand-maintain this assignment for want of a generated equivalent. `eks/cluster`:

```go
// templates/hooks/cluster/sdk_create_post_set_output.go.tpl
if desired.ko.Spec.ResourcesVPCConfig.SubnetRefs != nil {
	ko.Spec.ResourcesVPCConfig.SubnetRefs = desired.ko.Spec.ResourcesVPCConfig.SubnetRefs
}
if desired.ko.Spec.ResourcesVPCConfig.SecurityGroupRefs != nil {
	ko.Spec.ResourcesVPCConfig.SecurityGroupRefs = desired.ko.Spec.ResourcesVPCConfig.SecurityGroupRefs
}
```

`lambda/function` does the same for `VPCConfig`, and `opensearchservice/domain` for `VPCOptions` — the latter with a comment naming this issue directly:

```go
// To prevent https://github.com/aws-controllers-k8s/community/issues/2431
```

The generated code is the same assignment with stricter guards: it nil-checks the container on **both** objects and only writes when the target is actually missing the reference, so it cannot clobber a reference the service did report. What changes is that every controller with struct-nested references gets the behaviour without hand-writing it.

## Scope

Classifying each reference by the shape of the path to its `*Ref`, across the controllers with a generated `references.go`:

| Shape | References | Resources | Controllers | This PR |
|---|---:|---:|---:|---|
| top-level | 315 | 157 | 53 | not needed |
| struct-nested | 117 | 37 | 25 | **fixed** |
| list-nested | 38 | 21 | 12 | unchanged |

## Testing

Eleven unit tests in `pkg/generate/code/resource_reference_test.go`: top-level emits nothing, struct-nested single ref, struct-nested list-of-refs, list path emits nothing, a resource mixing struct- and list-nested, indent level, the two rejection paths (a reference within a map, and a model missing an ancestor field), that every container on the path is materialised on the target and only inside the source-side reference guard, that no **guard** exceeds 120 characters and none concatenates (`&&`/`||`), and that a container's guard is emitted once however many references sit inside it.

That last one uses a new `eks` `Cluster` fixture carrying both grouping shapes: two references in one container (`ResourcesVpcConfig`), and a container holding a reference of its own plus a deeper container that holds another (`OutpostConfig`, `OutpostConfig.ControlPlanePlacement`). It asserts the exact output and, separately, that each guard appears exactly once, so a reshuffle that reintroduces a duplicate fails on the reason rather than on a diff.

Regenerated `lambda-controller`, `ecs-controller` and `eks-controller`; `references.go` diffs are purely additive, output is gofmt-clean, and all three build in full. `lambda/function` restores `Code.S3BucketRef`, `VPCConfig.SecurityGroupRefs` and `VPCConfig.SubnetRefs` — the shape #2431 was filed for.

Verified on a cluster as a three-way comparison, all on current `ecs-controller` main, using an `ecs` `Service` that declares `networkConfiguration.awsVPCConfiguration.subnetRefs` and `.securityGroupRefs` -- a two-level struct path:

| Generated method | Runtime | Result |
|---|---|---|
| absent | published v0.63.0 | both refs **deleted**, resolved `subnets`/`securityGroups` written |
| present | published v0.63.0 | **identical** -- the method alone does nothing |
| present | runtime#267 | both refs **kept**, nothing leaked |

The middle row is what makes the comparison controlled: it isolates this PR's output from the runtime change, and confirms the method is inert without runtime support. Re-applying the manifest onto the corrupted spec halts the resource on `both resource reference wrapper and ID cannot be used together`, which is the failure this prevents.

A container mixing forms -- `subnetRefs` declared alongside a literal `securityGroups` -- restores only the reference and leaves the literal untouched, with no phantom `securityGroupRefs` invented.

## Not addressed

**References reached through a list.** No fixed address to assign to, and no sound way to pair an observed element with a declared one: an AWS response need not preserve request order. These behave exactly as they do today.

**The existing read-path hooks.** `lambda/function`, `eks/cluster` and `opensearchservice/domain` restore these same references by hand in `sdk_read_one_post_set_output`. Only their create-path halves are subsumed, and **the read-path halves must stay**: the `AdoptionPolicy_Adopt` and delete-path spec writes are not covered by runtime#267.


gustavodiaz7722:
Also addresses the following open issue https://github.com/aws-controllers-k8s/community/issues/2878

pkg/generate/code/resource_reference.go
```diff
@@ -503,3 +503,147 @@ func getReferencedStateForField(field *model.Field, indentLevel int) string {
...
+
+// hasListAncestor reports whether a list lies on the path to a field's reference
+// (`*Ref`) sibling.
+//
+// Only ancestors count. A reference field that is itself a list (`*Refs`, whose
+// concrete sibling is a list of scalars) is the leaf rather than part of the path,
+// so nothing has to be indexed to reach it.
+//
+// EnsureReferences uses this to skip such a reference; see its doc comment.
+func hasListAncestor(field *model.Field) (bool, error) {
```
knottnt: nit: Name/description is somewhat misleading since `map` will also return `true`. 
```suggestion
func hasCollectionAncestor(field *model.Field) (bool, error) {
```

pkg/generate/code/resource_reference.go
```diff
@@ -503,3 +503,147 @@ func getReferencedStateForField(field *model.Field, indentLevel int) string {
...
+// manifest puts the `*Ref` back beside that value, a pair
+// validateReferenceFields rejects, stopping reconciliation. See
+// aws-controllers-k8s/community#2361 and #2431.
+//
+// Only a reference reached through STRUCTS is emitted. It has one fixed address,
+// so exactly that field is assigned and every value the service reported stands.
+//
+// A TOP-LEVEL reference is skipped: every write path starts from a DeepCopy of the
+// object it was handed, and the `*Ref` is a sibling of the concrete field rather
+// than part of it, so nothing rebuilds it.
```
knottnt: nit: While largely true this is an assumption. Technically a custom code path could violate this.

pkg/generate/code/resource_reference.go
```diff
@@ -503,3 +503,147 @@ func getReferencedStateForField(field *model.Field, indentLevel int) string {
...
+//
+// A reference reached through a LIST is also skipped, and stays broken as it is
+// today. It has no fixed address: an AWS response need not preserve request order,
+// and element identity cannot be recovered from the resolved value either, since
+// many references resolve to a `Spec.*` path that is not unique across CRs. The
```
knottnt: > since many references resolve to a `Spec.*` path that is not unique across CRs

Could you expand on this? I'm not totally sure what this means.

pkg/generate/code/resource_reference.go
```diff
@@ -503,3 +503,147 @@ func getReferencedStateForField(field *model.Field, indentLevel int) string {
...
+// only assignment that avoids guessing a correspondence is to replace the whole
+// outermost list, which discards whatever the service populated inside it -- and
+// where the element type carries an AWS-assigned field the generated delta
+// compares (ec2's `NetworkACL.Associations`), leaves the stored spec permanently
+// disagreeing with the ReadOne result, i.e. an AWS Update every resync. Restoring
+// these precisely is left to a follow-up.
```
knottnt: Hmm we could enhance this feature to take into account a set of IDs provided in the generator.yaml that uniquely identify an entry in the list. But agree addressing this would make more sense as a follow-up.

knottnt:
@gustavodiaz7722 it would also be help if you could link a draft PR that shows an example of the generated code in the context of a full service controller.

gustavodiaz7722:
@knottnt Created a demo here https://github.com/aws-controllers-k8s/lambda-controller/pull/242

pkg/generate/code/resource_reference.go
gustavodiaz7722: Updated comment

pkg/generate/code/resource_reference.go
gustavodiaz7722: Fair point, will reword comment to properly frame the issue.

pkg/generate/code/resource_reference.go
gustavodiaz7722: Fixed in latest revision

knottnt (approved):

knottnt (approved):
@gustavodiaz7722 holding off on the lgtm until https://github.com/aws-controllers-k8s/runtime/pull/267 is merged

knottnt (approved):

PR #744, opened 2026-09-22
Apply input_wrapper_field_path unwrapping to the ReadMany identifier codegen

knottnt:
Issue #, if available:

Description of changes:
The identifier-discovery codegen read the raw ReadMany input shape, so it never saw through a list/struct wrapper (e.g. opensearchserverless's BatchGetLifecyclePolicy op whose input wraps the real identifiers in an Identifiers []Identifier batch member). For a composite-key resource this dropped every key past the primary from the generated adoption, required-check, and request-payload code.

This routes the four ReadMany-reachable sites through the wrapper-aware GetInputShape, so with `input_wrapper_field_path` set the generated code discovers the full composite key from the unwrapped element shape:

- SetResourceIdentifiers / PopulateResourceFromAnnotation adopt on all key components, not just the primary
- requiredFieldsMissingFromReadManyInput gate on every required component (Name == nil || Type == nil)
- newListRequestPayload build the one-element identifier slice (new setSDKReadManyListWrapper helper), replacing the need for a hand-written sdk_read_many_post_build_request hook

No-op for resources without a wrapper configured (existing ReadMany inputs have empty Required lists), verified byte-identical for EC2 Vpc, memorydb User, and cloudwatch-logs ResourcePolicy. Adds paired fixed/regression tests for each of the four generated surfaces against opensearchserverless LifecyclePolicy.


michaelhtm (approved):
/lgtm

